PR #9, opened 2025-07-08
assembler comment fix

moradna:


core/assembler.go
```diff
@@ -97,7 +97,7 @@ func (a *Assembler) Run() {
 	go a.processOrderedBatchAttestations()
 }
 
-// Core Assebler stops by the node Assebler when the channels for batches and BAs are closed.
+// Core Assembler stops by the node Assebler when the channels for batches and BAs are closed.
```
tock-ibm, line 100: Go comment style is:
```
// WaitTermination core Assembler is stopped by the node Assembler when the channels for batches and BAs are closed.
```

tock-ibm (approved):

Issue #121, opened 2025-07-15
Get rid of core


HagarMeir:
No need for a separation between node and core

PR #130, opened 2025-07-17
move type batch from core to types

HagarMeir:
See https://github.com/hyperledger/fabric-x-orderer/issues/121


common/types/mocks/batch.go
tock-ibm, line 1: Avoid mocks in `common/types`, in next PR, please

tock-ibm (approved):
@HagarMeir with comment for next PR

PR #133, opened 2025-07-20
avoid using batch mock in assembler tests

HagarMeir:


node/assembler/utils_test.go
```diff
@@ -0,0 +1,62 @@
...
+		requests = append(requests, make([]byte, requestSize))
+	}
+	return &TestBatch{seq: seq, shard: shard, primary: primary, requests: requests, digest: requests.Digest()}
+}
+
+func createTestBatchId(shard types.ShardID, primary types.PartyID, seq types.BatchSequence, digest []byte) types.BatchID {
+	return &TestBatch{seq: seq, shard: shard, primary: primary, digest: digest}
+}
+
+func assertBatchIdsEquals(t *testing.T, expectedBatchId, actualBatchId types.BatchID) {
```
tock-ibm, line 57: We have this at line 230 batch_fetcher

`if fb.Shard() == batchID.Shard() && fb.Primary() == batchID.Primary() && fb.Seq() == batchID.Seq() && bytes.Equal(fb.Digest(), batchID.Digest()) `

lets make a function that does 

`BatchIDEqual(a BatchID, b BatchID) bool` and use it everywhere. Next PR maybe

PR #140, opened 2025-07-22
Docker release pipeline

anushka159-a:
Fixed the Comparison operator logic.


anushka159-a:
I have changed the comparison logic in the Image tag prefix loop to use "$GITHUB_REPOSITORY_OWNER"  variable to compare the value to our repository owner as this former variable is available in GitHub Actions as an environment variable.

PR #142, opened 2025-07-24
move mapper from core to node/router

DorKatzelnick:
rename (router -> mapper)
moved from core to node/router
added a test


node/router/mapper_test.go
```diff
@@ -0,0 +1,40 @@
...
+	require.False(t, bytes.Equal(reqID1, reqID2))
+
+	// many shards test
+	numOfShards = 100
+	m = Mapper{logger, uint16(numOfShards)}
+	s1, reqID1 = m.Map(req1)
+	s2, reqID2 = m.Map(req2)
+	require.True(t, s1 != s2)
+	require.False(t, bytes.Equal(reqID1, reqID2))
+}
```
HagarMeir, line 40: maybe check what happens if you try to map it again
that it should be consistent 

also when mapping a lot of requests to lets say two shards they should be about evenly spread 
maybe we can test that?

node/batcher/requests_inspector_verifier.go
```diff
@@ -46,7 +46,7 @@ type RequestsInspectorVerifier struct {
 	shards                   []types.ShardID
 	shardID                  types.ShardID
 	logger                   types.Logger
-	mapper                   *core.Router // TODO make this into an interface?
+	mapper                   *router.Mapper // TODO make this into an interface?
```
HagarMeir, line 49: can you also get this todo done? 
this is good for testing 
and if we ever want to replace the mapping function

node/router/mapper_test.go
DorKatzelnick: done: added more tests

node/router/mapper.go
```diff
@@ -13,14 +13,19 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/common/types"
 )
 
-type Router struct {
+type MapperInterface interface {
```
HagarMeir, line 16: Just `Mapper` is enough
Don't need to add `Interface` to the name

HagarMeir: And I would call it maybe `ShardMapper`

node/router/mapper.go
```diff
@@ -13,14 +13,19 @@ import (
...
+	Map(request []byte) (shard uint16, reqID []byte)
+	GetNumberOfShards() uint16
```
HagarMeir, line 18: I see this is used only in tests? 
Maybe leave it out of the interface then

node/router/mapper.go
DorKatzelnick: fixed

node/router/mapper.go
DorKatzelnick: removed GetNumberOfShards from the interface

node/router/mapper_test.go
```diff
@@ -0,0 +1,102 @@
...
+		for i := 0; i < numOfRequest; i++ {
+			req := []byte("request " + strconv.Itoa(i))
+			_, reqID := m.Map(req)
+			IDsMap[string(reqID)]++
+		}
+		require.Equal(t, numOfRequest, len(IDsMap))
+	})
+}
+
+func TestMappers(t *testing.T) {
```
HagarMeir, line 92: please check how to write multiple tests in one testing function
for example here: common/types/batched_requests_test.go
or here: common/ledger/blkstorage/block_serialization_test.go

node/router/mapper.go
```diff
@@ -35,3 +39,7 @@ func CRC64RequestToShard(shardCount uint16) func([]byte) ([]byte, uint16) {
 		return buff, uint16(reqID) % shardCount
 	}
 }
+
+func (m MapperCRC64) GetNumberOfShards() uint16 {
```
HagarMeir, line 43: I don't think you really need this function

node/router/mapper_test.go
```diff
@@ -0,0 +1,103 @@
...
+	logger := testutil.CreateLogger(t, 0)
+
+	// test MapperCRC64
+	m := MapperCRC64{logger, uint16(4)}
```
HagarMeir, line 99: do you need this `uint16`?
also the line after

node/router/mapper.go
DorKatzelnick: removed it

node/router/mapper_test.go
DorKatzelnick: fixed

node/router/mapper_test.go
```diff
@@ -0,0 +1,103 @@
...
+	MultiShardTests(t, m, m.GetNumberOfShards())
+	// one shard test
+	OneShardTest(t, MapperCRC64{logger, uint16(1)})
```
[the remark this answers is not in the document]
DorKatzelnick: separated to two functions

PR #143, opened 2025-07-27
Create local msp final

MayRosenbaum:


config/bccsp/bccsp.go
```diff
@@ -0,0 +1,47 @@
...
+	Library  string `yaml:"Library,omitempty"`
+
+	AltID  string         `yaml:"AltID,omitempty"`
+	KeyIDs []KeyIDMapping `yaml:"KeyIDs,omitempty"`
+}
+
+type KeyIDMapping struct {
+	SKI string `yaml:"SKI,omitempty"`
+	ID  string `yaml:"ID,omitempty"`
+}
```
tock-ibm, line 47: Instead of redoing this, why can't we use the stuff that is already available at `github.com/hyperledger/fabric-lib-go/bccsp/factory/swfactory.go`?

and the `github.com/hyperledger/fabric-lib-go/bccsp/` package in general?

common/msp/localmsp.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hyperledger/fabric-lib-go/bccsp/factory"
+	"github.com/hyperledger/fabric-x-orderer/config/bccsp"
+	"github.com/hyperledger/fabric/msp"
+)
+
+func BuildLocalMSP(localMSPDir string, localMSPID string, bccspConfig *bccsp.BCCSP) *msp.MSP {
```
tock-ibm, line 17: What if, instead of our `bccspConfig *bccsp.BCCSP` we would read the `factory.FactoryOpts` directly from the local config  yaml ?

node/server/arma.go
```diff
@@ -150,8 +151,15 @@ func launchBatcher(stop chan struct{}) func(configFile *os.File) {
 		if err != nil {
 			panic(fmt.Sprintf("error launching batcher, err: %s", err))
 		}
+
 		conf := config.ExtractBatcherConfig()
 
+		localmsp := msp.BuildLocalMSP(config.LocalConfig.NodeLocalConfig.GeneralConfig.LocalMSPDir, config.LocalConfig.NodeLocalConfig.GeneralConfig.LocalMSPID, config.LocalConfig.NodeLocalConfig.GeneralConfig.BCCSP)
+		signer, err := (*localmsp).GetDefaultSigningIdentity()
```
tock-ibm, line 158: if you return the interface and not a pointer to it, you could avoid the `*`

config/bccsp/bccsp.go
MayRosenbaum: We can use the staff and have a dependency on `fabric-lib-go`
In a separate PR

tock-ibm (approved):

PR #144, opened 2025-07-28
Adding make targets for docker-local and docker-multiarch

anushka159-a:
I have added make targets for local-OS image locally and to make multi-arch images locally.


MayRosenbaum:
I updated the Dockerfile following the Docker documentation:
https://docs.docker.com/build/building/multi-platform/#cross-compiling-a-go-application

PR #148, opened 2025-07-31
add backoff to pull of deliver client

HagarMeir:


node/delivery/deliver_client.go
```diff
@@ -63,18 +63,23 @@ func SingleSpecifiedSeekInfo(seq uint64) *orderer.SeekInfo {
...
 
-		select {
-		case <-context.Done():
-			logger.Infof("Returning since context is done")
-			if onClose != nil {
-				onClose()
-			}
-			return
-		default:
+	time.Sleep(time.Second)
```
tock-ibm, line 67: why sleep for 1s before even starting? I think the first attempt should be done immediately

node/delivery/deliver_client.go
```diff
@@ -63,18 +63,23 @@ func SingleSpecifiedSeekInfo(seq uint64) *orderer.SeekInfo {
...
+			onClose()
 		}
+		return
+	default:
+	}
+
+	count := 0
+	retryInterval := minRetryInterval
+
+	for {
```
tock-ibm, line 82: if the endpoint() function return "" is there any hope of getting an actual endpoint? I think not... and then we are in a busy loop running full speed. Let's get rid of that... I don't think getting "" is valid for `endpoint()`

tock-ibm (dismissed):

PR #157, opened 2025-08-03
move type batch from core to types

HagarMeir:
#121


common/types/types.go
```diff
@@ -64,3 +64,9 @@ type BatchAttestationFragment interface {
 	Epoch() int64
 	String() string
 }
+
+//go:generate counterfeiter -o mocks/batch.go . Batch
```
tock-ibm, line 68: avoid creating and using mocks in the types package, in the next PR

PR #165, opened 2025-08-11
check that delivery clients gets all of the submitted transactions

gengur:
check that delivery clients get all of the submitted transactions

### DIFF
--- a/test/utils_test.go
+++ b/test/utils_test.go
@@ -36,6 +36,7 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/node/router"
 	"github.com/hyperledger/fabric-x-orderer/testutil"
 	"github.com/hyperledger/fabric-x-orderer/testutil/client"
+	"github.com/hyperledger/fabric/protoutil"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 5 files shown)


test/utils_test.go
```diff
@@ -52,6 +54,10 @@ func (n *node) ToString() string {
 	return fmt.Sprintf("GRPC.Address: %s", n.GRPCServer.Address())
 }
 
+const (
+	NEED_VERIFICATION_ON_PREFIX = "__verify__"
```
HagarMeir, line 58: isn't it easier to just pass a flag to the `PullFromAssemblers `  func if we want to check for missing or not? 
instead of encoding it inside the transaction?

HagarMeir: actually why not just always check if there are any missing txs in all tests

test/utils_test.go
```diff
@@ -329,52 +330,86 @@ func sendTxn(workerID int, txnNum int, routers []*router.Router) {
...
 	}
 
 	waitForPullDone.Wait()
 	return pullInfos
 }
 
-func PullFromAssembler(t *testing.T, userConfig *armageddon.UserConfig, partyID types.PartyID, startBlock uint64, endBlock uint64, transactions int, blocks int, timeout int) (*BlockPullerInfo, error) {
-	require.NotNil(t, userConfig)
+func pullFromAssembler(t *testing.T, userConfig *armageddon.UserConfig, partyID types.PartyID, startBlock uint64, endBlock uint64, transactions int, blocks int, timeout int, need_verification bool) (*BlockPullerInfo, error) {
+	logger := testutil.CreateLoggerForModule(t, fmt.Sprintf("[BlockPuller%d]", partyID), zap.DebugLevel)
```
HagarMeir, line 394: why not use t.log?

test/utils_test.go
```diff
@@ -400,8 +435,35 @@ func PullFromAssembler(t *testing.T, userConfig *armageddon.UserConfig, partyID
...
 		}
 
+		data := block.GetData().GetData()
+		blocksNumber := len(data)
+
 		totalBlocks++
-		totalTxs += uint64(len(block.GetData().GetData()))
+
+		if blocksNumber == 1 {
+			return nil
```
HagarMeir, line 444: can't there be a block with only one tx that is not the genesis? 
just ignore the genesis block

test/utils_test.go
```diff
@@ -413,8 +475,19 @@ func PullFromAssembler(t *testing.T, userConfig *armageddon.UserConfig, partyID
 		return nil
 	}
 
-	fmt.Printf("Pulling from party: %d\n", partyID)
+	logger.Debugf("Pulling from party: %d\n", partyID)
```
HagarMeir, line 478: just use t.log

test/utils_test.go
```diff
@@ -399,8 +415,38 @@ func PullFromAssembler(t *testing.T, userConfig *armageddon.UserConfig, partyID
 			}
 		}
 
+		data := block.GetData().GetData()
+		blocksNumber := len(data)
```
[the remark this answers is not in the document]
gengur: right

Issue #170, opened 2025-08-12
Update armageddon load to use a broadcast client that can tolerate router failure


MayRosenbaum:
To test Router failures, the armageddon `load` command should use a broadcast client that tolerates router failure.
When a router is stopped, the broadcast client continues to send transactions to all available routers and, in parallel, tries to reconnect to the failed router.

PR #180, opened 2025-08-14
[WIP] Update load command to tolerate router faulty

MayRosenbaum:
issue #170 

`UserConfig` was removed to prevent an import cycle
The armageddon.load command uses the Broadcast client


MayRosenbaum:
In this PR, the loader is based on the broadcast client written for integration tests.
For the performance tests, a new broadcast client was implemented, in PR #227

PR #217, opened 2025-08-26
Fast broadcast client

gengur:


tock-ibm:
Lets put this aside util we get to this. We can re-open it later if the need arises.

PR #234, opened 2025-09-02
extract channelID from block

DorKatzelnick:
- extract the channel name from block
- add the channel name to the router filter flow (w.o verifying)
- add RouterFilterConfig struct and appropriate get methods

related issue: #195

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -85,7 +85,7 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 				return nil, nil, errors.Wrapf(err, "failed to unmarshal consensus metadata to a shared configuration")
 			}
 		} else {
-			return nil, nil, errors.Wrapf(err, "failed to read a cofig block, path is empty")
+			return nil, nil, errors.Wrapf(err, "failed to read a config block, path is empty")
 		}
 	default:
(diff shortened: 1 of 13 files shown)


config/config.go
```diff
@@ -375,3 +373,15 @@ func blockToPublicKey(block *pem.Block) []byte {
 
 	return pemPublicKey
 }
+
```
tock-ibm, line 376: Why separate? I think it is better to keep it as one... easier to understand

node/config/config.go
```diff
@@ -65,20 +65,25 @@ type ConsenterInfo struct {
 	TLSCACerts []RawBytes
 }
 
+type RouterFilterConfig struct {
```
tock-ibm, line 68: I think keep it as one struct, easier to understand... Is there another reason for this?

DorKatzelnick: I’m changing it so they’re all in one struct, but I’m keeping the RouterFilterConfig struct (now in the router) so that it implements the NodeConfig interface, and we can have the proper variables and getters according to Go conventions.

config/config.go
DorKatzelnick: This was in a separate function because the genesis block was needed for the channel ID, but it was not an input for the other function. Anyway, I’m changing it.

PR #236, opened 2025-09-02
[WIP] change broadcast error flow

DorKatzelnick:
This PR address two (possible) issues:

1. handle errors when router send responses to the client
2. handle a closeSend by the client, before the router send all the responses.


tock-ibm:
@DorKatzelnick Closing, when you need a reference simply look it up

PR #244, opened 2025-09-07
Config store

tock-ibm:
A simple config store.

filerepo package copied from hyperledger fabric.


common/configstore/configstore.go
```diff
@@ -0,0 +1,151 @@
...
+)
+
+const (
+	fileSuffix        = "pb.bin"
+	fileNameSeparator = "-"
+	fileNameTemplate  = "configblock" + fileNameSeparator + "%d"
+)
+
+// Store holds a set of numbered blocks, typically config blocks.
+// Blocks get be added, listed, and retrieved.
```
HagarMeir, line 26: Blocks can be

common/configstore/configstore.go
```diff
@@ -0,0 +1,151 @@
...
+// A convenience method allows fetching the last block.
+type Store struct {
+	fileRepo *filerepo.Repo
+}
+
+// NewStore initializes a new config store at the storeDir.
+// All file system operations on the returned file repo are thread safe.
+func NewStore(storeDir string) (*Store, error) {
+	repo, err := filerepo.New(storeDir, fileSuffix)
```
HagarMeir, line 35: why use file suffix here?

common/configstore/configstore_test.go
```diff
@@ -0,0 +1,148 @@
...
+		s, err := configstore.NewStore(dir)
+		assert.NoError(t, err)
+		assert.NotNil(t, s)
+		nums, err := s.ListBlockNumbers()
+		assert.NoError(t, err)
+		assert.Len(t, nums, 0)
+	})
+
+	t.Run("existing", func(t *testing.T) {
+		s, err := configstore.NewStore(dir)
```
HagarMeir, line 32: test also existing with a block already inside

common/configstore/configstore.go
tock-ibm: The filerepo lib wants a file suffix, as it can handle multiple types of files by their suffix, in the same folder.

HagarMeir (approved):

PR #253, opened 2025-09-10
Sig filter

MayRosenbaum:


common/policy/policy.go
```diff
@@ -11,10 +11,28 @@ import (
...
+//go:generate counterfeiter -o mocks/policy.go . policy
+type policy interface {
+	policies.Policy
+}
+
+var (
+	_ policyManager = &mocks.FakePolicyManager{}
+	_ policy        = &mocks.FakePolicy{}
+)
+
```
tock-ibm, line 35: Don't expose the fakes in the main code

MayRosenbaum: removed the test

PR #255, opened 2025-09-11
broadcast client log improvements

MayRosenbaum:


common/tools/armageddon/broadcast_client.go
```diff
@@ -135,9 +150,28 @@ func (c *BroadcastTxClient) InitStreams() error {
...
+		for {
+			select {
+			case <-c.stopChan:
+				return
+			case <-ticker.C:
+				for i := range c.txsCounter {
+					streamInfoLogger := c.streamsToRouters[i].GetStreamLogger()
+					sentTxs := atomic.LoadUint64(&c.txsCounter[i])
+					streamInfoLogger.Infof("BroadcastClient to Router %d sent %d transactions in the last 10 seconds", i+1, sentTxs)
+					atomic.StoreUint64(&c.txsCounter[i], 0)
```
HagarMeir, line 169: try using swap instead of load and store

common/tools/armageddon/broadcast_client.go
```diff
@@ -30,6 +31,12 @@ type StreamInfo struct {
 	endpoint              string
 	lock                  sync.Mutex
 	logger                *flogging.FabricLogger
+	sentTxs               uint64
+}
+
+func (streamInfo *StreamInfo) Report() {
+	sentTxs := atomic.SwapUint64(&streamInfo.sentTxs, 0)
+	streamInfo.logger.Infof("BroadcastClient to Router %v sent %d transactions in the last 10 seconds", streamInfo.endpoint, sentTxs)
```
HagarMeir, line 39: maybe pass the number "10" as a var if we ever want to change it

PR #260, opened 2025-09-14
fix statistics routine in router

DorKatzelnick:
add statisticsReport to the router
with start and stop methods.

### DIFF
--- a/node/router/router.go
+++ b/node/router/router.go
@@ -41,6 +41,7 @@ type Router struct {
 	incoming         uint64
 	routerNodeConfig *nodeconfig.RouterNodeConfig
 	verifier         *requestfilter.RulesVerifier
+	stopChan         chan struct{}
 }
 
 func NewRouter(config *nodeconfig.RouterNodeConfig, logger types.Logger) *Router {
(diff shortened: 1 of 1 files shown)


node/router/router.go
```diff
@@ -180,18 +183,27 @@ func createRouter(shardIDs []types.ShardID, batcherEndpoints map[types.ShardID]s
 		shardIDs:         shardIDs,
 		routerNodeConfig: rconfig,
 		verifier:         verifier,
+		stopChannel:      make(chan bool, 1),
```
HagarMeir, line 186: what happens if you make this channel not buffered?

HagarMeir: `stopChan: make(chan struct{})`

node/router/router.go
```diff
@@ -180,18 +183,27 @@ func createRouter(shardIDs []types.ShardID, batcherEndpoints map[types.ShardID]s
...
 	}
 
 	for _, shardId := range shardIDs {
 		r.shardRouters[shardId] = NewShardRouter(logger, batcherEndpoints[shardId], batcherRootCAs[shardId], rconfig.TLSCertificateFile, rconfig.TLSPrivateKeyFile, rconfig.NumOfConnectionsForBatcher, rconfig.NumOfgRPCStreamsPerConnection, verifier)
 	}
 
 	go func() {
+		r.logger.Debugf("statistics reporter in router is starting")
```
HagarMeir, line 194: change comment, this is not a statistics reporter 
also I think it is safe to make it info instead of debug

HagarMeir (approved):

PR #261, opened 2025-09-15
[WIP] configure mem pool buckets rotation

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/269

### DIFF
--- a/request/pool.go
+++ b/request/pool.go
@@ -9,6 +9,7 @@ package request
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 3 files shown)


request/pool.go
```diff
@@ -93,15 +94,15 @@ func (rp *Pool) createPendingStore() *PendingStore {
...
+		Time:                  time.NewTicker(rp.options.FirstStrikeThreshold / 10).C,
 		StartTime:             time.Now(),
 		Logger:                rp.logger,
 		SecondStrikeThreshold: rp.options.SecondStrikeThreshold,
 		FirstStrikeThreshold:  rp.options.FirstStrikeThreshold,
 		OnDelete: func(key string) {
 			rp.semaphore.Release(1)
 		},
-		Epoch:                time.Second,
+		Epoch:                time.Duration(rand.Intn(50))*time.Millisecond + rp.options.FirstStrikeThreshold/10,
```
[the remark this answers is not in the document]
HagarMeir: ok

request/pool.go
```diff
@@ -88,20 +90,31 @@ func (rp *Pool) start() {
 	rp.pending.Start()
 }
 
+func (rp *Pool) addRandomnessToFirstStrike() time.Duration {
```
tock-ibm, line 93: Lets make sure that in the worse case the first strike is still > BatchTimeout. IE compare the BatchTimeout with the FirstStrikeThreshold. It seems to me the FirstStrikeThreshold should be at least 5*BatchTimeout or so...

Even without randomization, if FirstStrikeThreshold < BatchTimeout + network-delay we'll get constant forwarding. So some sanity checks are required.

HagarMeir: Will add checks 
Currently the batch timeout is 500ms and we are running tests with first strike of 2s

Issue #205, opened 2025-08-19
On restart - every node starts from last config


tock-ibm:
- [x] Router and batcher start from last config from config store
- [x] Assembler starts from last config from block ledger
- [x] Consenter starts from last config from decision ledger

Issue #204, opened 2025-08-19
On first boot - routers and batchers copy genesis block to config store


tock-ibm:
On the first boot the router and the batcher initialize their config store, and copy the genesis block to the config-store.
If the config store already exists and is not empty, this step is skipped and the node starts from the last config (#205 ).

PR #275, opened 2025-09-21
Router config store

MayRosenbaum:
issue #204 #205 

1. ReadConfig initializes the shared configuration for routers and batchers by loading the last block from the config store if it exists. If not, it initializes the shared configuration from the genesis block and adds the genesis block to the config store.
2. Test ReadConfig functionality
3. Init config store in router

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,8 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 	nodeRole := findRoleFromLocalConfig(conf.LocalConfig)
 
 	var lastConfigBlock *common.Block
+	var configStore *configstore.Store
+
 	switch conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.Method {
 	case "yaml":
 		if conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.File != "" {
@@ -70,7 +72,7 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 	case "block":
 		// If node is router or batcher, check if config store has blocks, and if yes bootstrap from the last block
 		if nodeRole == "router" || nodeRole == "batcher" {
-			configStore, err := configstore.NewStore(conf.LocalConfig.NodeLocalConfig.FileStore.Path)
+			configStore, err = configstore.NewStore(conf.LocalConfig.NodeLocalConfig.FileStore.Path)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed creating router config store: %s", err)
 			}
@@ -89,6 +91,13 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 			lastConfigBlock, err = readGenesisBlockFromBootstrapPath(conf.LocalConfig)
 			if err != nil {
 				return nil, nil, fmt.Errorf("failed to read genesis block from bootstrap file, err: %v", err)
+			}
+
+			if configStore != nil {
+				err = configStore.Add(lastConfigBlock)
+				if err != nil {
+					return nil, nil, fmt.Errorf("failed to add the genesis block to the config store: %s", err)
+				}
 			}
 		}
 
--- a/config/config.go
+++ b/config/config.go
@@ -83,7 +83,8 @@ func ReadConfig(configFilePath string, logger types.Logger) (*Configuration, *co
 			}
 
 			if len(listBlockNumbers) > 0 {
-				lastBlockNumber := uint64(len(listBlockNumbers) - 1)
+				lastBlockNumberIdx := uint64(len(listBlockNumbers) - 1)
+				lastBlockNumber := listBlockNumbers[lastBlockNumberIdx]
 				lastConfigBlock, err = configStore.GetByNumber(lastBlockNumber)
 				if err != nil {
 					return nil, nil, fmt.Errorf("failed to retrieve the last block from %s config store: %s", nodeRole, err)
--- a/config/config.go
+++ b/config/config.go
@@ -21,6 +21,7 @@ import (
 	"github.com/hyperledger/fabric-lib-go/bccsp/factory"
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-x-common/protoutil"
+	"github.com/hyperledger/fabric-x-orderer/common/configstore"
 	"github.com/hyperledger/fabric-x-orderer/common/policy"
 	"github.com/hyperledger/fabric-x-orderer/common/types"
 	"github.com/hyperledger/fabric-x-orderer/common/utils"
@@ -53,7 +54,9 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 		return nil, nil, err
 	}
 
-	var genesisBlock *common.Block
+	nodeRole := findRoleFromLocalConfig(conf.LocalConfig)
+
+	var lastConfigBlock *common.Block
 	switch conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.Method {
 	case "yaml":
 		if conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.File != "" {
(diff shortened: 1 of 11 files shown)


config/config.go
```diff
@@ -65,34 +70,95 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
+		return nil, errors.Wrapf(err, "failed to unmarshal consensus metadata to a shared configuration")
+	}
+
+	return sharedConfig, nil
+}
+
+func findRoleFromLocalConfig(config *LocalConfig) string {
+	if config.NodeLocalConfig.RouterParams != nil {
+		return "router"
+	}
```
tock-ibm, line 151: What if the local config has both RouterParams and BatcherParams?
How do we enforce it has only one?

config/config.go
```diff
@@ -53,7 +54,11 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
 		return nil, nil, err
 	}
 
-	var genesisBlock *common.Block
+	nodeRole := findRoleFromLocalConfig(conf.LocalConfig)
```
tock-ibm, line 57: Log what the role is and panic if there is none

tock-ibm: Maybe we should encode the role in the local config? then we can unify all the local configs into one except with a different role? Not for now - lets talk about it and see if it makes sense or simplifies deployment.

config/config.go
```diff
@@ -65,34 +70,95 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
 	case "block":
-		if conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.File != "" {
-			blockPath := conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.File
-			data, err := os.ReadFile(conf.LocalConfig.NodeLocalConfig.GeneralConfig.Bootstrap.File)
+		// If node is router or batcher, check if config store has blocks, and if yes bootstrap from the last block
+		if nodeRole == "router" || nodeRole == "batcher" {
+			configStore, err = configstore.NewStore(conf.LocalConfig.NodeLocalConfig.FileStore.Path)
 			if err != nil {
-				return nil, nil, fmt.Errorf("could not read block %s", blockPath)
+				return nil, nil, fmt.Errorf("failed creating router config store: %s", err)
```
tock-ibm, line 77: It is not only a router, also possibly a batcher; i.e. print the node role: %s <- nodeRole

config/config.go
```diff
@@ -65,34 +70,95 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
 			}
-			genesisBlock, err = protoutil.UnmarshalBlock(data)
+
+			listBlocks, err := configStore.ListBlocks()
 			if err != nil {
-				return nil, nil, fmt.Errorf("error unmarshalling to block: %s", err)
+				return nil, nil, fmt.Errorf("failed retrieve a sorted array of all block numbers: %s", err)
```
tock-ibm, line 82: failed to list blocks from %s config store: %s

add the role

config/config.go
```diff
@@ -65,34 +70,95 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
+		// If node is router or batcher, check if config store has blocks, and if yes bootstrap from the last block
+		if nodeRole == "router" || nodeRole == "batcher" {
+			configStore, err = configstore.NewStore(conf.LocalConfig.NodeLocalConfig.FileStore.Path)
 			if err != nil {
-				return nil, nil, fmt.Errorf("could not read block %s", blockPath)
+				return nil, nil, fmt.Errorf("failed creating router config store: %s", err)
 			}
-			genesisBlock, err = protoutil.UnmarshalBlock(data)
+
+			listBlocks, err := configStore.ListBlocks()
```
tock-ibm, line 80: This is reading all the blocks, which is unnecessary, you can `ListBlockNumbers` and then `GetByNumber`.

Another option is to call "Last()" and catch the error `"config store is empty"`.

config/config.go
```diff
@@ -65,34 +70,95 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
 			if err != nil {
-				return nil, nil, errors.Wrapf(err, "failed to unmarshal consensus metadata to a shared configuration")
+				return nil, nil, fmt.Errorf("failed to read genesis block from bootstrap file, err: %v", err)
+			}
+
+			if configStore != nil {
+				err = configStore.Add(lastConfigBlock)
+				if err != nil {
+					return nil, nil, fmt.Errorf("failed to add the genesis block to the config store: %s", err)
+				}
```
tock-ibm, line 100: Add logs info when it is successful, it only happens once; log the number of the last config block and whether the config store was initialized.

config/config.go
MayRosenbaum: Ok

config/config.go
MayRosenbaum: Currently, we dont enforce it.
I propose to check it in the "LoadLocalConfig" and this method will also return the role.
In a case of multiple params or empty parmas a panic occurs.

config/local_config_test.go
```diff
@@ -52,39 +52,43 @@ func TestLoadARMALocalConfigAndCrypto(t *testing.T) {
 	// 3.
 	for i := 1; i <= len(networkLocalConfig.PartiesLocalConfig); i++ {
 		configPath := path.Join(dir, "config", fmt.Sprintf("party%d", i), "local_config_router.yaml")
-		routerLocalConfigLoaded, err := config.LoadLocalConfig(configPath)
+		routerLocalConfigLoaded, role, err := config.LoadLocalConfig(configPath)
```
tock-ibm, line 55: Add a test in which the LoadLocalConfig fails because of multiple roles or no roles, to check the `validateNodeLocalConfigParams`.

node/server/arma_test.go
```diff
@@ -185,6 +186,52 @@ func TestLaunchArmaNode(t *testing.T) {
...
+		require.NoError(t, err)
+
+		_, lastConfigBlock, err := config.ReadConfig(configPath, testLogger)
+		require.NoError(t, err)
+		require.NotNil(t, lastConfigBlock)
+
+		listBlocks, err = configStore.ListBlocks()
+		require.NoError(t, err)
+		require.Equal(t, len(listBlocks), 2)
+		require.Equal(t, genesisBlock.Header.Number, uint64(1))
```
tock-ibm, line 232: genesis should be 0. this is happening because `newConfigBlock := genesisBlock` assigns the pointer, they both point to the same block which you then change... make newConfigBlock a new block and recreate the header... 

or check that 
`require.Equal(t, listBlocks[0].Header.Number, uint64(0))`

config/config.go
```diff
@@ -65,34 +69,86 @@ func ReadConfig(configFilePath string) (*Configuration, *common.Block, error) {
...
 			if err != nil {
-				return nil, nil, fmt.Errorf("error unmarshalling to block: %s", err)
+				return nil, nil, fmt.Errorf("failed to list blocks from %s config store: %s", nodeRole, err)
 			}
 
-			consensusMetaData, err := ReadSharedConfigFromBootstrapConfigBlock(genesisBlock)
-			if err != nil {
-				return nil, nil, err
+			if len(listBlockNumbers) > 0 {
+				lastBlockNumber := uint64(len(listBlockNumbers) - 1)
```
tock-ibm, line 86: This is not the last block number, this is the index of the last...
The last block number is `realLastBlockNumber := listBlockNumbers[lastBlockNumber]`

node/server/arma_test.go
```diff
@@ -185,6 +186,52 @@ func TestLaunchArmaNode(t *testing.T) {
...
+		require.Equal(t, len(listBlocks), 1)
+		require.Equal(t, genesisBlock.Header.Number, uint64(0))
+		require.Equal(t, listBlocks[0].Header.Number, uint64(0))
+
+		// Add a fake block to the config store
+		// ReadConfig again, expect for the fake block to be the last block
+		newConfigBlock := genesisBlock
+		newConfigBlock.Header.Number = 1
+		err = configStore.Add(newConfigBlock)
+		require.NoError(t, err)
```
tock-ibm, line 223: In light of the index bug above... check it when the next config block is not 1, but 5; this way the index is different than the block number

node/server/arma_test.go
MayRosenbaum: Correct,  good point. I wanted to avoid cloning the block since it is just a test

PR #276, opened 2025-09-22
Batcher metrics in logs

moradna:
#262

### DIFF
--- a/node/batcher/batcher.go
+++ b/node/batcher/batcher.go
@@ -69,6 +69,8 @@ type Batcher struct {
 	primaryLock sync.RWMutex
 	term        uint64
 	primaryID   types.PartyID
+
+	Metrics *BatcherMetrics
 }
 
 func (b *Batcher) Run() {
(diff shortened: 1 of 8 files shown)


node/batcher/batcher.go
HagarMeir, line 315: add second strike metrics?

node/batcher/batcher_metrics.go
```diff
@@ -0,0 +1,111 @@
...
+				m.role(),
+				sec,
+				created-prevC, float64(created-prevC)/sec, created,
+				pulled-prevP, float64(pulled-prevP)/sec, pulled,
+				m.batchedTxsTotal.Load(),
+				m.memPoolSize.Load(),
+				m.routerTxsTotal.Load(),
+				m.roleChangesTotal.Load(),
+				m.complaintsTotal.Load(),
+				m.firstResendsTotal.Load(),
```
HagarMeir, line 96: maybe also print resends in an interval and not just total 
there could be a lot

HagarMeir (approved):

Issue #75, opened 2025-07-09
Batcher - Router - shared verification pipeline 


tock-ibm:
Both the routers and the batchers need to verify clients requests
Clients requests should go through a verification pipeline that checks signatures but also formatting, size, and it should also determine the mapping into shards.
This can be inspired by the verification pipeline in Fabric.

tock-ibm:
### Comments from original issue:

---

**Comment from @tock-ibm:**

The package in Fabric that implements the verification process is here:
`orderer/common/msgprocessor`

it is called for example here:
`orderer/common/broadcast/broadcast.go  L162` 


---

**Comment from @tock-ibm:**

See also: #146 #420 

---

DorKatzelnick:
Added the verification flow.
The remaining signature filter is addressed in issue #293.

Issue #73, opened 2025-07-09
Batcher and router shard mapper


tock-ibm:
Both the router and the batcher need to map requests into shard; router to send to the correct batcher and batcher to verify the request. 
However, currently the batcher needs to unmarshal the request to make it match the request the router is using for mapping, in order to get the same result. 
We need to avoid this unmarshaling. 
Also a mapper interface should be introduced.

PR #279, opened 2025-09-25
request verification in batcher

DorKatzelnick:
issues: #73 and #75

### DIFF
--- a/node/batcher/batcher.go
+++ b/node/batcher/batcher.go
@@ -144,22 +144,24 @@ func (b *Batcher) Deliver(stream orderer.AtomicBroadcast_DeliverServer) error {
 }
 
 func (b *Batcher) Submit(ctx context.Context, req *protos.Request) (*protos.SubmitResponse, error) {
+	// TODO: certificate pinning (bathcer trust router from his own party.)
+	b.logger.Debugf("Received request %x", req.Payload)
+
+	if err := b.requestsInspectorVerifier.VerifyRequestFromRouter(req); err != nil {
+		b.logger.Panicf("Failed verifying request before submitting from router; err: %v", err)
+		// TODO should return response with error?
+	}
+
 	traceId := req.TraceId
 	req.TraceId = nil
-
 	rawReq, err := proto.Marshal(req)
 	if err != nil {
 		b.logger.Panicf("Failed marshaling request: %v", err)
 	}
 
-	b.logger.Infof("Received request %x", req.Payload)
-
 	var resp protos.SubmitResponse
 	resp.TraceId = traceId
-	if err := b.requestsInspectorVerifier.VerifyRequest(rawReq); err != nil {
-		// TODO make sure the router verifies the request
-		b.logger.Panicf("Failed verifying request before submitting from router; err: %v", err)
-	}
+
 	if err := b.batcher.Submit(rawReq); err != nil {
 		resp.Error = err.Error()
 	}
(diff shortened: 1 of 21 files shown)


node/batcher/batcher.go
```diff
@@ -145,22 +145,24 @@ func (b *Batcher) Deliver(stream orderer.AtomicBroadcast_DeliverServer) error {
 }
 
 func (b *Batcher) Submit(ctx context.Context, req *protos.Request) (*protos.SubmitResponse, error) {
+	// TODO: certificate pinning (bathcer trust router from his own party.)
+	b.logger.Infof("Received request %x", req.Payload)
```
HagarMeir, line 149: Not sure we want this in Infof

HagarMeir: And maybe we can print something else? ID or something?
the payload won't give us much

node/batcher/requests_inspector_verifier.go
```diff
@@ -21,54 +22,50 @@ import (
...
 	}
 	riv.mapper = router.MapperCRC64{Logger: riv.logger, ShardCount: uint16(len(riv.shards))}
 	return riv
 }
 
+func createBatcherRulesVerifier(config *config.BatcherNodeConfig) *requestfilter.RulesVerifier {
+	rv := requestfilter.NewRulesVerifier(nil)
+	rv.AddRule(requestfilter.PayloadNotEmptyRule{})
+	rv.AddRule(requestfilter.NewMaxSizeFilter(config))
+	// rv.AddRule(requestfilter.NewSigFilter(config, policies.ChannelWriters)) - TBD
```
HagarMeir, line 65: add a TODO?

node/batcher/requests_inspector_verifier.go
```diff
@@ -117,27 +114,48 @@ func (r *RequestsInspectorVerifier) VerifyBatchedRequests(reqs types.BatchedRequ
...
+
+// VerifyRequestFromRouter verifies a request that comes from the router in the batcher's party.
+func (r *RequestsInspectorVerifier) VerifyRequestFromRouter(req *comm.Request) error {
+	// check that the shard is correct
+	if err := r.VerifyRequestShard(req); err != nil {
+		rawReq, _ := proto.Marshal(req)
+		return errors.Errorf("failed verifying request from router with id: %s; err: %v", r.RequestID(rawReq), err)
+	}
+
+	// TBD - check config sequence, and verify again if needed.
```
HagarMeir, line 136: this is just to reverify if we have a higher config?

node/batcher/batcher.go
DorKatzelnick: I can change it to debug. 
We can extract the RequestID but this will require to marshall the request first.
Or we can just omit this log, it is only in submit, and we are using submitStream from the router (with no such log)

node/batcher/requests_inspector_verifier.go
DorKatzelnick: added TODO comment

node/batcher/requests_inspector_verifier.go
```diff
@@ -21,54 +22,51 @@ import (
...
+		logger:          logger,
+		shardID:         config.ShardId,
+		shards:          config.GetShardsIDs(),
+		batchMaxSize:    config.BatchMaxSize,
+		batchMaxBytes:   config.BatchMaxBytes,
+		requestMaxBytes: config.RequestMaxBytes,
 	}
-	if requestVerifier == nil {
-		riv.requestVerifier = riv
+	if requestVerifier != nil {
```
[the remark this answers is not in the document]
HagarMeir: actually maybe this isn't easy to do because the tests actually uses this mock 
then maybe we can leave it as is

node/batcher/requests_inspector_verifier_test.go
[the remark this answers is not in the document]
DorKatzelnick: For now there is only the _payload not empty_ and _max size_ rules, which are tested in the second and third tests.
I'm working on the next PR, which will include the signature filter. It will also check the request structure. This will require additional tests and updates to the existing ones.

HagarMeir (approved):

Issue #278, opened 2025-09-25
Bug - Panic - Batch not found: 329, err: no such block number [329] in index


HagarMeir:
[31m2025-09-21 06:20:09.670 CDT 0a0a ERRO[0m [Batcher3Shard1] [31;1mRetrieveBatchByNumber[0m -> Batch not found: 329, err: no such block number [329] in index
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xc9b6a5]

goroutine 78 [running]:
github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).resubmitPendingBAFs(0xc0003bed80, 0xc000465750?, 0x3)
	/root/fabric-x-orderer/node/batcher/batcher_role.go:218 +0x225
github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).getTermAndNotifyChange(0xc0003bed80)
	/root/fabric-x-orderer/node/batcher/batcher_role.go:202 +0x18f
created by github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).Start in goroutine 1
	/root/fabric-x-orderer/node/batcher/batcher_role.go:128 +0x138

PR #283, opened 2025-09-25
also check equal shard

HagarMeir:
#278


HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/278

PR #289, opened 2025-09-28
batcher config store

MayRosenbaum:
issue #204 #205

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -260,7 +260,8 @@ func (config *Configuration) ExtractRouterConfig(configBlock *common.Block) *nod
 	return routerConfig
 }
 
-func (config *Configuration) ExtractBatcherConfig() *nodeconfig.BatcherNodeConfig {
+func (config *Configuration) ExtractBatcherConfig(configBlock *common.Block) *nodeconfig.BatcherNodeConfig {
+	// TODO: use config block to build bundle in the batcher
 	signingPrivateKey, err := utils.ReadPem(filepath.Join(config.LocalConfig.NodeLocalConfig.GeneralConfig.LocalMSPDir, "keystore", "priv_sk"))
 	if err != nil {
 		panic(fmt.Sprintf("error launching batcher, failed extracting batcher config: %s", err))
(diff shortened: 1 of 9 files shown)


node/batcher/utils_test.go
```diff
@@ -108,12 +110,17 @@ func createBatchers(t *testing.T, num int, shardID types.ShardID, batcherNodes [
 
 		key, err := x509.MarshalPKCS8PrivateKey(batcherNodes[i].sk)
 		signer := crypto.ECDSASigner(*batcherNodes[i].sk)
+		require.NoError(t, err)
+
+		configStore, err := os.MkdirTemp("", fmt.Sprintf("config-store-batcher%d%d", i, shardID))
+		require.NoError(t, err)
```
tock-ibm, line 116: You have to clean it when the test ends... use `t.TempDir` and it will happen automatically

tock-ibm (commented):
Aren't you missing all the part in 
```

func launchBatcher(stop chan struct{}) func(configFile *os.File) {
	return func(configFile *os.File) {
		config, _, err := config.ReadConfig(configFile.Name(), flogging.MustGetLogger("ReadConfigBatcher"))
		if err != nil {
			panic(fmt.Sprintf("error launching batcher, err: %s", err))
		}

		conf := config.ExtractBatcherConfig()
...
```
Where

`conf := config.ExtractBatcherConfig(lastConfigBlock)`

Like in the router, or do you plan to do it in the next PR?

MayRosenbaum:
I don't miss this part.
Actually, the `ExtractRouterConfig` uses the last config block to extract the bundle and the channel ID from. 
The batcher, currently, doesn't extract anything from the block directly, but from the shared config (only BatchingConfig.RequestMaxBytes parameter). 

`ReadConfig` is responsible for bootstrapping from the right block and the corresponding shared config derived from it.
The Extract* methods may use it (like the Router) and may not.

node/batcher/utils_test.go
```diff
@@ -107,13 +107,16 @@ func createBatchers(t *testing.T, num int, shardID types.ShardID, batcherNodes [
...
 		signer := crypto.ECDSASigner(*batcherNodes[i].sk)
 
-		require.NoError(t, err)
+		configStore := t.TempDir()
+
 		conf := &config.BatcherNodeConfig{
 			Shards:                []config.ShardInfo{{ShardId: shardID, Batchers: batchersInfo}},
 			Consenters:            consentersInfo,
 			Directory:             t.TempDir(),
+			ConfigStorePath:       configStore,
```
HagarMeir, line 119: just call t.TempDir() here like we do with Directory the line before this one

config/config.go
```diff
@@ -234,7 +234,8 @@ func (config *Configuration) ExtractRouterConfig(configBlock *common.Block) *nod
 	return routerConfig
 }
 
-func (config *Configuration) ExtractBatcherConfig() *nodeconfig.BatcherNodeConfig {
+func (config *Configuration) ExtractBatcherConfig(configBlock *common.Block) *nodeconfig.BatcherNodeConfig {
+	// TODO: use config block to build bundle in the batcher
```
HagarMeir, line 238: is there an issue for this?

HagarMeir (approved):

Issue #293, opened 2025-10-05
Add bundle and sig filter to batcher


DorKatzelnick:
A bundle needs to be added to the batcherNodeConfig. This bundle should be constructed from the config block and include the policy manager and channel ID.
Additionally, a signature filter should be added to the batcher's RulesVerifier, which will use this bundle.

PR #296, opened 2025-10-12
add bundle and sig filter to batcher

DorKatzelnick:
- add bundle to batcher node config
- add sig filter to request verifier in batcher
- fix tx-test-utils, so that payload bytes in created requests will be deterministic, and avoid shard-mapping discrepancy 
- fix related tests

related issue: #293

### DIFF
--- a/node/batcher/requests_inspector_verifier_test.go
+++ b/node/batcher/requests_inspector_verifier_test.go
@@ -9,11 +9,15 @@ package batcher_test
 import (
 	"testing"
 
+	policyMocks "github.com/hyperledger/fabric-x-orderer/common/policy/mocks"
+	"github.com/hyperledger/fabric-x-orderer/common/types"
 	"github.com/hyperledger/fabric-x-orderer/node/batcher"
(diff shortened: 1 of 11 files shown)


node/batcher/requests_inspector_verifier_test.go
```diff
@@ -21,56 +26,67 @@ import (
...
 	config := &config.BatcherNodeConfig{
 		ShardId:         1,
-		Shards:          []config.ShardInfo{{ShardId: 1}, {ShardId: 2}},
+		Shards:          []config.ShardInfo{{ShardId: types.ShardID(1)}, {ShardId: types.ShardID(2)}},
 		BatchMaxSize:    3,
-		BatchMaxBytes:   10,
-		RequestMaxBytes: 10,
+		BatchMaxBytes:   180,
+		RequestMaxBytes: 60,
+		Bundle:          bundle,
```
HagarMeir: can we add a simple test to check if the sig filter works?

HagarMeir (approved):

PR #304, opened 2025-10-19
change request verifier to return request type

DorKatzelnick:
- change the verifier so that it also returns the type of a request (data / config / etc.. )
- update the sig-filter, to extract the request type from the channel header


common/requestfilter/rulesverifier.go
```diff
@@ -7,15 +7,17 @@ SPDX-License-Identifier: Apache-2.0
...
 
 	"github.com/hyperledger/fabric-x-orderer/node/protos/comm"
 )
 
 // Verifier is a struct that holds a slice of rules, used to verify a request.
 type RulesVerifier struct {
-	rules []Rule
-	lock  sync.RWMutex
+	rules          []Rule
+	structureRules []StructureRule
```
tock-ibm: Are we expecting more than one?

Issue #81, opened 2025-07-09
Router - Envelope to Request 


tock-ibm:
1. comm.Request is an extension of common.Envelope
2. proto message bytes are not deterministic, and should not be used to compute a digest or sign on

- [ ] Make sure in a test that they are always compatible
- [ ] Limit transitions from one to next
- [ ] Check again that we compute the digest on bytes that are not a result of proto.Marshal

PR #314, opened 2025-10-26
batched request has only Envelope bytes

DorKatzelnick:
- make sure batched request has only Envelope bytes
- add a test that verifies it

issue: #81

### DIFF
--- a/node/batcher/batcher.go
+++ b/node/batcher/batcher.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-protos-go-apiv2/orderer"
 	"github.com/hyperledger/fabric-x-orderer/common/configstore"
 	"github.com/hyperledger/fabric-x-orderer/common/types"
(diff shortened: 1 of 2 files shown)


node/batcher/batcher.go
```diff
@@ -158,15 +158,15 @@ func (b *Batcher) Submit(ctx context.Context, req *protos.Request) (*protos.Subm
 	}
 
 	b.Metrics.routerTxsTotal.Add(1)
-	traceId := req.TraceId
-	req.TraceId = nil
-	rawReq, err := proto.Marshal(req)
+
+	// make sure batched requests have only bytes of Envelope, not Request.
```
tock-ibm: // Make sure batched requests have only bytes of Envelope, not Request.
// We do this in order to maintain compatibility with Fabric block structure.

node/batcher/batcher.go
```diff
@@ -158,15 +158,15 @@ func (b *Batcher) Submit(ctx context.Context, req *protos.Request) (*protos.Subm
...
+	rawReq, err := proto.Marshal(&protos.Request{Payload: req.Payload, Signature: req.Signature})
```
tock-ibm: This looks good, but why not use the `&Envelope` directly?

Issue #185, opened 2025-08-18
Metrics with Prometheus


tock-ibm:
- [x] Config for Prometheus - address, port
- [x] Infrastructure in router
- [x] Infrastructure in assembler
- [x] Infrastructure in batcher
- [x] Infrastructure in consenter
- [ ] Extend metrics #414

PR #319, opened 2025-10-28
Config for monitoring

gengur:
#185

### DIFF
--- a/test/utils_test.go
+++ b/test/utils_test.go
@@ -99,11 +99,13 @@ func createRouters(t *testing.T, num int, batcherInfos []nodeconfig.BatcherInfo,
 		bundle.ConfigtxValidatorReturns(configtxValidator)
 
 		config := &nodeconfig.RouterNodeConfig{
-			ListenAddress:      "0.0.0.0:0",
-			ConfigStorePath:    t.TempDir(),
-			TLSPrivateKeyFile:  kp.Key,
-			TLSCertificateFile: kp.Cert,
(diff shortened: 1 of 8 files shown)


test/utils_test.go
```diff
@@ -99,11 +99,13 @@ func createRouters(t *testing.T, num int, batcherInfos []nodeconfig.BatcherInfo,
 		bundle.ConfigtxValidatorReturns(configtxValidator)
 
 		config := &nodeconfig.RouterNodeConfig{
-			ListenAddress:      "0.0.0.0:0",
-			ConfigStorePath:    t.TempDir(),
-			TLSPrivateKeyFile:  kp.Key,
-			TLSCertificateFile: kp.Cert,
-			PartyID:            types.PartyID(i + 1),
+			ListenAddress:           "0.0.0.0:0",
+			MonitoringListenAddress: "0.0.0.0:0",
```
tock-ibm: It is much better/safer to bind to `127.0.0.1` in tests, 0.0.0.0 is bind all, which is listening to all external comm

node/router/router.go
```diff
@@ -279,7 +252,6 @@ func (r *Router) getShardRouterAndReqID(req *protos.Request) ([]byte, *ShardRout
 }
 
 func (r *Router) Submit(ctx context.Context, request *protos.Request) (*protos.SubmitResponse, error) {
-	atomic.AddUint64(&r.incoming, 1)
```
tock-ibm: why is this removed?

Issue #192, opened 2025-08-19
Dissemination: Service for submitting config to consenter, getting a response after validation


tock-ibm:
- The router sends the config TX to the consenter of its own party. It waits for a response, which comes after the consenter validates the TX. 
- The consenter receives a config TX, processes it, and provides an answer to the router.

This can be done by reusing the fabric Broadcast service. 
The consenter exposes the gRPC service; the router uses the Broadcast client to submit the TX.
The consenter accepts submissions only from the router of its own party. 
Use the client TLS cert to verify the router's identity.
The consenter already implements "Broadcast()" (since it fully implements Deliver() to deliver decisions), it just always return an error now.

PR #326, opened 2025-10-30
config submitter part 3

DorKatzelnick:
- check request type in stream, and then forward to batcher or config-submitter
- add config submitter tests to router
- fix some other tests

issue: https://github.com/hyperledger/fabric-x-orderer/issues/192

### DIFF
--- a/node/router/stream.go
+++ b/node/router/stream.go
@@ -35,7 +35,7 @@ type stream struct {
 	srReconnectChan                   chan reconnectReq
 	notifiedReconnect                 bool
 	verifier                          *requestfilter.RulesVerifier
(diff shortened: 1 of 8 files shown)


node/router/stream.go
```diff
@@ -83,26 +83,29 @@ func (s *stream) sendRequests() {
 				continue
 			}
 
-			// TODO: forward request to batcher or consenter
-			_ = reqType
+			if reqType == common.HeaderType_CONFIG_UPDATE {
+				s.logger.Debugf("received request with type %s, forwarding to consenter", reqType)
```
HagarMeir: This can be info and not debug if it is only for config requests

PR #361, opened 2025-11-13
log only when err in not nil

MayRosenbaum:


node/router/shard_router.go
```diff
@@ -318,7 +318,9 @@ func (sr *ShardRouter) initConnPoolAndStreams() {
 
 	// try to initialize the connections and streams. if one fail, the reconnection routine will take care of it when requests arrive
 	err := sr.fillConnPool()
-	sr.logger.Errorf("Initial connection pool fill resulted in error: %v", err)
+	if err != nil {
```
HagarMeir: can you please just do this in one line? calling the func and checking the error

Issue #345, opened 2025-11-10
Apply Admin policy for config TX


tock-ibm:
- [x] In router
- [x] In consenter

PR #376, opened 2025-11-17
verification and authorization in consenter

MayRosenbaum:
issue #377, #345

### DIFF
--- a/node/consensus/consensus.go
+++ b/node/consensus/consensus.go
@@ -618,7 +618,8 @@ func (c *Consensus) verifyCE(req []byte) (smartbft_types.RequestInfo, *state.Con
 		if err != nil {
 			return reqID, ce, errors.Wrapf(err, "failed to verify and classify request")
 		}
-		return smartbft_types.RequestInfo{}, nil, nil
+		// TODO: revisit this return
+		return reqID, ce, nil
 	} else {
 		return smartbft_types.RequestInfo{}, ce, fmt.Errorf("empty control event")
 	}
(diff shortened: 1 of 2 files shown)


node/consensus/consensus.go
```diff
@@ -600,7 +611,14 @@ func (c *Consensus) verifyCE(req []byte) (smartbft_types.RequestInfo, *state.Con
...
 	} else if ce.ConfigRequest != nil {
-		return reqID, ce, nil // TODO verify sig over config
+		_, err := c.verifyAndClassifyRequest(&protos.Request{
+			Payload:   ce.ConfigRequest.Envelope.Payload,
+			Signature: ce.ConfigRequest.Envelope.Signature,
+		})
+		if err != nil {
+			return reqID, ce, errors.Wrapf(err, "failed to verify and classify request")
+		}
+		return smartbft_types.RequestInfo{}, nil, nil
```
tock-ibm: This is an empty return.... if it is in preparation for the next stage put a TODO with an explanation

node/consensus/consensus.go
```diff
@@ -627,3 +645,34 @@ func (c *Consensus) validateRouterFromContext(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (c *Consensus) verifyAndClassifyRequest(request *protos.Request) (*protos.Request, error) {
+	var reqType common.HeaderType
+
+	// Do we need to check the first step anyway or it depends on the flag?
```
tock-ibm: check anyway, we always check for config requests, also in the router.

node/consensus/consensus.go
HagarMeir: please leave the TODO but in the meantime return reqID, ce, nil

node/consensus/consensus.go
```diff
@@ -627,3 +646,31 @@ func (c *Consensus) validateRouterFromContext(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (c *Consensus) verifyAndClassifyRequest(request *protos.Request) (*protos.Request, error) {
+	var reqType common.HeaderType
+
+	err := c.RequestVerifier.Verify(request)
+	if err != nil {
+		c.Logger.Debugf("request is invalid: %s", err)
```
HagarMeir: I don't know if we need both log and returning error 
and please use one line if you just check the error

MayRosenbaum: logs are in debug
I made it one line

PR #378, opened 2025-11-18
Feature/production docker image

anushka159-a:
I have modified the current Dockerfile to be able to generate a Production grade Docker Image based on UBI, the new Dockerfile follows the below good practices for Production.

- Multi-stage build: keeps the final image small and clean.
- Dependency caching
- Copies go.mod + go.sum first and runs go mod download.
- Improves build speed for incremental changes.
- UBI-based runtime: using ubi-minimal is the correct production choice.
- Non-root execution: creates appuser (uid 10001) and runs the service as non-root.
- Minimal runtime dependencies: no extra tools/packages installed (keeps attack surface small).
- Metadata labels: provides traceability (name, version, maintainer, vendor, etc.).
- Volumes for writable dirs: /workspace, /logs, /run defined.
- Deterministic build: dependencies are pulled in the build stage, final image only has the binary.
- Absolute path: in ENTRYPOINT: safer for production.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,6 @@
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
 .github


HagarMeir:
please sign your commit message with `git commit -s`

Issue #193, opened 2025-08-19
Dissemination: Consenter includes config block in a decision


tock-ibm:
The consenter will include the new config block in a decision proposal.
In preparing the proposal, the consenter will apply the logic specified in the Issue XXX.
The config block will be the last in a potential list of available blocks (BAs). 

It is proposed to alter the structure of the AvailableBlocks array to include fabric common.Block rather than the custom structure currentyl there. A block that corresponds to a BA will include only the header and metadata, whereas a config block will include data as well.

A decision that marks a config block must be marked so that other node would easily identify it without sifting through the available blocks. It is proposed to use the "Last Config Index" of the fabric block that carries the decision in order to point to the last decision that carries a config. If a decision carries a config, the last config index points to the same block. The config block would then reside as the last item of the available blocks array.

PR #384, opened 2025-11-19
test submit config consensus node

MayRosenbaum:
issue #193

### DIFF
--- a/node/consensus/consensus.go
+++ b/node/consensus/consensus.go
@@ -161,10 +161,6 @@ func (c *Consensus) SubmitConfig(ctx context.Context, request *protos.Request) (
 		return nil, errors.Wrapf(err, "failed to verify and classify request")
 	}
 
-	if configRequest == nil {
-		return nil, errors.Errorf("unexpected config request was verified and proposed")
-	}
-
 	ce := &state.ControlEvent{
 		ConfigRequest: &state.ConfigRequest{Envelope: &common.Envelope{
 			Payload:   configRequest.Payload,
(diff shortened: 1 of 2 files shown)


node/consensus/consensus.go
```diff
@@ -160,9 +160,16 @@ func (c *Consensus) SubmitConfig(ctx context.Context, request *protos.Request) (
 	if err != nil {
```
HagarMeir: maybe also make sure here that `configRequest != nil`

node/consensus/consensus.go
```diff
@@ -661,7 +668,9 @@ func (c *Consensus) verifyAndClassifyRequest(request *protos.Request) (*protos.R
 		return nil, fmt.Errorf("request structure verification error: %s", err)
 	}
 
-	if reqType != common.HeaderType_CONFIG_UPDATE {
+	// if the request comes from the Router then we expect reType = HeaderType_CONFIG_UPDATE
+	// if the request comes from the Consensus leader then we expect reType = HeaderType_CONFIG
```
HagarMeir: reqType

node/consensus/consensus.go
```diff
@@ -160,9 +160,20 @@ func (c *Consensus) SubmitConfig(ctx context.Context, request *protos.Request) (
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to verify and classify request")
 	}
-	_ = configRequest
-	// TODO: submit configRequest.Bytes()
-	return &protos.SubmitResponse{Error: "SubmitConfig is not implemented in consenter", TraceId: request.TraceId}, nil
+
+	if configRequest == nil {
+		return nil, errors.Errorf("unexpected config request was verified and proposed")
+	}
```
tock-ibm: In line 152 above you are verifying the request an generating a `configRequest`. I expect the method to never return a nil object unless `err!=nil`.
This check is redundant, or at least push it inside `verifyAndClassifyRequest`.

PR #397, opened 2025-11-26
Implement soft stop in batcher

moradna:
#369


node/batcher/batcher.go
HagarMeir: maybe you should add a check if stopped here

PR #425, opened 2025-12-07
test the verification of a config tx

MayRosenbaum:
issue #332

### DIFF
--- a/node/crypto/signer.go
+++ b/node/crypto/signer.go
@@ -10,7 +10,6 @@ import (
 	"crypto/ecdsa"
 	"crypto/rand"
 	"encoding/asn1"
-	"fmt"
 	"math/big"
 
 	"github.com/hyperledger/fabric-lib-go/bccsp/utils"
@@ -25,8 +24,9 @@ func (s ECDSASigner) Sign(message []byte) ([]byte, error) {
 	return signECDSA(&sk, digest)
 }
 
+// Serialize is called when a SignatureHeader.Creator is created. Since this creator is placeholder, the SignatureHeader.Creator must be updated with correct creator.
 func (s ECDSASigner) Serialize() ([]byte, error) {
-	return nil, fmt.Errorf("not implemented")
+	return []byte("creator"), nil
 }
 
 func signECDSA(k *ecdsa.PrivateKey, digest []byte) (signature []byte, err error) {
--- a/test/config_disseminate_test.go
+++ b/test/config_disseminate_test.go
@@ -8,9 +8,25 @@ package test
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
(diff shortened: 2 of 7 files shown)


test/config_disseminate_test.go
```diff
@@ -153,3 +169,196 @@ func TestConfigDisseminate(t *testing.T) {
...
+	defer gexec.CleanupBuildArtifacts()
+	require.NoError(t, err)
+	require.NotNil(t, armaBinaryPath)
+
+	// Generate the configuration with clientSignatureVerificationRequired = True
+	dir := t.TempDir()
+	configPath := filepath.Join(dir, "config.yaml")
+	numOfParties := 4
+	numOfShards := 2
+	orgNum := 1
```
tock-ibm: This name is a bit confusing as it is actually the id of the party which submits the TX

test/config_disseminate_test.go
```diff
@@ -153,3 +169,196 @@ func TestConfigDisseminate(t *testing.T) {
...
+	// Create an admin signer serves as the client which signs the config transaction
+	keyPath := filepath.Join(dir, "crypto", "ordererOrganizations", fmt.Sprintf("org%d", orgNum), "users", "admin", "msp", "keystore", "priv_sk")
+	signer, err := CreateSigner(keyPath)
+	require.NoError(t, err)
+	require.NotNil(t, signer)
+
+	certPath := filepath.Join(dir, "crypto", "ordererOrganizations", fmt.Sprintf("org%d", orgNum), "users", "admin", "msp", "signcerts", fmt.Sprintf("Admin@Org%d-cert.pem", orgNum))
+	certBytes, err := os.ReadFile(certPath)
+	require.NoError(t, err)
+	require.NotNil(t, certBytes)
```
tock-ibm: there is a bit of a duplication here, as all this is also done within `createConfigTX`. instead you could have just just indicated to  `createConfigTX` which admin id submits, and thus signs the TX itself, no?

node/crypto/signer.go
```diff
@@ -25,8 +24,9 @@ func (s ECDSASigner) Sign(message []byte) ([]byte, error) {
 	return signECDSA(&sk, digest)
 }
 
+// Serialize is called when a SignatureHeader.Creator is created. Since this creator is placeholder, the SignatureHeader.Creator must be updated with correct creator.
 func (s ECDSASigner) Serialize() ([]byte, error) {
-	return nil, fmt.Errorf("not implemented")
+	return []byte("creator"), nil
```
tock-ibm: I seems we never actually use this implementation

tock-ibm (approved):
Approved with some suggestions for a future PR.

node/crypto/signer.go
MayRosenbaum: I added a comment above the function that explains when this method is called.

MayRosenbaum: This is method in which the Serialize() is called:
```
func NewSignatureHeader(id identity.Serializer) (*cb.SignatureHeader, error) {
	creator, err := id.Serialize()
	if err != nil {
		return nil, err
	}
	nonce, err := CreateNonce()
	if err != nil {
		return nil, err
	}

	return &cb.SignatureHeader{
		Creator: creator,
		Nonce:   nonce,
	}, nil
}

```

PR #455, opened 2025-12-11
add config sequence to router and requests

DorKatzelnick:
add configSeq to router and requests forwarded from router to batcher

issues: #450 , #451


test/utils_test.go
```diff
@@ -502,7 +502,7 @@ func recoverRouter(conf *node_config.RouterNodeConfig, logger *zap.SugaredLogger
 	configUpdateProposer := &policyMocks.FakeConfigUpdateProposer{}
 	configUpdateProposer.ProposeConfigUpdateReturns(nil, nil)
 
-	router := router.NewRouter(conf, logger, fakeSigner, configUpdateProposer)
+	router := router.NewRouter(conf, 0, logger, fakeSigner, configUpdateProposer)
```
HagarMeir: it will work for now but here the recover is used after a reconfig in the test so the config seq should change as well

node/router/router.go
```diff
@@ -110,7 +111,7 @@ func NewRouter(config *nodeconfig.RouterNodeConfig, logger types.Logger, signer
 
 	metrics := NewRouterMetrics(config, logger)
 
-	r := createRouter(shardIDs, batcherEndpoints, tlsCAsOfBatchers, metrics, config, logger, verifier, configStore, configSubmitter, configPuller)
+	r := createRouter(shardIDs, batcherEndpoints, tlsCAsOfBatchers, metrics, config, configSeq, logger, verifier, configStore, configSubmitter, configPuller)
```
tock-ibm: sequence in the config

node/config/config.go
```diff
@@ -92,6 +92,8 @@ type RouterNodeConfig struct {
 	Bundle                  channelconfig.Resources
 	MonitoringListenAddress string
 	MetricsLogInterval      time.Duration
+	// ConfigSeq is the number of the last of config block.
```
HagarMeir: it is not
it is the sequence of the config blocks
or how many config blocks were there 
this is how we reach it in the consenter:

`func (c *Consensus) VerificationSequence() uint64 {
	return c.Config.Bundle.ConfigtxValidator().Sequence()
}`

PR #459, opened 2025-12-14
Improve assembler ledger API

tock-ibm:
AppendConfig only needs the OrderingInfo
AppendConfig organizes the signatures in the block metadata

#403

### DIFF
--- a/node/ledger/assembler_ledger.go
+++ b/node/ledger/assembler_ledger.go
@@ -33,7 +33,7 @@ type AssemblerLedgerReaderWriter interface {
 	GetTxCount() uint64
 	Metrics() *AssemblerLedgerMetrics
 	Append(batch types.Batch, orderingInfo *state.OrderingInformation)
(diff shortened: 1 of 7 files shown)


node/ledger/assembler_ledger.go
```diff
@@ -189,36 +176,50 @@ func (l *AssemblerLedger) Append(batch types.Batch, ordInfo *state.OrderingInfor
...
+
+	configBlock := orderingInfo.CommonBlock
 	if !protoutil.IsConfigBlock(configBlock) {
 		l.Logger.Panicf("attempting to AppendConfig a block which is not a config block: %d", configBlock.GetHeader().GetNumber())
 	}
 
+	// TODO update the signature on the block in consensus
+	sigs, signers := arrangeSignatures(orderingInfo)
+
+	if len(sigs) > 0 {
```
HagarMeir: why add this check? 
this check does not exist in the regular append

Issue #405, opened 2025-11-30
Apply config - Consenter


HagarMeir:
Three phases of implementation:
1. with restart - No party add/remove
2. with restart - With party add/remove
3. dynamic - no restart needed
 

Update the state accordingly:

- [ ] Number of parties / Threshold / Quorum 
- [ ] Complaints
- [ ] Pending BAFs

Update the SmartBFT library

tock-ibm:
Unit tests moved to #854

PR #473, opened 2025-12-21
apply config consensus compute state update parties only

MayRosenbaum:
issue #405 
and https://github.com/hyperledger/fabric-x-orderer/issues/486

### DIFF
--- a/test/utils_test.go
+++ b/test/utils_test.go
@@ -41,6 +41,7 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/node/comm/tlsgen"
 	node_config "github.com/hyperledger/fabric-x-orderer/node/config"
 	"github.com/hyperledger/fabric-x-orderer/node/consensus"
+	consensusMocks "github.com/hyperledger/fabric-x-orderer/node/consensus/mocks"
 	"github.com/hyperledger/fabric-x-orderer/node/consensus/state"
 	"github.com/hyperledger/fabric-x-orderer/node/crypto"
 	protos "github.com/hyperledger/fabric-x-orderer/node/protos/comm"
@@ -94,7 +95,7 @@ func keygen(t *testing.T) (*ecdsa.PrivateKey, []byte) {
 	return sk, rawPK
 }
 
-func createRouters(t *testing.T, num int, batcherInfos []node_config.BatcherInfo, ca tlsgen.CA, shardId types.ShardID, consenterEndpoint string, genesisBlock *common.Block) ([]*router.Router, []node_config.RawBytes, []*node_config.RouterNodeConfig, []*zap.SugaredLogger) {
+func createRouters(t *testing.T, num int, batcherInfos []node_config.BatcherInfo, ca tlsgen.CA, shardId types.ShardID, consenterEndpoint []string, genesisBlock *common.Block) ([]*router.Router, []node_config.RawBytes, []*node_config.RouterNodeConfig, []*zap.SugaredLogger) {
 	var routers []*router.Router
 	var certs []node_config.RawBytes
 	var configs []*node_config.RouterNodeConfig
@@ -138,7 +139,7 @@ func createRouters(t *testing.T, num int, batcherInfos []node_config.BatcherInfo
 			RequestMaxBytes:                     1 << 10,
 			ClientSignatureVerificationRequired: false,
 			Bundle:                              bundle,
-			Consenter:                           node_config.ConsenterInfo{PartyID: types.PartyID(i + 1), Endpoint: consenterEndpoint, TLSCACerts: []node_config.RawBytes{ca.CertBytes()}},
+			Consenter:                           node_config.ConsenterInfo{PartyID: types.PartyID(i + 1), Endpoint: consenterEndpoint[i], TLSCACerts: []node_config.RawBytes{ca.CertBytes()}},
 		}
 		configs = append(configs, config)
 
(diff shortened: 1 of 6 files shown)


node/consensus/consenter.go
HagarMeir: is this used anywhere now? 
the `State` in the `Consenter` struct

node/consensus/consensus_config_applier.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/hyperledger/fabric-lib-go/bccsp/factory"
+	"github.com/hyperledger/fabric-x-common/common/channelconfig"
+	config_protos "github.com/hyperledger/fabric-x-orderer/config/protos"
+	"github.com/hyperledger/fabric-x-orderer/node/consensus/state"
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
+)
+
+//go:generate counterfeiter -o mocks/config_applier.go . ConfigApplier
```
HagarMeir: can we just use the mock instead of having a `NoOpDefaultConfigApplier`?

node/consensus/consenter.go
MayRosenbaum: No, only var assignment
I removed the field

node/consensus/consenter.go
```diff
@@ -58,13 +56,10 @@ func (c *Consenter) SimulateStateTransition(prevState *state.State, requests [][
 // Commit indexes BAs and updates the state.
 // Note that this must hold: Commit(controlEvents) with the same controlEvents is idempotent.
 // TODO revise the recovery from failure or shutdown, specifically the order of Commit and Append.
-func (c *Consenter) Commit(events [][]byte) {
-	state, batchAttestations, _ := c.SimulateStateTransition(c.State, events)
-	// TODO apply config to get a new state
+func (c *Consenter) Commit(state *state.State, batchAttestations [][]types.BatchAttestationFragment) {
```
HagarMeir: is state used here?

node/consensus/consensus.go
```diff
@@ -605,15 +605,15 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
 	// we index, next time we spawn, we will not recognize we did not index and as a result we will may sign
 	// a batch attestation twice.
 	// This is true because a Commit(controlEvents) with the same controlEvents is idempotent.
-	state, batchAttestations, configRequests := c.Arma.SimulateStateTransition(c.State, controlEvents)
+	state, batchAttestations, configRequests := c.Arma.SimulateStateTransition(hdr.State, controlEvents)
 	if configRequests != nil {
 		var err error
 		if state, err = c.ConfigApplier.ApplyConfigToState(state, configRequests[0]); err != nil {
```
HagarMeir: also here I don't think this state is used

node/consensus/consensus.go
```diff
@@ -605,15 +605,15 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
 	// we index, next time we spawn, we will not recognize we did not index and as a result we will may sign
 	// a batch attestation twice.
 	// This is true because a Commit(controlEvents) with the same controlEvents is idempotent.
-	state, batchAttestations, configRequests := c.Arma.SimulateStateTransition(c.State, controlEvents)
+	state, batchAttestations, configRequests := c.Arma.SimulateStateTransition(hdr.State, controlEvents)
```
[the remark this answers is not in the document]
HagarMeir: you can't use hdr.State here
you need the c.State with a lock

node/consensus/consensus_builder.go
```diff
@@ -11,12 +11,13 @@ import (
...
 	"encoding/pem"
 	"fmt"
-	"math"
 	"os"
 	"path/filepath"
 	"sort"
 	"time"
 
+	"github.com/hyperledger/fabric-x-orderer/common/utils"
+
```
HagarMeir: remove line and run go fmt and imports again please

common/utils/state.go
```diff
@@ -0,0 +1,19 @@
...
+*/
+
+package utils
+
+import (
+	"math"
+)
+
+// ComputeFTQ computes the F, Threshold and Quorum corresponds to N.
+func ComputeFTQ(n uint16) (uint16, uint16, uint16) {
```
tock-ibm: Unit test please

test/utils_test.go
```diff
@@ -267,6 +268,7 @@ func createConsenters(t *testing.T, num int, consenterNodes []*node, consenterIn
 		mockConfigUpdateProposer.ProposeConfigUpdateReturns(nil, nil)
 
 		c := consensus.CreateConsensus(conf, net, genesisBlock, logger, signer, mockConfigUpdateProposer)
+		c.ConfigApplier = &NoOpDefaultConfigApplier{}
```
tock-ibm: Why not a real one? is it because it is used only for abcr_test and config_dissemination_test, which is based on it?

node/consensus/consensus_config_applier.go
```diff
@@ -0,0 +1,51 @@
...
+	if !exists {
+		return nil, errors.Wrapf(err, "orderer entry in the config block is empty")
+	}
+
+	consensusMetadata := ordererConfig.ConsensusMetadata()
+	sharedConfig := &config_protos.SharedConfig{}
+	err = proto.Unmarshal(consensusMetadata, sharedConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal consensus metadata to a shared configuration")
+	}
```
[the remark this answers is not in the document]
MayRosenbaum: Correct, in a later PR

test/utils_test.go
MayRosenbaum: It is used in the `abcr_test`, `config_dissemination_test` and `batcher_consenter_test`
Maybe we should consider moving these tests to another package, as mocks are used.

The configApplier is mocked since the real method calls to the following function
`bundle, err := channelconfig.NewBundleFromEnvelope(env....)`
The function is expected to fail when the tx is not real.
In the `config_dissemination_test` we send a fake config transaction, so we need the mock.
The test only checks successful dissemination (no validity, policy....)

PR #478, opened 2025-12-23
Create config moved to the testutils/configutil package

gengur:


go.mod
```diff
@@ -4,6 +4,7 @@ go 1.24.3
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/gogo/protobuf v1.3.2
```
tock-ibm: why are we using gogo protobuf? it is a mistake in the import

testutil/configutil/config_update_utils.go
```diff
@@ -0,0 +1,241 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/gogo/protobuf/proto"
```
tock-ibm: This is an error, needs to be google's protobuf lib

Issue #485, opened 2026-01-04
Consenter apply config with restart - no party change


HagarMeir:
**Note**: In a decision that includes a config change the pending BAFs (with old config sqn) are saved. The batchers should process these batches while processing the config change.

- [x] The consenter will clean said expired BAFs in the next decision.
- [x] Complaints with an expired config sqn are dropped from the state. (Batchers in dynamic reconfig should handle that).
- [x] Reject incoming control events with mismatched config sqn in them.
- [x] When the config change includes party change, the term is increased to make the primary change explicit. 

**Note**:
To shutdown the entire cluster safely, one must flush the system from pending TXs.

PR #513, opened 2026-01-08
use config seq in simulate state

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/485

### DIFF
--- a/node/consensus/state/state.go
+++ b/node/consensus/state/state.go
@@ -24,6 +24,7 @@ import (
 type Rule func(*State, types.ConfigSequence, types.Logger, ...ControlEvent)
 
 var Rules = []Rule{
+	FilterPendingEventsWithDiffConfigSeq,
 	CollectAndDeduplicateEvents,
 	// DetectEquivocation, // TODO: false positive, lets find out why
 	PrimaryRotateDueToComplaints,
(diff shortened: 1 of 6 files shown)


node/consensus/state/state.go
```diff
@@ -21,9 +21,10 @@ import (
 	"github.com/hyperledger/fabric/protoutil"
 )
 
-type Rule func(*State, types.Logger, ...ControlEvent)
+type Rule func(*State, types.ConfigSequence, types.Logger, ...ControlEvent)
```
tock-ibm: Would it not make sense to include the config sequence in the state?

node/consensus/state/state.go
```diff
@@ -21,9 +21,10 @@ import (
...
 
 var Rules = []Rule{
+	FilterPendingEventsWithDiffConfigSeq,
```
tock-ibm: Only this rule needs the config sequence, and yet because of the Rule array pattern we are forced to give it to all rules. 
I have to admit I am not a big fan of the rule table pattern, I'd rather have those as member functions of the state, called one by one. IMO, it's much easier to follow.

node/consensus/state/state.go
HagarMeir: I think it is easier this way 
And don't think we need to save it in the state

PR #534, opened 2026-01-19
mTLS from client which is not specified in local config ClientRootCAs

gengur:
issue #532

### DIFF
--- a/test/router_test.go
+++ b/test/router_test.go
@@ -145,7 +145,7 @@ func TestRouterRestartRecover(t *testing.T) {
 	routerToStop.StopArmaNode()
 	gotError := false
 
-	for i := 0; i < totalTxNumber; i++ {
+	for i := range totalTxNumber {
 		status := rl.GetToken()
 		if !status {
 			fmt.Fprintf(os.Stderr, "failed to send tx %d", i+1)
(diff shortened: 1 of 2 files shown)


test/router_test.go
```diff
@@ -479,3 +479,113 @@ func TestVerifySignedTxsByRouterSingleParty(t *testing.T) {
...
+	uc.TLSPrivateKey = fakeClientCertKeyPair.Key
+	uc.TLSCertificate = fakeClientCertKeyPair.Cert
+
+	broadcastClientInvalid := client.NewBroadcastTxClient(uc, 10*time.Second)
+	defer broadcastClientInvalid.Stop()
+
+	txContent := tx.PrepareTxWithTimestamp(0, 64, []byte("sessionNumber"))
+	env := tx.CreateStructuredEnvelope(txContent)
+	err = broadcastClientInvalid.SendTx(env)
+	require.ErrorContains(t, err, "no available streams to send the transaction")
```
MayRosenbaum: I would expect another error like the one we get with the assembler.
Actually, you created a new user config, but didn't create a new corresponding broadcast client. I would expect a new broadcast client to be unable to create connections to the router.

PR #535, opened 2026-01-19
commit digests instead of bas

HagarMeir:


node/consensus/consensus.go
```diff
@@ -601,24 +601,23 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
 		c.Logger.Panicf("Failed deserializing header: %v", err)
 	}
 
-	var controlEvents arma_types.BatchedRequests
-	if err := controlEvents.Deserialize(proposal.Payload); err != nil {
-		c.Logger.Panicf("Failed deserializing proposal payload: %v", err)
+	var digests [][]byte
```
tock-ibm: You know how many AvailableCommonBlocks there are, so it makes sense to allocate a slice with that capacity.
`digests := make([]byte, 0, len(hdr.AvailableCommonBlocks))`

Issue #502, opened 2026-01-07
Batcher apply config with restart - no party change


HagarMeir:
- [x] Resubmit expired BAFs (the pending BAFs of the config update decision)
- [x] Store the last decision number fully processed in a WAL
- [x] Start to pull decisions from the last decision processed (deduced from the WAL)

After getting a decision with config, the config block is saved to the store.
Then soft stop.

After restart: 
Start from the last decision processed (not newest, can be deduced from config store or WAL), and reapply; in the current case re-pull the last config decision, and (append will be no-op) process the stale BAFs.

Stale BAFs signed by the same batcher should be read from the ledger and TXs added to mempool.

In general batcher should persist the last decision number fully processed and start from there.
A WAL can be used. Since 
- in normal operation it only appends and prunes;
- in recovery it gets the last item.
Check whether we can reuse the SmartBFT WAL.

This can also be used in the router to inform the restarting router from which decision to pull.

Note that without party change there will be no term change.

**NOTE** (dynamic only)
Pulling batches from a primary and pulling configs from consensus are asynch!

So a primary could verify  a batch on conf-sqn K and send a BAF 
Secondaries may pull said batch and verify it with either K, K-1, or K+1; 
We should make sure no TXs are lost in this situation;

Remedy:
The thread that pulls batches, stops on complaint. It is reset on term change. The conclusion is that it should reset and restart on config change as well.

**NOTE**
- [x] The config-sequence is written to the batch locally - this is the config-sqn in which the batch was verified locally and the config-sqn in the corresponding BAF. This is done for tracking and debugging; the remote secondaries ignore the config-sqn received from the primary, and verify with their own. They emit a Warning when there is a mismatch for diagnostics.

PR #566, opened 2026-02-01
add wal path in router and batcher config

MayRosenbaum:
issue #502


tock-ibm:
not needed anymore, the path is directly derived from the storage directory

Issue #565, opened 2026-02-01
TestSubmitStopThenRestartConsenter test flake


MayRosenbaum:
--- FAIL: TestSubmitStopThenRestartConsenter (97.41s)
    utils_test.go:791: Pulling from party: 2
    utils_test.go:791: Pulling from party: 3
    utils_test.go:791: Pulling from party: 1
    utils_test.go:791: Pulling from party: 4
    utils_test.go:793: Finished pull and count: blocks 11, txs 1001 from party: 2
    utils_test.go:793: Finished pull and count: blocks 11, txs 1001 from party: 3
    utils_test.go:793: Finished pull and count: blocks 11, txs 1001 from party: 1
    utils_test.go:793: Finished pull and count: blocks 9, txs 919 from party: 4
    utils_test.go:641: 
        	Error Trace:	/home/runner/work/fabric-x-orderer/fabric-x-orderer/test/utils_test.go:641
        	            				/opt/hostedtoolcache/go/1.25.5/x64/src/runtime/asm_amd64.s:1693
        	Error:      	"919" is not greater than or equal to "1000"
        	Test:       	TestSubmitStopThenRestartConsenter
    utils_test.go:791: Pulling from party: 2
    utils_test.go:791: Pulling from party: 4
    utils_test.go:791: Pulling from party: 1
    utils_test.go:793: Finished pull and count: blocks 17, txs 1501 from party: 4
    utils_test.go:793: Finished pull and count: blocks 17, txs 1501 from party: 2
    utils_test.go:793: Finished pull and count: blocks 17, txs 1501 from party: 1
    utils_test.go:791: Pulling from party: 4
    utils_test.go:791: Pulling from party: 3
    utils_test.go:791: Pulling from party: 2
    utils_test.go:791: Pulling from party: 1
    utils_test.go:793: Finished pull and count: blocks 23, txs 2001 from party: 4
    utils_test.go:793: Finished pull and count: blocks 23, txs 2001 from party: 2
    utils_test.go:793: Finished pull and count: blocks 23, txs 2001 from party: 1
    utils_test.go:793: Finished pull and count: blocks 11, txs 1001 from party: 3
    utils_test.go:641: 
        	Error Trace:	/home/runner/work/fabric-x-orderer/fabric-x-orderer/test/utils_test.go:641
        	            				/opt/hostedtoolcache/go/1.25.5/x64/src/runtime/asm_amd64.s:1693
[...]

PR #595, opened 2026-02-08
add sleep at start to avoid flakes

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/565

### DIFF
--- a/test/consensus_test.go
+++ b/test/consensus_test.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/hyperledger/fabric-x-orderer/common/tools/armageddon"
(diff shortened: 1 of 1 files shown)


test/consensus_test.go
```diff
@@ -56,6 +57,8 @@ func TestSubmitStopThenRestartConsenter(t *testing.T) {
 
 	testutil.WaitReady(t, readyChan, len(netInfo), 10)
 
+	time.Sleep(10 * time.Second)
```
tock-ibm: If you need to add sleeps here it means that the `testutil.WaitReady(t, readyChan, len(netInfo), 10)` is not catching the right trigger to start the test safely.

HagarMeir: you are correct
however it is not easy to catch the right trigger as we are not sure yet what it is 
so we will leave the issue open and try to fix it in a different PR

PR #607, opened 2026-02-11
Update fabric x common dependency

MayRosenbaum:


common/deliver/acl.go
```diff
@@ -30,12 +32,16 @@ func NewSessionAC(chain ConfigSequencer, env *common.Envelope, policyChecker Pol
...
+	if err != nil {
+		return nil, err
+	}
 	return &SessionAccessControl{
 		envelope:       env,
 		channelID:      channelID,
 		sequencer:      chain,
 		policyChecker:  policyChecker,
-		sessionEndTime: expiresAt(signedData[0].Identity),
+		sessionEndTime: expiresAt(id),
```
tock-ibm: github.com/hyperledger/fabric-x-common/common/crypto/expiration.go
Line 22

does this work with an identity that is using a cert-hash rather than a cert?

Here you marshal msppb.Identity 
but at fabric-x-common they are unmarshaling to 
```

// ExpiresAt returns when the given identity expires, or a zero time.Time
// in case we cannot determine that
func ExpiresAt(identityBytes []byte) time.Time {
	sId := &msp.SerializedIdentity{}
	// If protobuf parsing failed, we make no decisions about the expiration time
	if err := proto.Unmarshal(identityBytes, sId); err != nil {
		return time.Time{}
	}
	return certExpirationTime(sId.IdBytes)
}
```
and parsing a PEM for the IdBytes?

common/deliver/deliver_test.go
```diff
@@ -13,22 +13,23 @@ import (
...
-	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
 	ab "github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	"google.golang.org/protobuf/types/known/timestamppb"
+
 	"github.com/hyperledger/fabric-x-common/common/crypto/tlsgen"
+	"github.com/hyperledger/fabric-x-common/common/deliver"
+	"github.com/hyperledger/fabric-x-common/common/deliver/mock"
```
tock-ibm: you have to take the deliver package from the orderer, we copy it here

common/deliver/deliver_test.go
```diff
@@ -13,22 +13,23 @@ import (
...
+	"github.com/hyperledger/fabric-x-common/common/ledger/blockledger"
```
tock-ibm: blockledger from orderer!

tock-ibm (approved):

PR #608, opened 2026-02-11
Arma network changes

gengur:


testutil/utils.go
```diff
@@ -105,38 +110,40 @@ func CreateNetwork(t *testing.T, path string, numOfParties int, numOfBatcherShar
...
+		nodeName = NodeName{PartyID: types.PartyID(i + 1), NodeType: Assembler}
+		netInfo[nodeName] = &ArmaNodeInfo{Listener: lla, NodeType: Assembler, PartyId: types.PartyID(i + 1), RunningOrder: 2}
 	}
 
 	network := genconfig.Network{
 		Parties:         parties,
 		UseTLSRouter:    useTLSRouter,
 		UseTLSAssembler: useTLSAssembler,
-		MaxPartyID:      types.PartyID(maxPartyID),
+		MaxPartyID:      types.PartyID(numOfParties),
```
tock-ibm: This works only on a fresh network where PartyIDs is a sequence without gaps... in theory we can start with {1,3,5,7} ...

testutil/utils.go
```diff
@@ -105,38 +110,40 @@ func CreateNetwork(t *testing.T, path string, numOfParties int, numOfBatcherShar
 
 		parties = append(parties, party)
 
-		if partyID > maxPartyID {
```
tock-ibm: Keep this calculation... since in theory we can start with {1,3,5,7} ...

PR #635, opened 2026-02-22
refactor in router

DorKatzelnick:
refactor in router, prepare for dynamic reconfig

### DIFF
--- a/node/router/config_submitter.go
+++ b/node/router/config_submitter.go
@@ -22,6 +22,7 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/common/types"
 	"github.com/hyperledger/fabric-x-orderer/config/verify"
 	"github.com/hyperledger/fabric-x-orderer/node/comm"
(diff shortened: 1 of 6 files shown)


node/router/config_submitter.go
```diff
@@ -48,6 +49,15 @@ type configSubmitter struct {
 	configRulesVerifier   verify.OrdererRules
 }
 
+func CreateConfigSubmitter(conf *nodeconfig.RouterNodeConfig, logger types.Logger, verifier *requestfilter.RulesVerifier, signer identity.SignerSerializer, configUpdateProposer policy.ConfigUpdateProposer, configRulesVerifier verify.OrdererRules) *configSubmitter {
```
tock-ibm: It seems Create* calls New* and that is it; why not just change the New* ? 
New* is now only use in tests, so you can probably change the test to take Create* or the modified New*, right?
This is confusing...

Issue #35, opened 2025-07-09
Consensus - Synchronizer


tock-ibm:
- [x] Review the code and test the existing synchronizer
- [x] Make the synchronizer BFT, for example we need to pull from more nodes and not from a single node (maybe use the Fabric synchronizer)
- [x] Create the synchronizer like in Fabric - the BFT Sync() recreates a synchronizer that reads the config every time, and a config change updates a running synchronizer as well

See #44 which is a building block

PR #636, opened 2026-02-22
BFT synchronizer with noop-verifier and block-puller

tock-ibm:
The BFT synchronizer, with a noop-block-verifier, and a block puller which functions only as a target-height detector.

#35


node/consensus/synchronizer/block_puller_factory.go
```diff
@@ -7,18 +7,28 @@ SPDX-License-Identifier: Apache-2.0
...
+	"github.com/hyperledger/fabric-x-orderer/config/protos"
 	"github.com/hyperledger/fabric-x-orderer/node/comm"
 	"github.com/pkg/errors"
+	"google.golang.org/protobuf/proto"
 )
 
 //go:generate counterfeiter -o mocks/block_puller.go . BlockPuller
 
-// BlockPuller is used to pull blocks from other OSN
+// BlockPuller is used to pull blocks from other OSN and detect the ledger height of other OSNs.
```
HagarMeir: please don't use OSN

Issue #504, opened 2026-01-07
Batcher apply config - dynamic


HagarMeir:
The batcher will soft stop.
It will notify the consenter via the ConfigAck API. See #641 
It will then analyze whether the new config require it to wait admin restart (e.g. changing endpoint, certificate, or eviction).
If yes, it will remain in soft stop. 
If not, it will apply the incoming config and restart internally. 

Upon internal restart the memory pool is kept, and every TX in it is verified again using the new config. TXs that fail verification with the new config are dropped, the drop event is logged. 

When a batcher does hard restart by admin it loses its mem pool.

### NOTE
Pulling batches from a primary and pulling configs from consensus are asynch!

So a primary could verify a batch on conf-sqn K and send a BAF
Secondaries may pull said batch and verify it with either K, K-1, or K+1;
We should make sure no TXs are lost in this situation;

Remedy:
- [ ] The thread that pulls batches, stops on complaint. It is reset on term change. The conclusion is that it should reset and restart on config change as well.

tock-ibm:
Config Ack in #641 
Dangling BAFs in #810 
Memory pool params & options #799

PR #649, opened 2026-02-25
prepare batcher for dynamic reconfig

MayRosenbaum:
issue #504


node/batcher/batcher_role.go
```diff
@@ -183,6 +185,17 @@ func (b *BatcherRole) Stop() {
 	b.running.Wait()
 }
 
+func (b *BatcherRole) SoftStop() {
```
HagarMeir: add comments

Issue #442, opened 2025-12-09
Router apply config - dynamic


tock-ibm:
The router will enter a soft stop. 
It will notify the consenter via the ConfigAck API.
It will then analyze whether the new config require it to wait admin restart (e.g. changing endpoint, certificate, or eviction).
If yes, it will remain in soft stop. 
If not, it will apply the incoming config and restart internally.

Update incoming Comm stack
Update outgoing Comm stack

tock-ibm:
Config Ack in #641

PR #659, opened 2026-03-03
stop and channel name in router

DorKatzelnick:
change channel name
issue #442

### DIFF
--- a/node/router/router.go
+++ b/node/router/router.go
@@ -61,7 +61,7 @@ type Router struct {
 	stopOnce         sync.Once
 	drainChan        chan struct{}
 	drainOnce        sync.Once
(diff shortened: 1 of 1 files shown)


node/router/router.go
```diff
@@ -337,7 +338,7 @@ func createRouter(shardIDs []types.ShardID, batcherEndpoints map[types.ShardID]s
 		decisionPuller:   decisionPuller,
 		stopChan:         make(chan struct{}),
 		drainChan:        make(chan struct{}),
-		armaStopChan:     armaStopChan,
+		mainExitChan:     armaStopChan,
```
HagarMeir: also change var name?

PR #660, opened 2026-03-03
connect prometheus to assembler deliver

moradna:
#601


node/assembler/metrics.go
```diff
@@ -47,13 +49,18 @@ func NewMetrics(assemblerNodeConfig *config.AssemblerNodeConfig, al *node_ledger
 	p := monitor.Provider
 	al.NewAssemblerLedgerMetrics(p, partyID, logger)
 
+	deliverMetrics := deliver.NewMetrics(p)
+	deliverMetrics.StreamsOpened = deliverMetrics.StreamsOpened.With()
+	deliverMetrics.StreamsClosed = deliverMetrics.StreamsClosed.With()
```
tock-ibm: `With() ` what? if no labels you can drop the With, no?

PR #674, opened 2026-03-11
prepare batcher creator and stopper for dynamic reconfig

MayRosenbaum:
issue #504

### DIFF
--- a/node/batcher/batcher.go
+++ b/node/batcher/batcher.go
@@ -70,16 +70,14 @@ type Batcher struct {
 	signer                             Signer
 	wal                                *smartbft_wal.WriteAheadLogFile
 
-	stateChan chan *state.State
-
+	stopLock             sync.Mutex
+	stateChan            chan *state.State
 	running              sync.WaitGroup // maybe change the name, it is only for state replicator
 	stopChan             chan struct{}
 	stopSignalListenChan chan struct{}
 	mainExitChan         chan struct{}
-
-	stopLock      sync.Mutex
-	isStopped     bool
-	isSoftStopped bool
+	isStopped            bool
+	isSoftStopped        bool
 
 	primaryLock sync.RWMutex
 	term        uint64
--- a/node/batcher/batcher_builder.go
+++ b/node/batcher/batcher_builder.go
@@ -25,9 +25,9 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/request"
 )
 
-func CreateBatcher(config *node_config.BatcherNodeConfig, fullConfig *config.Configuration, logger *flogging.FabricLogger, mainExitChan chan struct{}, cdrc ConsensusDecisionReplicatorCreator, senderCreator ConsenterControlEventSenderCreator, signer Signer) *Batcher {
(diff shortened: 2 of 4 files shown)


node/batcher/batcher_builder.go
```diff
@@ -18,30 +18,39 @@ import (
...
+		batcher:                            &BatcherRole{},
+		consensusDecisionReplicatorCreator: cdrc,
+	}
+
+	b.configureBatcher(logger, mainExitChan, senderCreator, signer, nil)
+	return b
 }
 
-func CreateBatcherWithMemPool(config *node_config.BatcherNodeConfig, logger *flogging.FabricLogger, mainExitChan chan struct{}, cdrc ConsensusDecisionReplicatorCreator, senderCreator ConsenterControlEventSenderCreator, signer Signer, memPool MemPool) *Batcher {
+func (b *Batcher) configureBatcher(logger *flogging.FabricLogger, mainExitChan chan struct{}, senderCreator ConsenterControlEventSenderCreator, signer Signer, memPool MemPool) {
```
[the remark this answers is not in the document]
MayRosenbaum: We decided to implement the reconfig, keeping the same batcher.
The configureBatcher is called in the ApplyConfig()

node/batcher/batcher.go
```diff
@@ -49,30 +50,36 @@ type Net interface {
...
-	running      sync.WaitGroup
-	stopOnce     sync.Once
-	stopChan     chan struct{}
-	mainExitChan chan struct{}
+	running              sync.WaitGroup // maybe change the name, it is only for state replicator
+	stopChan             chan struct{}
+	stopSignalListenChan chan struct{}
+	mainExitChan         chan struct{}
+
+	stopLock      sync.Mutex
```
tock-ibm: This is protecting also the life cycle of the channels above. It is good practice to put all the fields the mutex is protected jut below it.

node/batcher/batcher_role.go
```diff
@@ -129,9 +128,19 @@ type BatcherRole struct {
 	ackerLock               sync.RWMutex
 	acker                   SeqAcker
 	Metrics                 *BatcherMetrics
+
+	lock          sync.Mutex
+	isStopped     bool
+	isSoftStopped bool
```
tock-ibm: Note that you don't really need a lock here to protect `isStopped` and `isSoftStopped`, as `BatcherRole` methods `Stop Start SoftStop` are called from within  `Batcher` methods `Stop Start SoftStop` which do hold the lifecycle lock.
This is not an error, just nested locking, but it is redundant.

node/server/arma.go
```diff
@@ -181,7 +181,7 @@ func launchBatcher(stop chan struct{}) func(configFile *os.File) {
 			batcherLogger = flogging.MustGetLogger(fmt.Sprintf("Batcher%dShard%d", conf.PartyId, conf.ShardId))
 		}
 
-		batcher := batcher.CreateBatcher(conf, batcherLogger, stop, &batcher.ConsensusDecisionReplicatorFactory{}, &batcher.ConsenterControlEventSenderFactory{}, signer)
+		batcher := batcher.CreateBatcher(conf, config, batcherLogger, stop, &batcher.ConsensusDecisionReplicatorFactory{}, &batcher.ConsenterControlEventSenderFactory{}, signer)
```
tock-ibm: We can later push the extraction of conf into the Create, correct?

tock-ibm (commented):
overall it looks good, some minor comments

Issue #591, opened 2026-02-08
System Testing


tock-ibm:
### **Nightly runs**
Running small setup (4p2s) at low rate and for a long time. Pull from all assemblers. 
Success criteria: check that all transactions had arrived and that assembler ledgers are equal.
- Without failures
- With failures: every several minutes, pick a component (R,B,C,A) from party X, shut it down, and after a minute, restart it. Cycle among components and parties. At the end verify the success criteria.

### **Performance with signature verification**
- Augment armageddon load to produce signed TXs
- Measure performance with signed TXs (4p2s, 7p2s).

tock-ibm:
@moran-abilea

PR #702, opened 2026-03-18
fix armageddon to support assembler crashes

moran-abilea:
as part of #591 

as well as the chos run support that will be later added in the CI/CD part this PR fixes the armageddon.go file to support the fact that if the assembler crashes, the armageddon will just keep writing it's statsitics regullary and will return back to write more data once the assembler is fully up.

### DIFF
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -689,11 +689,9 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 		os.Exit(3)
 	}
 
-	var stream ab.AtomicBroadcast_DeliverClient
-	var gRPCAssemblerClientConn *grpc.ClientConn
 	endpointToPullFrom := userConfig.AssemblerEndpoints[pullFromPartyId-1]
 
-	gRPCAssemblerClientConn, err = gRPCAssemblerClient.Dial(endpointToPullFrom)
+	gRPCAssemblerClientConn, err := gRPCAssemblerClient.Dial(endpointToPullFrom)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to create a gRPC client connection to assembler %d: %v", pullFromPartyId, err)
 		os.Exit(3)
@@ -702,7 +700,7 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 	abc := ab.NewAtomicBroadcastClient(gRPCAssemblerClientConn)
 
 	// create a deliver stream
-	stream, err = abc.Deliver(context.TODO())
+	stream, err := abc.Deliver(context.TODO())
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to create a deliver stream to assembler %d: %v", pullFromPartyId, err)
 		os.Exit(3)
@@ -735,7 +733,7 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 		waitToFinish.Done()
 	}()
 
-	// every second read the statistics and send it to the manageStatistics
+	// every second read the statistics and send it to manageStatistics
 	go func() {
 		ticker := time.NewTicker(timeIntervalToSampleStat)
 		defer ticker.Stop()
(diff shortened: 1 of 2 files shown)


MayRosenbaum (commented):
As a first step, can you add a test ?

common/tools/armageddon/armageddon.go
```diff
@@ -973,6 +1013,10 @@ func writeStatisticsToCSV(file *os.File, statistic Statistics, timeIntervalToSam
 }
 
 func createRequestEnvelopeForUser(userConfig *UserConfig) (*common.Envelope, error) {
+	return createRequestEnvelopeForUserFromSeq(userConfig, 0)
```
MayRosenbaum: I think we don't need this.
Just call `createRequestEnvelopeForUserFromSeq` with Seq 0 in the right places.

common/tools/armageddon/armageddon.go
```diff
@@ -973,6 +1013,10 @@ func writeStatisticsToCSV(file *os.File, statistic Statistics, timeIntervalToSam
...
+}
+
+func createRequestEnvelopeForUserFromSeq(userConfig *UserConfig, startSeq uint64) (*common.Envelope, error) {
```
MayRosenbaum: Maybe `createDeliverRequestWithSeekInfo` ?

common/tools/armageddon/armageddon.go
```diff
@@ -831,35 +874,32 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
...
 }
 
-func calculateDelayOfTx(data []byte, acceptedTime time.Time) time.Duration {
-	sendTime := tx.ExtractTimestampFromTx(data)
-	delayTime := acceptedTime.Sub(sendTime)
-	return delayTime
-}
-
-func pullBlock(stream ab.AtomicBroadcast_DeliverClient, endpointToPullFrom string, gRPCAssemblerClientConn *grpc.ClientConn) (*common.Block, error) {
+// pullBlock no longer closes the connection internally — the caller manages reconnection
```
MayRosenbaum: Remove comment.
This comment explains the change you did, so you don't need to leave it here

common/tools/armageddon/armageddon.go
```diff
@@ -1000,6 +1044,9 @@ func createRequestEnvelopeForUser(userConfig *UserConfig) (*common.Envelope, err
 	return requestEnvelope, err
 }
 
+// receiveResponseFromAssembler is used by the submit command, which is a short-lived operation.
+// Unlike pullBlocksFromAssemblerAndCollectStatistics, no reconnect logic is needed here —
+// if the assembler goes down during a submit run, it is correct to fail fast.
```
MayRosenbaum:
```
no reconnect logic is needed here —
if the assembler goes down during a submit run, it is correct to fail fast. 
```
 --- >

No reconnect logic is supported.

common/tools/armageddon/armageddon.go
```diff
@@ -1059,7 +1106,7 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap,
 	numOfTxsCalculated := 0
 	var sumOfDelayTimes float64
 	for {
-		block, err := pullBlock(stream, endpointToPullFrom, gRPCAssemblerClientConn)
+		block, err := pullBlock(stream, endpointToPullFrom)
```
MayRosenbaum, line 1105: Now that `gRPCAssemblerClientConn` is not passed as an argument, I think that in the case of error, you should close the connection and the stream

common/tools/armageddon/armageddon_test.go
```diff
@@ -798,6 +799,175 @@ func checkCryptoDir(outputDir string) error {
...
+	checkCSVHasZeroRows(t, dir, 3) // Check at least 3 rows with zero values
+
+	// 9. Restart the assembler
+	t.Log("Phase 3: Restarting assembler to simulate recovery")
+	assembler.RestartArmaNode(t, readyChan)
+	testutil.WaitReady(t, readyChan, 1, 10)
+
+	// 10. Wait 30 seconds, then check assembler is up and CSV has non-zero rows again
+	time.Sleep(30 * time.Second)
+	t.Log("Phase 3: Verifying assembler is UP again and processing blocks")
```
MayRosenbaum: Verifying that the assembler is operational again and processing blocks

common/tools/armageddon/armageddon_test.go
```diff
@@ -798,6 +799,175 @@ func checkCryptoDir(outputDir string) error {
...
+		defer wg.Done()
+		armageddon.Run([]string{"load", "--config", userConfigPath, "--transactions", totalTxs, "--rate", rate, "--txSize", txSize})
+	}()
+
+	// 6. Wait 30 seconds, then check assembler is up and CSV has non-zero rows
+	time.Sleep(30 * time.Second)
+	t.Log("Phase 1: Checking assembler is UP and processing blocks")
+	assembler := armaNetwork.GetAssembler(t, 1)
+	require.NotNil(t, assembler.RunInfo.Session, "Assembler process should be running")
+	checkCSVHasNonZeroRows(t, dir, 3) // Check at least 3 rows with non-zero values
```
MayRosenbaum: Why 3 ?
Do we have any way to calculate the expected number of roles, or why 3 is the chosen number ?

common/tools/armageddon/armageddon.go
```diff
@@ -752,22 +750,65 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
...
+
 		for {
-			block, err := pullBlock(stream, endpointToPullFrom, gRPCAssemblerClientConn)
+			block, err := pullBlock(stream, endpointToPullFrom)
 			if err != nil {
-				fmt.Fprintf(os.Stderr, "failed to pull block from assembler %d: %v", pullFromPartyId, err)
-				os.Exit(3)
+				// Assembler is down — log and retry, do NOT exit
+				logger.Warnf("lost connection to assembler %d: %v — will retry in 1s", pullFromPartyId, err)
+
```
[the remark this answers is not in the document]
moran-abilea: I decided not to fix this, the rest of the comments I can agree with and it is already fixed but this is just overkill for nothing

MayRosenbaum (approved):

tock-ibm (approved):

PR #709, opened 2026-03-19
stream report with backoff

DorKatzelnick:
add backoff before reporting a bad stream between router and batcher
#711

### DIFF
--- a/node/router/stream.go
+++ b/node/router/stream.go
@@ -10,7 +10,9 @@ import (
 	"context"
 	"fmt"
 	"maps"
+	"strings"
 	"sync"
+	"time"
 
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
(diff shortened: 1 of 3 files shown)


node/router/stream.go
```diff
@@ -211,15 +215,35 @@ DrainChannelLoop:
...
 	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// if a we got an error that do not requires backoff, we reset the backoff interval to the minimum
+	if !withBackoff {
+		s.reportBackoffInterval = minRetryInterval
+	}
 	if !s.notifiedReconnect {
 		s.logger.Debugf("Reporting stream cIndex: %d, sIndex: %d to reconnection goroutine in shard-router", s.connNum, s.streamNum)
+		// report
```
tock-ibm: report the stream to the reconnect routine

Issue #707, opened 2026-03-19
All-in-one docker image for testing


tock-ibm:
- Generate a single docker image that has a 4 party 1 shard Arma network running in it. 
- Generate a test that send a few TXs and consumes them.
- The only ports exposed to the outside should be the 4 router and 4 assembler endpoints.
- Storage should be mapped outside, you'll need 4x4=16 different folders for "party<i>_<role>"

See node/examples, create another test that does the same but when arma is all within a single docker
We need this for integration testing between the orderer and committer

Don't change any of the existing dockerfile(s), create this in an entirely new folder under examples.

PR #715, opened 2026-03-22
All in one docker

moran-abilea:
related to #707 

in this PR there is 4 parties 1 shard container, plus a way to run it all together in a single script to upload the docker plus run a test for it.

### DIFF
--- a/node/consensus/synchronizer/synchronizer_bft.go
+++ b/node/consensus/synchronizer/synchronizer_bft.go
@@ -85,6 +85,10 @@ func (s *BFTSynchronizer) Stop() {
 	s.mutex.Lock()
 	defer s.mutex.Unlock()
 
+	if s.syncBuff == nil {
+		return
+	}
+
 	s.syncBuff.Stop() // This will stop the BFT deliverer and any go-routines we started in Sync()
 }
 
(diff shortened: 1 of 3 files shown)


node/examples/4parties1shard/entrypoint.sh
```diff
@@ -0,0 +1,120 @@
...
+    --config ${BASE_DIR}/config/party${i}/user_config.yaml \
+    --pullFromPartyId=${i} \
+    --expectedTxs=23000 \
+    --output="/tmp/output${i}" &
+done
+
+# LOADER
+armageddon load \
+  --config ${BASE_DIR}/config/party1/user_config.yaml \
+  --transactions=24000 \
```
tock-ibm: 1000 TXs is enough

tock-ibm: receive should equal load TX count

node/examples/all-in-one/config/example-deployment.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  - ID: 4
+    AssemblerEndpoint: "assembler.p4:6323"
+    ConsenterEndpoint: "consensus.p4:6325"
+    RouterEndpoint: "router.p4:6322"
+    BatchersEndpoints:
+      - "batcher1.p4:6324"
+
+UseTLSRouter: "none"
+UseTLSAssembler: "none"
```
tock-ibm: Do the test with mTLS please

Issue #445, opened 2025-12-09
Assembler apply config - dynamic


tock-ibm:
close comm, resize index, reopen comm

tock-ibm:
Config Ack in #641 
Join with catch-up #783

PR #746, opened 2026-03-29
Assembler apply config

DorKatzelnick:
issue #445 , #589

### DIFF
--- a/node/assembler/assembler.go
+++ b/node/assembler/assembler.go
@@ -14,6 +14,7 @@ import (
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-protos-go-apiv2/orderer"
+	"github.com/hyperledger/fabric-x-orderer/config"
 	node_config "github.com/hyperledger/fabric-x-orderer/node/config"
 
 	"github.com/hyperledger/fabric-x-orderer/node/consensus/state"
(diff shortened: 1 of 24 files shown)


node/assembler/assembler.go
```diff
@@ -292,3 +298,68 @@ func (a *Assembler) GetStatus() utils.NodeStatus {
...
+func (a *Assembler) ProcessNewConfigBlock(configBlock *common.Block) {
+	a.logger.Infof("Processing new config block number %d", configBlock.Header.Number)
+
+	a.SoftStop()
+
+	// extract new configuration and check if it can be applied.
+	newConfiguration, err := a.configuration.NewUpdatedConfigurationFromBlock(configBlock)
+	if err != nil {
+		a.logger.Panicf("Failed to apply last config: %v", err)
+	}
```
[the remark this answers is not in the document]
tock-ibm: Better - never allow NewAssembler  with nil configuration. I see no reason to allow this, it may be easier migrating tests like that, but may cause problems later.

node/assembler/assembler_test.go
```diff
@@ -195,6 +195,7 @@ func (at *assemblerTest) StartAssembler() {
 		at.logger,
 		&dummyAssemblerStopper{},
 		at.nodeConfig,
+		nil,
```
tock-ibm: Don't allow nil

test/send_config_update_test.go
```diff
@@ -477,7 +479,7 @@ func TestRemoveStoppedPartyThenRestart(t *testing.T) {
 	})
 
 	// Wait for Arma nodes to stop
-	testutil.WaitSoftStopped(t, netInfo)
+	testutil.WaitSoftStoppedByType(t, netInfo, []testutil.NodeType{testutil.Router, testutil.Batcher, testutil.Consensus})
```
tock-ibm: The router (and batcher?) does not support self relaunch already?

Issue #247, opened 2025-09-09
Epic: Integration tests - reconfiguration


tock-ibm:
### Prerequisites

Test the verification of data TXs #323 
Generate a config update TX #324 
Test the verification of  a config update TX #332 

### Integration tests of reconfiguration features

Dissemination

- [x] abcr dissemination
- [x] e2e dissemination 
- [x] abcr restart
- [x] e2e restart
- [x] check block signatures on config block emitted from assembler #463 

Change parameters
- [x] Ordering node endpoints
- [x] Ordering batching parameters
- [x] Ordering timeout parameters
- [x] SmartBFT parameters

Change certs
- [x] Change enrollment (signing) cert on single node in a party
- [x] Change TLS certs
- [x] Change CA certs

Change parties
- [x] remove party
- [x] add party

Verification of config changes
- [ ] make sure that illegal transitions are rejected

Advanced tests
- [x] join many parties to increase F: N=4 to N=7 etc.
- [x] remove parties to reduce F: N=7 to N=4 etc.
- [ ] check rules (internal and transition)
- [ ] add remove until all parties are replaced
- [ ] rejoin: start with 4 (1,2,3,4); shut down p=1; change endpoints of all other parties (2,3,4); restart p1, will not join; rejoin p1; give it the last config block - it will join;

PR #754, opened 2026-03-30
Test configuration update - Join many parties

gengur:
Push the configuration update and verify it was applied successfully.
Join several parties to increase F (maximum number of Byzantine nodes)

issue #247

### DIFF
--- a/test/send_config_update_test.go
+++ b/test/send_config_update_test.go
@@ -2521,3 +2521,188 @@ func TestRemoveMultipleParties(t *testing.T) {
 		Signer:       pullRequestSigner,
 	})
 }
+
+// TestJoinMultipleParties verifies that multiple parties can be added sequentially to a running Arma network.
+// The test bootstraps an initial multi-party network, sends baseline transactions to confirm liveness,
+// then iteratively adds three new parties via config updates. For each addition:
+//  1. A config update is created and submitted to add a new party.
+//  2. The network is stopped and restarted with the updated configuration.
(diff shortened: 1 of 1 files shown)


test/send_config_update_test.go
```diff
@@ -2521,3 +2520,188 @@ func TestRemoveMultipleParties(t *testing.T) {
...
+		Signer:       pullRequestSigner,
+	})
+
+	broadcastClient.Stop()
+
+	configBlockPath := filepath.Join(dir, "bootstrap", "bootstrap.block")
+
+	for range 3 {
+		// Create config update to add a new party
+		builder, _ := configutil.NewConfigUpdateBuilder(t, dir, configBlockPath)
```
HagarMeir, line 2617: please don't forget to clean up

test/send_config_update_test.go
```diff
@@ -2297,7 +2297,6 @@ func TestReplacePartiesPartially(t *testing.T) {
 		maps.Copy(netInfo, addedNetInfo)
 
 		numOfArmaNodes = len(netInfo)
-
```
HagarMeir, line 2300: no reason for this change here

test/send_config_update_test.go
gengur: The cleanup function calls gexec.CleanupBuildArtifacts, which removes all temporary artifacts created by gexec, including the arma compiled binaries. There is already a deferred call that will run before the test completes.

HagarMeir (approved):

PR #755, opened 2026-03-30
Bft synch genesis fetcher

tock-ibm:
Add to the old block puller, which now functions as a target height detector, to fetch the genesis blocks of every target node. We will then find (at least) F+1 identical blocks to select a valid block in a BFT fashion. 

This allows a consenter node to join an existing cluster of consenters using only a join-block, and fetch the entire chain. 

#35 
#757

### DIFF
--- a/node/comm/deliver.go
+++ b/node/comm/deliver.go
@@ -136,6 +136,148 @@ func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, string, error) {
 	return res, myEndpoint, endpointsInfo.err
 }
 
+// GenesisByEndpoints fetches the genesis block (block #0) from all accessible endpoints.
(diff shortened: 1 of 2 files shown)


node/comm/deliver.go
```diff
@@ -136,6 +136,148 @@ func (p *BlockPuller) HeightsByEndpoints() (map[string]uint64, string, error) {
...
+	}
+
+	blockNum := block.GetHeader().GetNumber()
+	if blockNum != 0 {
+		conn.Close()
+		return nil, errors.Errorf("expected block 0, got block %d", blockNum)
+	}
+	stream.CloseSend()
+
+	p.Logger.Debugf("Successfully fetched genesis block from %s", endpoint.Endpoint)
```
HagarMeir: could be info

PR #796, opened 2026-04-12
fix flake add retry logic for gRPC server port binding race conditions

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/791

### DIFF
--- a/node/utils/util.go
+++ b/node/utils/util.go
@@ -57,64 +57,109 @@ func ListenAddressForNode(endpointType ServerEndpointType, listenAddress string)
 }
 
 func CreateGRPCRouter(conf *config.RouterNodeConfig) *comm.GRPCServer {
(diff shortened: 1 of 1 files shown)


node/utils/util.go
```diff
@@ -57,64 +57,109 @@ func ListenAddressForNode(endpointType ServerEndpointType, listenAddress string)
...
+				RequireClientCert: conf.ClientAuthRequired,
+				Certificate:       conf.TLSCertificateFile,
+				Key:               conf.TLSPrivateKeyFile,
+			},
+		})
+		if err == nil {
+			return srv
+		}
+		// If address is in use, wait and retry
+		if strings.Contains(err.Error(), "address already in use") {
```
tock-ibm, line 81: I don't think it is a good idea to change the behavior of the production server because of test flakes. In production we want to fail fast. In testing we should allocate ports to different clusters such that it never happens.

Issue #835, opened 2026-04-16
CI: Split integration-tests job in github actions


tock-ibm:
Divide integration tests in `test` into multiple sub-packages by topic. For example:

`test/basic`
`test/faulttolerance`
`test/reconfig`

Divide the single integration test `job` into multiple jobs. See: #834 for reasoning.

PR #850, opened 2026-04-23
Split integration tests

moradna:
Split integration tests into basic, faulttolerance and reconfig.
Moved tests to the correct category.
Split CI jobs accordingly.

#835


tock-ibm:
@moradna 
The direction looks good!

I see that reconfig already takes 26m; can we split reconfig even further? e.g. reconfig/membership = add/remove party
reconfig/identity = change certs, CAs etc
reconfig/endpoints = change endpoints
reconfig/params = change params

The goal is that each such job will take < approx. 10m

Issue #94, opened 2025-07-09
ReConfig: Validate params 


tock-ibm:
Validate incoming config params, local and shared; for internal consistency.

For example the request  max bytes of routers, batchers and consenters should match or be consistent - router <= batcher, consenter.

Timeouts should make sense - not zero, have a minimal accepted value, etc.

This should be applied to genesis block as well to make sure we start form a legal block

And to enforce also the validity of incoming reconfiguration updates on the new block.

PR #856, opened 2026-04-27
Add block validation policy check to orderer rules

moradna:
#94

### DIFF
--- a/config/verify/orderer_rules.go
+++ b/config/verify/orderer_rules.go
@@ -141,8 +141,8 @@ func (or *DefaultOrdererRules) ValidateNewConfig(envelope *common.Envelope, bccs
 	config := bundle.ConfigtxValidator().ConfigProto()
 	ordererGroup := config.ChannelGroup.Groups["Orderer"]
 
-	if err := validateBlockValidationPolicy(ordererGroup.Policies["BlockValidation"], ordererConfig.Consenters()); err != nil {
-		return errors.Wrap(err, "block validation policy does not match current consenters")
+	if err := validateBlockValidationPolicy(ordererGroup.Policies[policies.BlockValidationPolicyKey], ordererConfig.Consenters()); err != nil {
+		return errors.Wrap(err, "invalid block validation policy")
 	}
 
 	return nil
--- a/config/verify/orderer_rules.go
+++ b/config/verify/orderer_rules.go
@@ -13,8 +13,13 @@ import (
 	smartbft_types "github.com/hyperledger-labs/SmartBFT/pkg/types"
 	"github.com/hyperledger/fabric-lib-go/bccsp"
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
+	"github.com/hyperledger/fabric-x-common/api/msppb"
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/cockroachdb/errors v1.12.0
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/gorilla/handlers v1.5.2
-	github.com/hyperledger/fabric-x-common v0.2.1
+	github.com/hyperledger/fabric-x-common v0.2.2-0.20260426075714-d5be85ec5abe
```
tock-ibm: There is already a tag for the fabric-x-common v0.2.2 that includes the fixes we need here

testutil/configutil/config_update_utils.go
```diff
@@ -1163,3 +1168,35 @@ func (c *ConfigUpdateBuilder) PrepareAndAddNewParty(t *testing.T, dir string) (t
 
 	return addedPartyId, addedNetInfo
 }
+
+func (c *ConfigUpdateBuilder) syncBlockValidationPolicy(t *testing.T, consenterMappingList []any) {
```
tock-ibm, line 1172: We are putting in the block but we never use `vendor/github.com/hyperledger/fabric-x-common/protoutil/blockutils.go`  `BlockSignatureVerifier` in any of our integration tests.

Can we add at least a single test case that uses it?

config/verify/orderer_rules.go
```diff
@@ -133,6 +138,14 @@ func (or *DefaultOrdererRules) ValidateNewConfig(envelope *common.Envelope, bccs
 		}
 	}
 
+	// 8.
+	config := bundle.ConfigtxValidator().ConfigProto()
+	ordererGroup := config.ChannelGroup.Groups["Orderer"]
+
+	if err := validateBlockValidationPolicy(ordererGroup.Policies["BlockValidation"], ordererConfig.Consenters()); err != nil {
```
tock-ibm: I am worried that ordererConfig.Consenters() may not be producing consistent order, which may lead to forks.
Can you verify?

testutil/configutil/config_update_utils.go
tock-ibm: We'll do it in #869

config/verify/orderer_rules.go
```diff
@@ -504,3 +517,43 @@ func validatePartyModification(curr, next *config_protos.PartyConfig) (bool, err
...
+		})
+	}
+
+	expected := &common.Policy{
+		Type: int32(common.Policy_SIGNATURE),
+		Value: protoutil.MarshalOrPanic(&common.SignaturePolicyEnvelope{
+			Rule:       policydsl.NOutOf(int32(policies.ComputeBFTQuorum(n, f)), signedBy),
+			Identities: identities,
+		}),
+	}
```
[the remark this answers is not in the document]
moradna: EncodeBFTBlockVerificationPolicy updates the config in-place and does not return a policy, so it can't be used here.

config/verify/orderer_rules.go
moradna: Fixed. Validation no longer depends on the order of `Consenters().`

tock-ibm (approved):

Issue #834, opened 2026-04-16
CI: Split unit-tests job in github actions


tock-ibm:
Split the single unit-test job in CI in to multiple jobs. This will make each job run on an isolated VM, in parallel. It will mitigate some port binding leakage, as well as allow us to identify and rerun only the flakey jobs. Jobs run in parallel by default, they should ` needs: basic-checks` only. This will shorten the CI time as well.

The `node/consensus` tests should have their own job.
The `node/batcher` tests should have their own job.
All the rest can be on a single job.

```
jobs:
  basic-checks:
   ...
  unit-tests:
    name: Unit Tests
    needs: basic-checks
    runs-on: ${{ github.repository == 'hyperledger/fabric-x-orderer' && 'ubuntu-24.04' }}
    steps:
      - uses: actions/checkout@v5
        with:
          fetch-depth: 0
        name: Checkout Fabric-X Code
      - uses: actions/setup-go@v6
        name: Install Go
        with:
          go-version-file: go.mod
      - run: make unit-tests
        name: Run Unit Tests
```

PR #866, opened 2026-04-30
Update CI unit test jobs

moradna:
Follow-up to comments in PR #858 
#834

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,9 @@ check-protos:
 	@echo "Checking protos..."
 	@./scripts/check_protos.sh
 
+.PHONY: unit-tests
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -75,14 +75,14 @@ check-protos:
...
+	go test -race -timeout 20m ./node/consensus/...
 
 .PHONY: unit-tests-batcher
 unit-tests-batcher:
 	go test -race -timeout 15m ./node/batcher/...
 
-.PHONY: unit-tests
-unit-tests:
+.PHONY: unit-tests-other
+unit-tests-other:
```
HagarMeir, line 88: add also make unit-tests that runs like the original (with all unit tests) 
but is not used by CI

HagarMeir (approved):

PR #905, opened 2026-05-31
shorten log message

HagarMeir:


node/comm/commauth.go
```diff
@@ -68,7 +68,9 @@ func (ac *AuthCommMgr) Remote(id uint64) (*RemoteContext, error) {
 }
 
 func (ac *AuthCommMgr) Configure(members []RemoteNode) {
-	ac.Logger.Infof("Configuring communication module with nodes:%v", members)
+	for _, node := range members {
+		ac.Logger.Infof("Configuring communication module with node - ID: %d, endpoint: %s", node.ID, node.Endpoint)
```
tock-ibm, line 72: make sure node is not `nil` or it will panic

node/consensus/synchronizer/height_detector_factory.go
```diff
@@ -122,7 +122,9 @@ func newBlockPuller(
 		Dialer:              stdDialer,
 	}
 
-	logger.Infof("Built new block puller (target height detector) with endpoints: %+v", endpoints)
+	for _, endpoint := range endpoints {
+		logger.Infof("Built new block puller (target height detector) with endpoint: %s", endpoint.Endpoint)
```
tock-ibm, line 126: make sure endpoint is not `nil` or it will panic

node/comm/commauth.go
HagarMeir: not a pointer so cannot be nil

node/consensus/synchronizer/height_detector_factory.go
HagarMeir: same idea...endpoint cannot be nil, not a pointer

tock-ibm (approved):

PR #920, opened 2026-06-14
add config seq to config req

HagarMeir:


node/consensus/consensus.go
```diff
@@ -244,11 +243,9 @@ func (c *Consensus) OnConsensus(channel string, sender uint64, request *orderer.
...
-	bafd := &state.BAFDeserialize{}
-	if err := ce.FromBytes(rawCE, bafd.Deserialize); err != nil {
-		c.Logger.Errorf("Failed unmarshaling control event %s: %v", base64.StdEncoding.EncodeToString(rawCE), err)
-		return errors.Wrap(err, "failed unmarshaling control event")
+	ri, _, err := c.verifyCE(rawCE)
+	if err != nil {
+		c.Logger.Errorf("Failed verifying control event %v: %v", ri, err)
 	}
 	c.BFT.HandleRequest(sender, rawCE)
 	return nil
```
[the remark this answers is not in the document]
tock-ibm: True

node/consensus/state/state.go
```diff
@@ -305,7 +319,7 @@ func (c *ConfigRequest) FromBytes(bytes []byte) error {
 
 func (c *ConfigRequest) String() string {
 	// TODO: add more info to this string, at least the config sequence
-	return "Config Request"
+	return fmt.Sprintf("Config Request with config sequence %d", c.ConfigSeq)
 }
```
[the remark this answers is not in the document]
tock-ibm: makes sense

node/consensus/state/state.go
```diff
@@ -281,7 +281,8 @@ func (c *Complaint) String() string {
...
 
 type ConfigRequest struct {
-	Envelope *common.Envelope
+	Envelope  *common.Envelope
+	ConfigSeq types.ConfigSequence
```
tock-ibm: Why do we need this? the Envelope  contains in it a Config, which has the config sequence in it, no?
IE it seems to me all we need is to extract it?

HagarMeir: It is just easy to use the config seq already present in the router's config request instead of extracting it 
And saving it here doesn't cost anything so why not keep it here outside 
Just making the behavior similar to other control events 
Let me know if you think differently

tock-ibm: It is just a source of mismatch.... now you have to be sure that what is in the envelope and what is in the `ConfigSeq` is the same...
So adding a method `ConfigRequest.ConfigSequence() types.ConfigSequence`, that simply parses the envelope, seems like a better approach to me.
This happens rarely so making it simple and consistent seems like a better approach.
You then don't need to marshal the config sequence etc.

node/consensus/state/state.go
```diff
@@ -304,8 +318,8 @@ func (c *ConfigRequest) FromBytes(bytes []byte) error {
 }
 
 func (c *ConfigRequest) String() string {
-	// TODO: add more info to this string, at least the config sequence
-	return "Config Request"
+	// TODO: add more info to this string
```
tock-ibm: Remove comment; not sure what else we can add here

node/consensus/state/state.go
HagarMeir: done

node/consensus/state/state.go
```diff
@@ -284,16 +285,34 @@ type ConfigRequest struct {
 	Envelope *common.Envelope
 }
 
+func (c *ConfigRequest) ConfigSequence() types.ConfigSequence {
+	payload, err := protoutil.UnmarshalPayload(c.Envelope.Payload)
+	if err != nil {
+		panic(fmt.Sprintf("failed to unmarshal payload: %v", err))
```
tock-ibm: Why panic and not error? 
Is this not an attack vector of the leader on all followers? 
I would return an error and reject / drop the TX instead of panic...

tock-ibm (approved):

Issue #783, opened 2026-04-09
Assembler - Synchronizer for node join and catchup


tock-ibm:
Create a version of the consensus synchronizer in order to facilitate node join and catchup.

Use a common base, move common blocks to `common/synchronizer`.

Change the bft deliverer so that it requires "Endpoint extractor" interface. This will allow the bft deliverer to be used when used against consenter or assembler endpoints.

PR #937, opened 2026-06-18
add assembler signer

DorKatzelnick:
#783

### DIFF
--- a/test/utils/utils.go
+++ b/test/utils/utils.go
@@ -208,7 +208,7 @@ func CreateAssemblers(t *testing.T, num int, ca tlsgen.CA, shards []node_config.
 		logger := testutil.CreateLogger(t, i+1)
 		loggers = append(loggers, logger)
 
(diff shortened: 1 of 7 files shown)


MayRosenbaum:
LGTM

test/utils/utils.go
```diff
@@ -208,7 +208,7 @@ func CreateAssemblers(t *testing.T, num int, ca tlsgen.CA, shards []node_config.
 		logger := testutil.CreateLogger(t, i+1)
 		loggers = append(loggers, logger)
 
-		assembler := assembler.NewAssembler(assemblerConf, &config.Configuration{}, genesisBlock, make(chan struct{}), logger)
+		assembler := assembler.NewAssembler(assemblerConf, &config.Configuration{}, genesisBlock, make(chan struct{}), logger, &mocks.SignerSerializer{})
```
tock-ibm, line 211: Although it passes now, it may fail pretty soon after we add tests that do synch. 
Even though this is used only in "ABCR" type tests.
In all other components we use a real signer.

Lets keep this in mind and change it in the next PRs, please

tock-ibm (approved):

PR #945, opened 2026-06-22
Signed armageddon

moran-abilea:
added --isSigned flag to armageddon load command.
added test as well

### DIFF
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -34,6 +34,7 @@ import (
 	"github.com/hyperledger/fabric-x-orderer/config"
 	genconfig "github.com/hyperledger/fabric-x-orderer/config/generate"
 	"github.com/hyperledger/fabric-x-orderer/node/comm"
+	"github.com/hyperledger/fabric-x-orderer/node/crypto"
 	"github.com/hyperledger/fabric-x-orderer/testutil/fabric"
 	"github.com/hyperledger/fabric-x-orderer/testutil/signutil"
 	"github.com/hyperledger/fabric-x-orderer/testutil/tx"
@@ -123,6 +124,7 @@ type CLI struct {
 	loadTransactions   *int
 	loadRate           *string
 	loadTxSize         *int
+	loadSignedMode     *string
 	// receive command flags
 	receiveUserConfigFile   **os.File
 	receiveExpectedNumOfTxs *int
@@ -173,6 +175,7 @@ func (cli *CLI) configureCommands() {
 	cli.loadTransactions = load.Flag("transactions", "The number of transactions to be sent").Int()
 	cli.loadRate = load.Flag("rate", "The rate specifies the number of transactions per second to be sent as one or more rate numbers separated by space").String()
 	cli.loadTxSize = load.Flag("txSize", "The required transaction size in bytes").Int()
+	cli.loadSignedMode = load.Flag("signedMode", "Signing mode has three option: none = unsigned, full = sign including the full certificate, short = sign using known certificate").String()
 	commands["load"] = load
 
 	receive := cli.app.Command("receive", "Pull txs from some assembler and report statistics")
@@ -219,7 +222,7 @@ func (cli *CLI) Run(args []string) {
 
 	// "load" command
 	case cli.commands["load"].FullCommand():
-		load(cli.loadUserConfigFile, cli.loadTransactions, cli.loadRate, cli.loadTxSize)
+		load(cli.loadUserConfigFile, cli.loadTransactions, cli.loadRate, cli.loadTxSize, cli.loadSignedMode)
 
 	// "receive" command
 	case cli.commands["receive"].FullCommand():
@@ -478,7 +481,7 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
 }
 
 // load command makes txs and sends them to all routers
-func load(userConfigFile **os.File, transactions *int, rate *string, txSize *int) {
+func load(userConfigFile **os.File, transactions *int, rate *string, txSize *int, signedMode *string) {
 	rates := strings.Fields(*rate)
 	// check transaction size
 	txMinimumSize := 16 + 8 + 8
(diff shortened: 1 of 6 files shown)


common/tools/armageddon/armageddon_test.go
```diff
@@ -276,6 +287,192 @@ func TestLoadAndReceive(t *testing.T) {
...
+
+		if err := verifySignedEnvelope(env); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func isSignedLoadTransactionEnvelope(env *common.Envelope) (bool, error) {
```
MayRosenbaum: I dont understand the purpose of this method

common/tools/armageddon/armageddon_test.go
```diff
@@ -276,6 +287,192 @@ func TestLoadAndReceive(t *testing.T) {
...
+	uc, err := testutil.GetUserConfig(dir, 1)
+	require.NoError(t, err)
+	require.NotNil(t, uc)
+
+	logger := testutil.CreateLogger(t, 0)
+	verifier := test_utils.BuildVerifier(dir, types.PartyID(1), logger)
+
+	parties := []types.PartyID{1, 2, 3, 4}
+	statusUnknown := common.Status_UNKNOWN
+	test_utils.PullFromAssemblers(t, &test_utils.BlockPullerOptions{
```
MayRosenbaum: You are using a different approach here.
The `PullFromAssemblers` is a method used in our integration tests.
Why not use the receive command as done in all armageddon tests ?

We want to confirm that armageddon works as expected, i.e., we can load signed transactions and receive them.

common/tools/armageddon/armageddon_test.go
```diff
@@ -276,6 +287,192 @@ func TestLoadAndReceive(t *testing.T) {
...
+		Transactions:     totalTxs,
+		NeedVerification: false,
+		BlockHandler:     &signedEnvelopeBlockHandler{},
+		ErrString:        "cancelled pull from assembler: %d; pull ended: failed to receive a deliver response: rpc error: code = Canceled desc = grpc: the client connection is closing",
+	})
+}
+
+type signedEnvelopeBlockHandler struct{}
+
+func (h *signedEnvelopeBlockHandler) HandleBlock(t *testing.T, block *common.Block) error {
```
MayRosenbaum: Do we need this?
What are you trying to verify ?

common/tools/armageddon/armageddon.go
```diff
@@ -541,8 +547,42 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
...
+	var org string
+
+	if isSigned {
+		org, err = signutil.GetMspIDfromDir(userConfig.MSPDir)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "failed to get mspID from user msp dir: %v", err)
+			os.Exit(3)
+		}
+
+		keyPath := filepath.Join(userConfig.MSPDir, "keystore", "priv_sk")
```
MayRosenbaum: I think you can use `LoadCryptoMaterialsFromDir` to get a signer from mspDIr, try to reuse code

testutil/tx/tx_utils.go
```diff
@@ -227,6 +227,58 @@ func PrepareEnvWithTimestamp(txNumber int, envSize int, sessionNumber []byte) *c
...
+		overheadSize -= 2
+	}
+
+	dataSize := envSize
+	if envSize >= overheadSize {
+		dataSize = envSize - overheadSize
+	}
+
+	var env *common.Envelope
+	for i := 0; i < 5; i++ {
```
MayRosenbaum: Why do we iterate 5 times?

The method `PrepareEnvWithTimestamp` creates an envelope of a requested size.

Suppose we want to create a transaction with size 300, meaning the final envelope size should be 300. 
To achieve that, we need to calculate the correct data size, specifically the size of `envelope.payload`, so that the full envelope ends up being exactly 300 bytes.

Now that we add signatures, they increase the total transaction size.

common/tools/armageddon/armageddon_test.go
moran-abilea: I thought that the test should also verify that each tx is really signed, but since we don't need to test it as part of this test i will change it to the simpler test approach instead

testutil/utils.go
```diff
@@ -694,16 +694,9 @@ func sortArmaNodeInfo(infos []*ArmaNodeInfo) func(i, j int) bool {
 }
 
 func LoadCryptoMaterialsFromDir(t *testing.T, mspDir string) (*crypto.ECDSASigner, []byte, error) {
```
MayRosenbaum, line 696: `LoadCryptoMaterialsFromDir` calls `LoadCryptoMaterials`, but they effectively do the same thing, one depends on the `t` and the other not. 
I would give `LoadCryptoMaterials` a clearer name, such as `LoadCryptoMaterialForSigner`.

common/tools/armageddon/armageddon.go
```diff
@@ -123,6 +124,7 @@ type CLI struct {
 	loadTransactions   *int
 	loadRate           *string
 	loadTxSize         *int
+	loadIsSigned       *bool
```
MayRosenbaum: We need to support three modes:
1. No signatures
2. Full signatures — including the certificate
3. Short signatures — using known certificates, referenced by hash

To prepare for this, let’s define a string flag, called "signature-mode," for example, with three meaningful values.

If the flag is missing from the command, it means no signatures should be used and the submitted transaction should remain unchanged from today (when txSize = 300, the size will refer to the full envelope size)

If the flag is set to `full`, the transaction is signed and includes the full certificate. This increases the transaction size. Therefore, when txSize = 300, the size will refer only to the payload data, not to the full envelope size.

If the flag is set to `short`, the transaction is signed and includes only the certificate **hash**. In this case, the transaction size is smaller than a transaction signed in full mode.

common/tools/armageddon/armageddon.go
```diff
@@ -541,16 +547,56 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
...
+		for i := 0; i < numOfTxs; i++ {
+			env = tx.PrepareSignedEnvelopeWithCertificate(i, txSize, sessionNumber, signer, certBytes, org)
+			status := rl.GetToken()
+			if !status {
+				fmt.Fprintf(os.Stderr, "failed to send tx %d in signed mode %s", i+1, signedMode)
+				os.Exit(3)
+			}
+			broadcastClient.SendTxToAllRouters(env)
+		}
+	case "short":
```
tock-ibm, line 580: If `short` is not yet supported, exit here with a message "`short` is not yet supported"

tock-ibm (commented):
LGTM, just one comment

tock-ibm (approved):

PR #947, opened 2026-06-23
Add batcher latency metrics

moradna:
#414

### DIFF
--- a/common/monitoring/provider.go
+++ b/common/monitoring/provider.go
@@ -180,3 +180,15 @@ func GetMetricValue(m prometheus.Metric, logger *flogging.FabricLogger) float64 
 		return 0
 	}
 }
+
+func GetHistogramAverage(m prometheus.Metric, logger *flogging.FabricLogger) float64 {
+	gm := promgo.Metric{}
+	if err := m.Write(&gm); err != nil {
+		logger.Error(err)
+		return 0
+	}
+	if gm.Histogram == nil || gm.Histogram.GetSampleCount() == 0 {
+		return 0
+	}
+	return gm.Histogram.GetSampleSum() / float64(gm.Histogram.GetSampleCount())
+}
--- a/node/batcher/batcher_role.go
+++ b/node/batcher/batcher_role.go
@@ -315,6 +315,7 @@ func (b *BatcherRole) runPrimary() {
 
 	var currentBatch types.BatchedRequests
 	var digest []byte
+	var batchToLedgerStart time.Time
 
 	for {
 		for {
@@ -329,12 +330,17 @@ func (b *BatcherRole) runPrimary() {
 				return
 			case <-ch:
 			}
+			mempoolStart := time.Now()
 			ctx, cancel := context.WithTimeout(b.stopCtx, b.BatchTimeout)
 			currentBatch = b.MemPool.NextRequests(ctx)
+			mempoolWaitLatency := time.Since(mempoolStart)
 			if len(currentBatch) == 0 {
 				cancel()
 				continue
 			}
+			b.Metrics.batchMempoolWaitLatency.Observe(mempoolWaitLatency.Seconds())
+			batchToLedgerStart = time.Now()
+
 			b.Logger.Infof("Batcher batched a total of %d requests for sequence %d", len(currentBatch), b.seq)
 			digest = currentBatch.Digest()
 			cancel()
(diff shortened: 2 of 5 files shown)


node/batcher/batcher_role.go
```diff
@@ -329,12 +330,19 @@ func (b *BatcherRole) runPrimary() {
 				return
 			case <-ch:
 			}
+			mempoolStart := time.Now()
 			ctx, cancel := context.WithTimeout(b.stopCtx, b.BatchTimeout)
 			currentBatch = b.MemPool.NextRequests(ctx)
+
+			b.Metrics.batchMempoolWaitLatency.Observe(time.Since(mempoolStart).Seconds())
```
tock-ibm: It is doing `NextRequests` so lets be clear and call it `batchMempoolNextRequestsLatency`

node/batcher/batcher_role.go
```diff
@@ -348,7 +356,7 @@ func (b *BatcherRole) runPrimary() {
 		// Once the BAF reached the consenters it is considered safe to remove the requests from the mem pool
 
 		b.Ledger.Append(b.ID, b.seq, b.ConfigSequenceGetter.ConfigSequence(), currentBatch, baf.Signature())
-
+		b.Metrics.batchMempoolToLedgerLatency.Observe(time.Since(batchToLedgerStart).Seconds())
```
tock-ibm: I think a more accurate name is `batchLedgerAppendLatency`.
It can be the same in primary and secondary.

node/batcher/metrics.go
tock-ibm, line 1: Add a test in integration that pulls the metrics and expects something that makes sense, in terms of formatting only, not actual values. There is an existing test that pulls some metrics, maybe expand it.

node/batcher/batcher_role.go
```diff
@@ -397,8 +405,10 @@ func (b *BatcherRole) runSecondary() {
 		out := b.BatchPuller.PullBatches(b.primary)
 		for {
 			var batch types.Batch
+			var batchReceivedTime time.Time
 			select {
 			case batch = <-out:
+				batchReceivedTime = time.Now()
```
tock-ibm: Separate the time it takes to validate from the time it takes to append.
The append can be actually the same metric as the primary, and they will appear on the same graph, which is good. 

`batchVerifyLatency` - from receive till after verify

`batchLedgerAppendLatency` - around the batch append in both primary and secondary.

common/monitoring/provider.go
```diff
@@ -180,3 +180,15 @@ func GetMetricValue(m prometheus.Metric, logger *flogging.FabricLogger) float64
...
+func GetHistogramAverage(m prometheus.Metric, logger *flogging.FabricLogger) float64 {
+	gm := promgo.Metric{}
+	if err := m.Write(&gm); err != nil {
+		logger.Error(err)
+		return 0
+	}
+	if gm.Histogram == nil || gm.Histogram.GetSampleCount() == 0 {
+		return 0
+	}
+	return gm.Histogram.GetSampleSum() / float64(gm.Histogram.GetSampleCount())
```
tock-ibm, line 193: Verify this is the correct way to calculate the average.

node/batcher/metrics.go
```diff
@@ -83,6 +83,30 @@ var (
 		Help:       "The total number of first resends performed.",
 		LabelNames: []string{"party_id", "shard_id"},
 	}
+
+	batchMempoolWaitLatencyOpts = metrics.HistogramOpts{
+		Namespace:  "batcher",
+		Name:       "batch_mempool_wait_latency_seconds",
+		Help:       "The latency of getting the next batch from the mempool.",
+		LabelNames: []string{"party_id", "shard_id"},
+		Buckets:    []float64{.0001, .001, .002, .003, .004, .005, .01, .03, .05, .1, .3, .5, 1}, // TODO: adjust buckets after reviewing Grafana
```
tock-ibm, line 92: Reviewing after a performance test, to have better resolution around the expected average

common/monitoring/provider.go
moradna: Verified. I checked the Prometheus documentation: https://prometheus.io/docs/prometheus/latest/querying/functions/#histogram_avg

node/batcher/batcher_role.go
```diff
@@ -415,6 +423,7 @@ func (b *BatcherRole) runSecondary() {
 				b.BatchPuller.Stop()
 				break // TODO maybe add backoff
 			}
+			b.Metrics.batchVerifyLatency.Observe(time.Since(verifyStart).Seconds())
```
tock-ibm: If we break, we don't observe... take the verify out of the `if`, observe it, and handle the error later

tock-ibm (approved):

Issue #950, opened 2026-06-24
Test flake in TestConfigTXDisseminationWithVerification


HagarMeir:
```
--- FAIL: TestConfigTXDisseminationWithVerification (84.19s)
    utils.go:832: Pulling from party: 1
    utils.go:832: Pulling from party: 3
    utils.go:832: Pulling from party: 2
    utils.go:832: Pulling from party: 4
    utils.go:834: Finished pull and count: blocks 2, txs 2 from party: 4
    utils.go:834: Finished pull and count: blocks 2, txs 2 from party: 2
    utils.go:834: Finished pull and count: blocks 2, txs 2 from party: 1
    utils.go:834: Finished pull and count: blocks 2, txs 2 from party: 3
    utils.go:832: Pulling from party: 3
    utils.go:832: Pulling from party: 4
    utils.go:832: Pulling from party: 2
    utils.go:832: Pulling from party: 1
    utils.go:834: Finished pull and count: blocks 3, txs 3 from party: 2
    utils.go:834: Finished pull and count: blocks 3, txs 3 from party: 1
    utils.go:834: Finished pull and count: blocks 3, txs 3 from party: 3
    utils.go:834: Finished pull and count: blocks 2, txs 2 from party: 4
    utils.go:652:
        	Error Trace:	/home/runner/work/fabric-x-orderer/fabric-x-orderer/test/utils/utils.go:652
        	            				/opt/hostedtoolcache/go/1.26.2/x64/src/sync/waitgroup.go:258
        	            				/opt/hostedtoolcache/go/1.26.2/x64/src/runtime/asm_amd64.s:1771
        	Error:      	"2" is not greater than or equal to "3"
        	Test:       	TestConfigTXDisseminationWithVerification
    config_disseminate_test.go:401:
        	Error Trace:	/home/runner/work/fabric-x-orderer/fabric-x-orderer/test/reconfig/configtx/config_disseminate_test.go:401
        	Error:      	Not equal:
        	            	expected: 0x2
        	            	actual  : 0x3
        	Test:       	TestConfigTXDisseminationWithVerification
```

Issue #638, opened 2026-02-23
Test TX with Cert-ID


tock-ibm:
- Construct a config block with a client that has a cert in known-identities (App client)
- Sign a TX an put the creator as Cert-ID rather than Cert
- Submit TX to router, verify it passes
- Check the negative case - when Cert-ID is not in known-identities verification fails

PR #954, opened 2026-06-25
remove config disseminate test and test data tx submission with verification

MayRosenbaum:
issue #950


test/basic/sigverify_test.go
```diff
@@ -123,3 +123,74 @@ func TestSubmitReceiveAndVerifySignaturesAssemblerBlocks(t *testing.T) {
...
+	t.Log("Run Arma nodes")
+	// NOTE: if one of the nodes is not started within 10 seconds, there is no point in continuing the test, so fail it
+	readyChan := make(chan string, numOfArmaNodes)
+	armaNetwork := testutil.RunArmaNodes(t, dir, armaBinaryPath, readyChan, netInfo)
+	defer armaNetwork.Stop()
+	testutil.WaitReady(t, readyChan, numOfArmaNodes, 10)
+
+	t.Log("Create a signer")
+	submittingParty := 1
+	uc, err := testutil.GetUserConfig(dir, types.PartyID(submittingParty))
```
tock-ibm, line 161: This is ok, but we need to add a use case that also uses:
1. an app org signer
2. sign with known certs

All of this should be done in #638 in a later PR.

tock-ibm (approved):
LGTM. 

In a different PR we need to add a use cases that also uses:

1. an app org signer
2. sign with known certs

All of this should be done in issue #638 and probably added here.

Issue #830, opened 2026-04-16
Consensus dynamically reconfigure SmartBFT


HagarMeir:
Currently when we reconfigure we stop BFT, rebuild it and start.
It should dynamically reconfigure on its own (by noticing what is returned from `Deliver`).
Notice that the `c.BFT.Comm` should also be updated somehow.
Maybe check how it is done in Fabric.

PR #959, opened 2026-06-28
smartbft dynamic reconfig

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/830

### DIFF
--- a/node/comm/commauth.go
+++ b/node/comm/commauth.go
@@ -136,7 +136,7 @@ func (ac *AuthCommMgr) getOrCreateMapping() MemberMapping {
 func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNode) {
 	stub := mapping.ByID(node.ID)
 	if stub == nil {
-		ac.Logger.Infof("Allocating a new stub for node %v with endpoint %v", node.ID, node.Endpoint)
+		ac.Logger.Infof("Allocating a new stub for node %d with endpoint %v", node.ID, node.Endpoint)
 		stub = &Stub{}
 	}
 
@@ -146,12 +146,11 @@ func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNod
 	// Put the stub into the mapping
 	mapping.Put(stub)
 
-	// Check if the stub needs activation.
-	if stub.Active() {
-		return
-	}
+	// First deactivate the stub
+	stub.Deactivate()
+	ac.Connections.Disconnect(stub.Endpoint)
 
-	// Activate the stub
+	// Now activate the stub
 	stub.Activate(ac.createRemoteContext(stub))
 }
 
--- a/node/comm/commauth.go
+++ b/node/comm/commauth.go
@@ -140,18 +140,20 @@ func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNod
 		stub = &Stub{}
 	}
 
+	// First disconnect and deactivate the stub
+	ac.Connections.Disconnect(stub.Endpoint)
+	stub.Deactivate()
+
 	// Overwrite the stub Node data with the new data
 	stub.RemoteNode = node
 
 	// Put the stub into the mapping
 	mapping.Put(stub)
 
-	// First deactivate the stub
-	stub.Deactivate()
-	ac.Connections.Disconnect(stub.Endpoint)
-
 	// Now activate the stub
-	stub.Activate(ac.createRemoteContext(stub))
+	if err := stub.Activate(ac.createRemoteContext(stub)); err != nil {
+		ac.Logger.Warnf("Unable to activate stub for node %d: %v", node.ID, err)
+	}
 }
 
 func (ac *AuthCommMgr) createRemoteContext(stub *Stub) func() (*RemoteContext, error) {
--- a/node/comm/commauth.go
+++ b/node/comm/commauth.go
@@ -7,8 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package comm
 
 import (
+	"bytes"
 	"context"
 	"encoding/asn1"
+	"slices"
 	"strconv"
 	"sync"
 
@@ -138,6 +140,11 @@ func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNod
 	if stub == nil {
 		ac.Logger.Infof("Allocating a new stub for node %d with endpoint %v", node.ID, node.Endpoint)
 		stub = &Stub{}
+	} else if stub.Active() && sameRemoteNode(stub.RemoteNode, node) {
+		// The node's data is unchanged and its connection is already active; leave it
+		// untouched so we don't needlessly tear down and rebuild a healthy connection.
+		ac.Logger.Debugf("Stub for node %d with endpoint %v is unchanged; leaving it active", node.ID, node.Endpoint)
+		return
 	}
 
 	// First disconnect and deactivate the stub
@@ -154,6 +161,17 @@ func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNod
 	if err := stub.Activate(ac.createRemoteContext(stub)); err != nil {
 		ac.Logger.Warnf("Unable to activate stub for node %d: %v", node.ID, err)
 	}
+}
+
+// sameRemoteNode reports whether two RemoteNodes carry identical address and certificate
+// data, i.e. reconfiguring to b would not change how we connect to or authenticate a.
+func sameRemoteNode(a, b RemoteNode) bool {
+	return a.ID == b.ID &&
+		a.Endpoint == b.Endpoint &&
+		bytes.Equal(a.ServerTLSCert, b.ServerTLSCert) &&
+		bytes.Equal(a.ClientTLSCert, b.ClientTLSCert) &&
+		bytes.Equal(a.Identity, b.Identity) &&
+		slices.EqualFunc(a.ServerRootCA, b.ServerRootCA, bytes.Equal)
 }
 
 func (ac *AuthCommMgr) createRemoteContext(stub *Stub) func() (*RemoteContext, error) {
--- a/node/comm/egress.go
+++ b/node/comm/egress.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package comm
 
 import (
+	"sync"
+
 	protos "github.com/hyperledger-labs/SmartBFT/smartbftprotos"
 	"github.com/hyperledger/fabric-lib-go/common/flogging"
 	cb "github.com/hyperledger/fabric-protos-go-apiv2/common"
@@ -22,15 +24,20 @@ type Egress struct {
 	RPC      *RPC
 	Logger   *flogging.FabricLogger
 	NodeList []uint64
+	lock     sync.RWMutex
 }
 
 // Nodes returns nodes from the runtime config
 func (e *Egress) Nodes() []uint64 {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
 	return e.NodeList
 }
 
 // SendConsensus sends the BFT message to the cluster
 func (e *Egress) SendConsensus(targetID uint64, m *protos.Message) {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
 	err := e.RPC.SendConsensus(targetID, bftMsgToClusterMsg(m))
 	if err != nil {
 		e.Logger.Warnf("Failed sending to %d: %v", targetID, err)
@@ -39,6 +46,8 @@ func (e *Egress) SendConsensus(targetID uint64, m *protos.Message) {
 
 // SendTransaction sends the transaction to the cluster
 func (e *Egress) SendTransaction(targetID uint64, request []byte) {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
 	msg := &ab.SubmitRequest{
 		Payload: &cb.Envelope{Payload: request},
 	}
@@ -55,4 +64,12 @@ func bftMsgToClusterMsg(message *protos.Message) *ab.ConsensusRequest {
 	return &ab.ConsensusRequest{
 		Payload: protoutil.MarshalOrPanic(message),
 	}
+}
+
+// Reconfigure updates the list of nodes and reconfigures the RPC
+func (e *Egress) Reconfigure(nodes []uint64, members []RemoteNode) {
+	e.lock.Lock()
+	defer e.lock.Unlock()
+	e.RPC.Reconfigure(members)
+	e.NodeList = nodes
 }
--- a/node/comm/egress.go
+++ b/node/comm/egress.go
@@ -24,20 +24,18 @@ type Egress struct {
 	RPC      *RPC
 	Logger   *flogging.FabricLogger
 	NodeList []uint64
-	lock     sync.RWMutex
+	lock     sync.Mutex
 }
 
 // Nodes returns nodes from the runtime config
 func (e *Egress) Nodes() []uint64 {
-	e.lock.RLock()
-	defer e.lock.RUnlock()
+	e.lock.Lock()
+	defer e.lock.Unlock()
 	return e.NodeList
 }
 
 // SendConsensus sends the BFT message to the cluster
 func (e *Egress) SendConsensus(targetID uint64, m *protos.Message) {
-	e.lock.RLock()
-	defer e.lock.RUnlock()
 	err := e.RPC.SendConsensus(targetID, bftMsgToClusterMsg(m))
 	if err != nil {
 		e.Logger.Warnf("Failed sending to %d: %v", targetID, err)
@@ -46,8 +44,6 @@ func (e *Egress) SendConsensus(targetID uint64, m *protos.Message) {
 
 // SendTransaction sends the transaction to the cluster
 func (e *Egress) SendTransaction(targetID uint64, request []byte) {
-	e.lock.RLock()
-	defer e.lock.RUnlock()
 	msg := &ab.SubmitRequest{
 		Payload: &cb.Envelope{Payload: request},
 	}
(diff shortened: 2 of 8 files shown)


HagarMeir:
## Follow-up needed: guard the synchronizer path against a panic on a closed BADB

### The gap

`SoftStop()` closes the `BADB` (and stops the `Synchronizer`/metrics). The **block-in-`Deliver`** mechanism protects the SmartBFT controller/viewchanger goroutine from touching a closed BADB during reconfiguration — but it does **not** protect the **BFT synchronizer goroutine**, which reaches `SoftStop` by a different route and is never parked.

### When/how this can happen

A consenter that fell behind (e.g. was down while the cluster committed a config block) catches up via the BFT synchronizer. The sync pull loop in `getBlocksFromSyncBuffer` (`node/consensus/synchronizer/synchronizer_bft.go`) processes pulled blocks one by one:

```
for seq <= targetSeq {
    block := PullBlock(seq)
    WriteConfigBlock(block) / WriteBlockSync(block)  // -> Arma.Index() -> BADB.Put()
    OnCommit(block)                                   // -> UpdateStateAndRuntimeConfig()
    seq++
}
```

When `OnCommit` → `UpdateStateAndRuntimeConfig` sees a **new config block**, it spawns `go processNewConfigBlock(...)`, which calls `SoftStop()` (→ `BADB.Close()`) — **asynchronously**. The pull loop does not wait; it continues to the next iteration and can call `WriteBlockSync`/`WriteConfigBlock` → `Arma.Index` → **`BADB.Put()` on the now-closed leveldb → panic.**

Concretely: a node syncing a range where a config block sits **mid-stream** (not last), e.g. pulling blocks `[100..105]` where block `102` is the config block. After `102` spawns the reconfig, iterations `103..105` keep writing to BADB while `SoftStop` closes it.

Note this is masked today because in the common case the config block is the **last** block in the sync range, so the loop ends right after spawning the reconfig. It is not guaranteed.

Also relevant: the sync-path `Arma.Index` (via `ConsenterSupportAdapter.indexAndWrite`) does **not** hold `c.lock`, unlike `Deliver`'s indexing — so even lock-based coordination in the reconfig teardown has a hole on this path.

### Sketch of the fix

Make `SoftStop` stop *freeing* resources and move the teardown to where it's serialized with the config swap:

- `SoftStop()` only flips state + rejects new input (does **not** close BADB / stop Synchronizer / stop metrics).
- The teardown moves to:
  - `stopAndReconfigure` (dynamic restart): free `Synchronizer`/`BADB`/metrics under `c.lock`, immediately before `configureConsensus` installs fresh instances.
  - the pending-admin branch of `processNewConfigBlock`: free them after `BFT.Stop()` has halted callbacks.
  - `Stop()`: free them unless already in `StatePendingAdmin` (guarded so we free exactly once).

Given the synchronizer path doesn't hold `c.lock` around `Arma.Index`, we should also confirm the teardown ordering (specifically that `Synchronizer.Stop()` fully drains the pull loop before `BADB.Close()`), or add locking to the sync-path index, so the fix is airtight and not just relying on the config-block-is-last coincidence.

tock-ibm (commented):
## Dynamic reconfiguration: reuse SmartBFT instead of full teardown

This reworks the consensus node's in-place reconfiguration so SmartBFT is *reused* across a config change rather than recreated. `Deliver` now spawns `processNewConfigBlock` and blocks (after releasing `c.lock`) on a per-invocation `reconfigDone` channel until the new config is installed; `SoftStop` no longer stops the BFT (its two exit branches manage the lifecycle); a node-lifetime `ReconfigAbort` channel lets `Stop` unblock a parked `Deliver`; and `Egress`/`RPC` gain `Reconfigure` + locking so the communicator/streams can be swapped under a live BFT.

The design is sound and the inline documentation of the goroutine ownership and lock-release ordering is genuinely excellent — it made this reviewable. The `sameRemoteNode` change also quietly fixes a latent stale-connection bug (see inline).

### Main asks before merge

1. **Confirm the `c.BFT.Synchronizer` reassignment on a live BFT is race-free.** The dynamic path (`configureBFT=false`) reuses the running BFT but `configureConsensus` still reassigns `c.BFT.Synchronizer` / `c.Synchronizer`, and SmartBFT reads its `Synchronizer` field without `c.lock`. Please run the `test/reconfig` suite with `-race` and confirm no SmartBFT goroutine can touch `Synchronizer` during the window. (inline on consensus_builder.go)
2. **Add targeted tests for the concurrency-critical edges** — the only test delta is initializing `ReconfigAbort` in a helper. There's no coverage of `Stop` unblocking a parked `Deliver` (the `ReconfigAbort` → `InLatestDecision:false` path) or of `sameRemoteNode` skip-vs-reactivate. These abort/teardown edges are exactly where this class of change breaks.

### Minor
- `SoftStop` removing `c.BFT.Stop()` is only safe because `processNewConfigBlock` is now its sole caller and both branches handle the BFT — worth a comment noting the invariant. (inline)
- Redundant `rpc := e.RPC` copy under `RLock` in `Egress` (inline).
- `RPC.Reconfigure` replaces `StreamsByType` wholesale — confirm no leak of abandoned in-flight streams (inline).

Overall LGTM pending (1) and (2).

node/comm/commauth.go
```diff
@@ -136,23 +138,40 @@ func (ac *AuthCommMgr) getOrCreateMapping() MemberMapping {
 func (ac *AuthCommMgr) updateStubInMapping(mapping MemberMapping, node RemoteNode) {
 	stub := mapping.ByID(node.ID)
 	if stub == nil {
-		ac.Logger.Infof("Allocating a new stub for node %v with endpoint %v", node.ID, node.Endpoint)
+		ac.Logger.Infof("Allocating a new stub for node %d with endpoint %v", node.ID, node.Endpoint)
 		stub = &Stub{}
+	} else if stub.Active() && sameRemoteNode(stub.RemoteNode, node) {
```
tock-ibm, line 143: Nice — this also fixes a latent bug, not just an optimization. The old `updateStubInMapping` returned early whenever the stub was already `Active()`, so it overwrote `stub.RemoteNode` with the new data but kept the **stale connection** even if the endpoint or certs changed. The new code disconnects + deactivates + reactivates whenever the data differs, and only short-circuits when the data is truly identical via `sameRemoteNode`. Good change.

node/comm/egress.go
```diff
@@ -22,23 +24,34 @@ type Egress struct {
...
+	e.lock.RLock()
+	defer e.lock.RUnlock()
 	return e.NodeList
 }
 
 // SendConsensus sends the BFT message to the cluster
 func (e *Egress) SendConsensus(targetID uint64, m *protos.Message) {
-	err := e.RPC.SendConsensus(targetID, bftMsgToClusterMsg(m))
+	e.lock.RLock()
+	rpc := e.RPC
```
tock-ibm: `e.RPC` is never reassigned — only its internals mutate via `RPC.Reconfigure` (which has its own `s.lock`). Copying the pointer under the egress `RLock` serializes sends against `Reconfigure` but doesn't protect the pointer itself. It's harmless, but the intent is unclear; consider either dropping the copy or adding a one-line comment ("serialize sends against the stream reset in Reconfigure"). Also worth confirming that serializing every consensus send against reconfig is intended, given `RPC` already locks internally.

node/comm/rpc.go
```diff
@@ -72,11 +72,19 @@ func (ot OperationType) String() string {
 	return "consensus"
 }
 
-// SendConsensus passes the given ConsensusRequest message to the raft.Node instance.
+// Reconfigure configures the communicator of the RPC and resets the streams.
+func (s *RPC) Reconfigure(members []RemoteNode) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	s.StreamsByType = NewStreamsByType()
```
tock-ibm, line 81: `Reconfigure` replaces `StreamsByType` wholesale, abandoning any in-flight streams without an explicit cancel/unmap. This is presumably fine because `Comm.Configure` deactivates stubs (which `Abort()`s their `RemoteContext`), but please confirm there's no stream/goroutine leak on the abandoned map. The doc comment says "resets the streams" — "replaces the stream map" would be more precise.

node/consensus/consensus_builder.go
```diff
@@ -134,8 +135,12 @@ func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConf
 	}
 	c.ConfigAckReceiver = configack.NewReceiver(c.Logger, shards)
 
-	c.BFT = createBFT(c, metadata, lastProposal, lastSigs, nodeConfig.WALDir)
-	setupComm(c)
+	if configureBFT {
+		c.BFT = createBFT(c, metadata, lastProposal, lastSigs, nodeConfig.WALDir)
+		setupComm(c)
+	} else {
+		configureComm(c)
```
tock-ibm, line 142: Race concern on the dynamic path: with `configureBFT=false` the existing BFT keeps running, but `configureConsensus` still creates a new synchronizer and assigns it to `c.BFT.Synchronizer` and `c.Synchronizer` (lines ~167-168, just below this hunk). SmartBFT reads its `Synchronizer` field without `c.lock` (e.g. `pkg/consensus.Sync()` → `c.Synchronizer.Sync()`), and the viewchanger goroutine can drive a sync independently of the controller goroutine parked in `Deliver`. Even though `SoftStop` calls `c.Synchronizer.Stop()` first, reassigning a field on a live BFT looks racy. Please confirm no SmartBFT goroutine can touch `Synchronizer` between the `reconfigDone` block and the internal `reconfigChan` reconfig, ideally verified under `-race`.

node/consensus/consensus.go
```diff
@@ -945,11 +979,28 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
...
+	// blocking it prevents that goroutine from proposing/verifying against the stale config,
+	// and defers SmartBFT's own internal reconfig (sent on reconfigChan after we return)
+	// until ours is done. On abort (Stop or pending-admin teardown) we return
+	// InLatestDecision:false so the unbuffered reconfigChan send in pkg/consensus is skipped.
+	if reconfigDone != nil {
+		select {
+		case <-reconfigDone:
+			// dynamic reconfiguration installed the new config, or we parked in pending-admin
+		case <-c.ReconfigAbort:
+			return smartbft_types.Reconfig{InLatestDecision: false}
```
tock-ibm, line 1016: This abort path (returning `InLatestDecision:false` so the unbuffered `reconfigChan` send in `pkg/consensus` is skipped) is the subtlest part of the change and currently has no test coverage. Please add a unit test that stops the node while a `Deliver` is parked on `reconfigDone` and asserts it returns here rather than deadlocking `bft.Stop()`'s goroutine-join.

node/consensus/consensus.go
```diff
@@ -244,7 +272,6 @@ func (c *Consensus) SoftStop() {
 
 	c.Logger.Infof("Soft stopping consensus node")
```
tock-ibm, line 284: Removing `c.BFT.Stop()` from `SoftStop` is only safe because `processNewConfigBlock` is now the sole caller and both of its branches manage the BFT lifecycle (dynamic path reuses it; admin path stops it explicitly after closing `done`). That invariant is load-bearing but implicit — consider a short comment here noting the BFT is intentionally left running for the caller to manage.

node/comm/commauth.go
HagarMeir: Correct — the short-circuit now only happens when `sameRemoteNode` confirms the data is identical, so a changed endpoint/cert always triggers disconnect + reactivate. No change needed here.

node/comm/egress.go
HagarMeir: You're right that the pointer copy protected nothing — `e.RPC` is never reassigned (only `RPC.Reconfigure` mutates its internals, under `RPC`'s own lock). Dropped the copy and the egress lock from both send paths entirely; the send is already serialized against reconfiguration at the RPC layer. The egress lock now only guards `NodeList` (written by `Reconfigure`, read by `Nodes`), so I also downgraded it from `RWMutex` to `Mutex`.

node/comm/rpc.go
HagarMeir: Confirmed no leak: the abandoned streams are aborted out-of-band — `Comm.Configure` deactivates the stubs and aborts their `RemoteContext`, and `cleanCanceledStreams` prunes canceled entries on the next `mapStream`. Reworded the doc comment to "replaces the stream map" and noted the out-of-band abort.

node/consensus/consensus.go
HagarMeir: Added a comment at the end of `SoftStop` noting the BFT is intentionally left running, and that `processNewConfigBlock` (the sole caller) owns the BFT lifecycle in both branches.

node/consensus/consensus.go
HagarMeir: Added `TestConsensusStopDuringReconfigAckWait` (in `consensus_real_reconfig_test.go`) covering exactly this. It parks every node mid-reconfiguration by submitting a valid config update and deliberately not sending the acks, so each node decides the config block and then blocks in `ApplyConfig`'s `WaitForAllAcks` with `Deliver` parked on `reconfigDone`. It waits for `StateSoftStopped`, then asserts `Stop()` returns within a timeout rather than hanging on `bft.Stop()`'s goroutine-join.

I confirmed the test actually catches the deadlock: with the `close(c.ReconfigAbort)` in `Stop` removed, the test fails with `Stop() ... did not return; Deliver is likely deadlocked waiting on reconfig`. With the abort branch in place it passes (~7s).

I used the real-BFT harness rather than a unit test on purpose — the ack-wait parks the node naturally and deterministically, whereas a unit-level test would need to fake the config-block decision and pause reconfig at exactly the right point with no clean seam.

node/consensus/consensus_builder.go
HagarMeir: Confirmed — this is a real data race, not just a lint concern. On the dynamic path (`configureBFT=false`) the BFT keeps running, and `configureConsensus` reassigns `c.BFT.Synchronizer`. SmartBFT reads that field without a lock in `pkg/consensus.Sync()` (`c.Synchronizer.Sync()`), and the viewchanger goroutine can drive a sync independently of the controller goroutine parked in `Deliver`, so the read can race our write. `SoftStop`'s `Synchronizer.Stop()` stops the old synchronizer's goroutines but doesn't stop SmartBFT from *calling* `Sync()` again, so it doesn't close the window.

**Recommended fix — a `synchronizerHolder`.** The `bft.Synchronizer` interface is just `Sync() bft.SyncResponse`, so we can insert a stable indirection that SmartBFT holds once and never sees reassigned:

```go
type synchronizerHolder struct {
    mu    sync.RWMutex
    inner SynchronizerWithStop
}
func (h *synchronizerHolder) Sync() smartbft_types.SyncResponse {
    h.mu.RLock(); s := h.inner; h.mu.RUnlock()
    return s.Sync()
}
func (h *synchronizerHolder) Stop() {
    h.mu.RLock(); s := h.inner; h.mu.RUnlock()
    s.Stop()
}
func (h *synchronizerHolder) swap(s SynchronizerWithStop) {
    h.mu.Lock(); h.inner = s; h.mu.Unlock()
}
```

- `c.BFT.Synchronizer` (and arma's `c.Synchronizer`) is set to the holder **once**, at BFT creation, and never reassigned.
- On the dynamic path, `configureConsensus` calls `holder.swap(newSynchronizer)` instead of reassigning the field.
- Reads (`Sync()`/`Stop()`) and the swap are serialized by the holder's own `RWMutex`, so there's no unsynchronized access — and no change to vendored SmartBFT.

I'd rather not land this in the already-large PR; I'll handle it as a follow-up. Tracking it here for now.

node/comm/egress.go
```diff
@@ -22,10 +24,13 @@ type Egress struct {
 	RPC      *RPC
 	Logger   *flogging.FabricLogger
 	NodeList []uint64
+	lock     sync.Mutex
```
tock-ibm, line 27: Nit: it seems that NodeList  can be retrieved from RPC, which is thread safe. Thus Nodes() can be delegated to RPC and save the lock in Egress.

node/consensus/consensus_builder.go
```diff
@@ -55,12 +56,12 @@ func CreateConsensus(nodeConfig *node_config.ConsenterNodeConfig, config *ord_co
...
 	}
 
-	c.configureConsensus(nodeConfig, config, lastConfigBlock, configUpdateProposer)
+	c.configureConsensus(nodeConfig, config, lastConfigBlock, configUpdateProposer, true)
 
 	return c
 }
 
-func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConfig, config *ord_config.Configuration, lastConfigBlock *common.Block, configUpdateProposer policy.ConfigUpdateProposer) {
+func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConfig, config *ord_config.Configuration, lastConfigBlock *common.Block, configUpdateProposer policy.ConfigUpdateProposer, configureBFT bool) {
```
tock-ibm: configureBFT -> createBFT or reverse the `if`; document the method

node/consensus/consensus_builder.go
```diff
@@ -426,28 +421,43 @@ func setupComm(c *Consensus) {
 			Id:       uint32(node.PartyID),
 		})
 	}
-	c.ConfigureNodeCerts(consenterConfigs)
+	c.ClusterService.ConfigureNodeCerts(consenterConfigs)
```
tock-ibm, line 432: Within ConfigureNodeCerts(consenterConfigs), verify what happens if the Identity field (cert) of an existing ID (uint64) changes. Does it invalidate an existing incoming authorized stream?

node/consensus/consensus_builder.go
HagarMeir: done

node/comm/egress.go
HagarMeir: This I will leave to another PR

node/consensus/consensus_builder.go
HagarMeir: added to the issue to check after this PR

tock-ibm (approved):

PR #977, opened 2026-07-02
fix panic issues in VerifyConsenterSig

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/978

### DIFF
--- a/node/consensus/consensus.go
+++ b/node/consensus/consensus.go
@@ -481,13 +481,34 @@ func (c *Consensus) VerifyRequest(req []byte) (smartbft_types.RequestInfo, error
 // It returns the auxiliary data in the signature
 // (from SmartBFT API)
 func (c *Consensus) VerifyConsenterSig(signature smartbft_types.Signature, prop smartbft_types.Proposal) ([]byte, error) {
+	// Validate signature.Value is not nil to prevent panic in asn1.Unmarshal
+	if signature.Value == nil {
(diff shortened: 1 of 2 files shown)


node/consensus/consensus.go
```diff
@@ -481,13 +481,34 @@ func (c *Consensus) VerifyRequest(req []byte) (smartbft_types.RequestInfo, error
 // It returns the auxiliary data in the signature
 // (from SmartBFT API)
 func (c *Consensus) VerifyConsenterSig(signature smartbft_types.Signature, prop smartbft_types.Proposal) ([]byte, error) {
+	// Validate signature.Value is not nil to prevent panic in asn1.Unmarshal
```
tock-ibm, line 484: It would be easier to test if all the new code is extracted to a function that simple extract a valis values and msgs, or emits an error.

node/consensus/consensus.go
tock-ibm, line 1: please associate an issue with this fix

HagarMeir: ok just did: https://github.com/hyperledger/fabric-x-orderer/issues/978

node/consensus/consensus.go
HagarMeir: ok but I have already tested it and this can be done some other time if we ever need it again

tock-ibm (approved):

PR #984, opened 2026-07-05
consensus synchronizer BFT deliverer initialization

DorKatzelnick:
use boot config block to initialize BFT deliverer in consensus synchronizer

fix for #986

### DIFF
--- a/node/consensus/consensus_builder.go
+++ b/node/consensus/consensus_builder.go
@@ -146,6 +146,7 @@ func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConf
 		&ConsenterSupportAdapter{consensus: c}, // support ConsenterSupport,
 		nodeConfig.BCCSP,
 		&comm.PredicateDialer{Config: c.clientConfig()},
+		lastConfigBlock, // lastConfigBlock is the config block that was used to join the cluster
 	)
 	c.Logger.Info("Created a BFT Synchronizer")
 
--- a/node/consensus/synchronizer/synchronizer_bft.go
+++ b/node/consensus/synchronizer/synchronizer_bft.go
@@ -54,6 +54,7 @@ type BFTSynchronizer struct {
 	VerifierFactory     VerifierFactory
 	BFTDelivererFactory BFTDelivererFactory
 	Logger              *flogging.FabricLogger
+	JoinConfigBlock     *common.Block // JoinConfigBlock is the config block that was used to join the cluster.
 
 	mutex    sync.Mutex
 	syncBuff *SyncBuffer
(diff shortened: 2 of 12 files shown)


node/consensus/synchronizer/synchronizer_factory.go
```diff
@@ -93,8 +94,9 @@ func (*SynchronizerCreator) CreateSynchronizer(
 	support ConsenterSupport,
 	bccsp bccsp.BCCSP,
 	clusterDialer *comm.PredicateDialer,
+	bootConfigBlock *cb.Block,
```
HagarMeir: maybe add some documentation

node/consensus/synchronizer/synchronizer_bft.go
```diff
@@ -242,18 +243,18 @@ func (s *BFTSynchronizer) computeTargetHeight(heights []uint64) uint64 {
...
+	ledgerLastConfigBlock, err := s.Support.LastConfigBlock(ledgerLastBlock)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to retrieve last config block")
 	}
 	blockOps := &utils.CommonConfigBlockOperations{}
-	lastConfigEnv, err := blockOps.ConfigFromBlock(lastConfigBlock)
+	currentConfigEnv, err := blockOps.ConfigFromBlock(s.BootConfigBlock)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to retrieve last config envelope")
 	}
```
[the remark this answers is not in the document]
HagarMeir: you could check if nil before and return an error 
also change the err to not include "last"

node/consensus/synchronizer/synchronizer_factory.go
DorKatzelnick: added in synchronizer creation in consensus_builder

node/consensus/synchronizer/synchronizer_bft.go
```diff
@@ -242,18 +243,31 @@ func (s *BFTSynchronizer) computeTargetHeight(heights []uint64) uint64 {
...
+	ledgerLastBlock := s.Support.Block(startHeight - 1)
+	ledgerLastConfigBlock, err := s.Support.LastConfigBlock(ledgerLastBlock)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to retrieve last config block")
 	}
+	if s.JoinConfigBlock == nil {
+		return nil, errors.New("join config block is nil")
+	}
+
+	s.Logger.Infof("Creating BFTDeliverer, last block in ledger: %d, last config block in ledger: %d, join config block: %d", ledgerLastBlock.Header.Number, ledgerLastConfigBlock.Header.Number, s.JoinConfigBlock.Header.Number)
```
HagarMeir: last block in ledger I think it is a decision 
just that the log needs to be clear

node/consensus/synchronizer/synchronizer_bft.go
```diff
@@ -242,18 +243,31 @@ func (s *BFTSynchronizer) computeTargetHeight(heights []uint64) uint64 {
...
+
+	var delivererInitializationBlock *common.Block
+	if s.JoinConfigBlock.Header.Number > ledgerLastConfigBlock.Header.Number {
+		delivererInitializationBlock = s.JoinConfigBlock
+	} else {
+		delivererInitializationBlock = ledgerLastConfigBlock
```
HagarMeir, line 262: I don't know if this will ever happen 
Maybe we can give a WARN or something if we think this will never happen

node/consensus/consensus_builder.go
```diff
@@ -42,7 +42,7 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-func CreateConsensus(nodeConfig *node_config.ConsenterNodeConfig, config *ord_config.Configuration, lastConfigBlock *common.Block, logger *flogging.FabricLogger, mainExitChan chan struct{}, signer Signer, configUpdateProposer policy.ConfigUpdateProposer) *Consensus {
+func CreateConsensus(nodeConfig *node_config.ConsenterNodeConfig, config *ord_config.Configuration, joinConfigBlock *common.Block, logger *flogging.FabricLogger, mainExitChan chan struct{}, signer Signer, configUpdateProposer policy.ConfigUpdateProposer) *Consensus {
```
HagarMeir: I think this is still the lastConfigBlock
In the case where the node failed and comes back up we take the last config from the ledger 
Please check
So it could be either a join or last config block

HagarMeir (approved):

Issue #641, opened 2026-02-23
Coordinate restart with ConfigAck API


tock-ibm:
In order to ensure that all the elements of a party receive the config block, the Consenter needs to wait for the party members to pull the new config block. We introduce the "ConfigAck" API, where the assembler, batcher and router notify the consenter when they had received the new config. Only then does the consenter of the party restart.

- [ ] Define ConfigAck API protos
- [ ] Send Acks from router batchers and assemblers
- [ ] Collect ConfigAcks in consenter

PR #985, opened 2026-07-05
config ack sender and receiver

MayRosenbaum:
issue #641


node/consensus/consensus.go
```diff
@@ -1275,7 +1275,7 @@ func (c *Consensus) UpdateStateAndRuntimeConfig(block *common.Block) smartbft_ty
 }
 
 // AckConfig handles ConfigAck RPC calls from party members (router, batchers, assembler).
-// TODO: implement
+// TODO: implement the following: check the client cert, add the ack to the ack receiver handler and send a response back
 func (c *Consensus) AckConfig(ctx context.Context, req *protos.ConfigAck) (*protos.ConfigAckResponse, error) {
-	return nil, fmt.Errorf("not implemented")
+	return &protos.ConfigAckResponse{}, nil
 }
```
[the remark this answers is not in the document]
MayRosenbaum: will be done in the integration PR

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+		ca.logger.Warnf("config ack has been received on sequence %d but the last acknowledged sequence is %d", request.ConfigSeq, ca.acks[clientID])
+		return fmt.Errorf("config ack has been received on sequence %d but the last acknowledged sequence is %d", request.ConfigSeq, ca.acks[clientID])
+	}
+
+	ca.acks[clientID] = uint64(request.ConfigSeq)
+	ca.logger.Infof("config ack has been received on sequence %d", request.ConfigSeq)
+
+	ca.signalChan <- struct{}{}
+	return nil
+}
```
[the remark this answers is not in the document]
MayRosenbaum: I will check the identity of the sender (including its certificate) before AddAck is called in the consensus side.

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (ca *ConfigAckHandler) ExpectAck(configSeq uint64) bool {
+	for {
+		select {
```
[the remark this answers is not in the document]
MayRosenbaum: AddAck send a signal to signalChan

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+	protos "github.com/hyperledger/fabric-x-orderer/node/protos/comm"
+)
+
+type ClientID struct {
+	NodeType protos.NodeType
+	Shard    types.ShardID
+}
+
+// ConfigAckHandler handle ConfigAck messages from party members (router, batchers, assembler)
+type ConfigAckHandler struct {
```
tock-ibm: The package is called `configack` and every file and structure is prefixed like that as well. This is called "stutter" in Go. (https://go.dev/blog/package-names).

Thus: 
configack.Sender
configack.Receiver
etc.

tock-ibm: This file is called _receiver.go but the struct is handler. lets harmonize this one way or the other.
e.g. receiver,go with Receiver struct etc...

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/hyperledger/fabric-x-orderer/common/types"
+
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	protos "github.com/hyperledger/fabric-x-orderer/node/protos/comm"
+)
+
+type ClientID struct {
```
tock-ibm: ID is confusing, how about ClientRole?
And add a stringer for easy reporting on logs.

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+	NodeType protos.NodeType
+	Shard    types.ShardID
+}
+
+// ConfigAckHandler handle ConfigAck messages from party members (router, batchers, assembler)
+type ConfigAckHandler struct {
+	acks   map[ClientID]uint64
+	logger *flogging.FabricLogger
+	lock   sync.Mutex
```
tock-ibm: Idiomatically, the lock comes before the fields it needs to protect.

```
lock
field1
field2
<space>
other fields

File: common/configack/config_ack_receiver.go [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+)
+
+type ClientID struct {
+	NodeType protos.NodeType
+	Shard    types.ShardID
+}
+
+// ConfigAckHandler handle ConfigAck messages from party members (router, batchers, assembler)
+type ConfigAckHandler struct {
+	acks   map[ClientID]uint64
```
tock-ibm: Document what this field holds and why.

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+	ca.logger.Infof("config ack handler is stopping")
+	ca.cancel()
+}
+
+// AddAck records an acknowledgment from a party member.
+func (ca *ConfigAckHandler) AddAck(request *protos.ConfigAck) error {
+	clientID := ClientID{
+		NodeType: request.NodeType,
+		Shard:    types.ShardID(request.Shard),
+	}
```
tock-ibm: do checks on the input - i.e. router shard == 0, etc

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+
+	ca.lock.Lock()
+	if uint64(request.ConfigSeq) != ca.acks[clientID]+1 {
```
tock-ibm: This is not safe, what if clientID is not in the map? first do the lookup and then check it exists.

tock-ibm: This is too strict, just check it is incoming >= current

common/configack/config_ack_receiver.go
```diff
@@ -0,0 +1,133 @@
...
+
+	ca.acks[clientID] = uint64(request.ConfigSeq)
+	ca.logger.Infof("config ack has been received on sequence %d", request.ConfigSeq)
+	ca.lock.Unlock()
+
+	ca.signalChan <- struct{}{}
+	return nil
+}
+
+func (ca *ConfigAckHandler) ExpectAck(configSeq uint64) bool {
```
tock-ibm: should not ExpectAck also receive a timeout context?

`ExpectAck(timeoutCtx Context, configSeq uint64) bool`

This way the consenter will continue after a timeout even if no ack is received?

common/configack/config_ack_sender.go
```diff
@@ -0,0 +1,192 @@
...
+		case <-ctx.Done():
+			return fmt.Errorf("sending config ack to consensus aborted: %w", ctx.Err())
+		case <-time.After(interval):
+			ca.logger.Debugf("Retry attempt #%d", numOfRetries)
+			numOfRetries++
+
+			err := ca.tryToHandleConfigAck(ctx, configSeq)
+			if err != nil {
+				interval = min(interval*2, ca.maxRetryInterval)
+				ca.logger.Errorf("Sending config ack to consensus failed: %v, trying again in: %s", err, interval)
```
tock-ibm: do this log also after the first failure above, make it a warning, transient failures can be expected.

common/configack/config_ack_sender.go
```diff
@@ -0,0 +1,192 @@
...
+				continue
+			}
+
+			ca.logger.Infof("config ack was successfully sent to consensus")
```
tock-ibm: add the sequence number in all logs

common/configack/receiver_test.go
tock-ibm, line 1: check also illegal input value, i.e. non existing shards etc., illegal types, and so on

common/configack/config_ack_receiver.go
MayRosenbaum: Do we enable a case in which config ack has been received on sequence 3 but the last acknowledged sequence is 1 ? 
If not, we need to keep the "+1"

common/configack/config_ack_receiver.go
MayRosenbaum: When it is a router or an assembler, we don't care about the shards, I think

tock-ibm (approved):

Issue #37, opened 2025-07-09
Consensus - Equivocation by primary batcher


tock-ibm:
There is an equivocation rule inside `core/state.go`, however it is currently commented out with a note that there are false positives.

tock-ibm:
### Comments from original issue:

---

**Comment from @HagarMeir:**

https://github.ibm.com/decentralized-trust-research/ARMA/issues/152

---

tock-ibm:
What's optionally left to do is to:
- check for primary self-signed equivocation even if they are not propagated to secondaries, 
- include the term in the BAF for easy filtering of secondary's bogus self signed BAFs (terms from the future, term != primary, etc).

PR #989, opened 2026-07-06
add DetectEquivocation rule

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/37

### DIFF
--- a/node/consensus/state/state.go
+++ b/node/consensus/state/state.go
@@ -726,7 +726,12 @@ func DetectEquivocation(s *State, _ types.ConfigSequence, l *flogging.FabricLogg
 	// and their digest was added to the BatchAttestationDB.
 
 	// <seq, shard, primary> --> { digest -->  signer }
-	m := batchAttestationVotesByDigests(s)
+	// Only count BAFs where Signer != Primary: a BAF where the signer claims to be
+	// the primary (Signer == Primary) is self-signed and can be forged by any node to
+	// manufacture false equivocation evidence.  Genuine attestations from secondaries
+	// always have Signer != Primary (the primary's own participation is captured by the
+	// empty PrimarySignature field, not by it re-broadcasting its own BAF as a voter).
+	m := batchAttestationVotesByDigestsExcludingSelfSigned(s)
 
 	// Sort votes for deterministic processing: by shard, then primary, then seq.
 	votes := make([]batchAttestationVote, 0, len(m))
@@ -801,6 +806,33 @@ func batchAttestationVotesByDigests(s *State) map[batchAttestationVote]map[strin
 	m := make(map[batchAttestationVote]map[string][]types.PartyID)
 
 	for _, baf := range s.Pending {
+		currentVote := batchAttestationVote{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary()}
+
+		digests2signers, exists := m[currentVote]
+		if !exists {
+			digests2signers = make(map[string][]types.PartyID)
+			m[currentVote] = digests2signers
+		}
+
+		hexDigest := hex.EncodeToString(baf.Digest())
+		digests2signers[hexDigest] = append(digests2signers[hexDigest], baf.Signer())
+	}
+	return m
+}
+
+// batchAttestationVotesByDigestsExcludingSelfSigned is like batchAttestationVotesByDigests
+// but skips BAFs where Signer() == Primary().  Such self-signed BAFs cannot serve as
+// secondary attestations and are filtered here to prevent a non-primary from
+// manufacturing equivocation evidence by sending two BAFs with the same Primary() == Signer()
+// but different digests.
+func batchAttestationVotesByDigestsExcludingSelfSigned(s *State) map[batchAttestationVote]map[string][]types.PartyID {
+	m := make(map[batchAttestationVote]map[string][]types.PartyID)
+
+	for _, baf := range s.Pending {
+		if baf.Signer() == baf.Primary() {
+			continue
+		}
+
 		currentVote := batchAttestationVote{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary()}
 
 		digests2signers, exists := m[currentVote]
--- a/node/consensus/state/state.go
+++ b/node/consensus/state/state.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"math"
 	"slices"
+	"strings"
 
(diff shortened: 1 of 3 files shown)


node/consensus/state/state.go
```diff
@@ -714,46 +715,62 @@ func FilterPendingEventsWithDiffConfigSeq(s *State, configSeq types.ConfigSequen
 	s.Complaints = filteredComplaints
 }
 
-func DetectEquivocation(s *State, l *flogging.FabricLogger, _ ...ControlEvent) {
-	// We have a total of N parties per shard.
-	// We collect a quorum of signatures and then wait for f+1 identical ones.
-	// If we can't collect such, it means the primary equivocated.
+func DetectEquivocation(s *State, _ types.ConfigSequence, l *flogging.FabricLogger, _ ...ControlEvent) {
```
tock-ibm, line 719: Please document what this detects and what it does not detect. 
Especially discuss corner cases like: if the primary causes (f+1) batchers to commit to batch B' in one round and then (f+1) batchers to commit to batch B'' in a second round, etc.

node/consensus/state/state.go
```diff
@@ -714,46 +715,62 @@ func FilterPendingEventsWithDiffConfigSeq(s *State, configSeq types.ConfigSequen
...
+			hexDigest = hexDigest[:8]
+		}
+		fmt.Fprintf(&summary, "digest=%s (signers=%d)", hexDigest, len(signers))
+		count++
+		if count >= 3 { // Limit to first 3 digests for brevity
+			if len(digest2signers) > 3 {
+				fmt.Fprintf(&summary, ", ... and %d more", len(digest2signers)-3)
+			}
+			break
+		}
```
tock-ibm: I think that if this ever happens, we can have the full report. That is, all the digest, and explicitly who reported (signer) of that BAF. no need to truncate anything. This is rare and severe, so lets have the full info in the log.

node/consensus/state/state.go
```diff
@@ -714,46 +715,62 @@ func FilterPendingEventsWithDiffConfigSeq(s *State, configSeq types.ConfigSequen
...
-
-		var totalSigners int
-
-		for _, signers := range digest2signers {
-			totalSigners += len(signers)
-			if len(signers) >= int(s.Threshold) {
-				foundThreshold = true
-				break
+	// For each <seq, shard, primary> check if it has multiple different digests
+	for vote, digest2signers := range m {
```
tock-ibm: Note that iteration order on a map can change from node to node, and from one time to the next. This maybe a cause for non-deterministic behavior among nodes, if there is more than one instance od equivocation in the map. One node may hit on instance A first, and the other on instance B.

node/consensus/state/state.go
```diff
@@ -714,46 +715,62 @@ func FilterPendingEventsWithDiffConfigSeq(s *State, configSeq types.ConfigSequen
...
+				"has %d different digests (%v). Primary has sent conflicting batches.",
+				vote.seq, vote.shard, vote.primary,
+				len(digest2signers), getDigestSummary(digest2signers))
+
+			// Rotate the primary in the affected shard
+			for i := range s.Shards {
+				if s.Shards[i].Shard == vote.shard {
+					l.Warnf("Rotating primary %d (term %d -> %d) in shard %d due to equivocation at sequence %d",
+						vote.primary, s.Shards[i].Term, s.Shards[i].Term+1, s.Shards[i].Shard, vote.seq)
+					s.Shards[i].Term++
```
tock-ibm, line 775: what if the primary has N equivocation events? the term++ is invoked N times which points back to the faulty primary. Detect a faulty primaries, in do term++ outside the loop, once.

node/consensus/state/state.go
tock-ibm: In addition, sort the output, so that we have detemenistic logs.

node/consensus/state/state.go
tock-ibm: https://stackoverflow.com/questions/55925822/why-are-iterations-over-maps-random

node/consensus/state/state.go
```diff
@@ -714,46 +715,83 @@ func FilterPendingEventsWithDiffConfigSeq(s *State, configSeq types.ConfigSequen
...
-				break
+			for i := range s.Shards {
+				if s.Shards[i].Shard == vote.shard {
+					l.Warnf("Rotating primary %d (term %d -> %d) in shard %d due to equivocation at sequence %d",
+						vote.primary, s.Shards[i].Term, s.Shards[i].Term+1, s.Shards[i].Shard, vote.seq)
+					s.Shards[i].Term++
+					rotatedShards[vote.shard] = struct{}{}
+					break
+				}
 			}
```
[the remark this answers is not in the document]
HagarMeir: good point!
instead of trying to calculate the primary here (I'm missing some key data) 
I will simply ignore self signed BAFs 
since if a primary (or secondary pretending) wants to send BAFs with different digests to itself this doesn't harm anyone but itself.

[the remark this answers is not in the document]
tock-ibm: How can secondaries implicate a true primary in equivocation? the BAF is double signed... those will drop immediately on entry.

A secondary can pretend to be a primary, and self-sign. 
A primary can send many self signed BAFs (with different digests) only to the consensus. I think we want to detect that.

In `s.Shards[i].Term` we have the current term, so why not match the offending votes against it?
Secondaries' self-signed can be dropped (with a warning) whereas primary equivocation causes a term change.

What am I missing?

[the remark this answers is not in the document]
HagarMeir: The secondaries cannot implicate a true primary in equivocation
But they can cause a term change because they can cause equivocation to be detected (by pretending to be a primary). 
The BAF does not contain the term but only the primary so unfortunately we cannot identify the party who is supposed to be the primary. 
The votes (BAFs) do not carry the term and so we cannot match against the current term in the state.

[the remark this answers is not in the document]
tock-ibm: But we can calculate whether the declared primary in a BAF matches the current term, it is not 1:1 but is good enough as a precaution, I think.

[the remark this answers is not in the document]
tock-ibm: We will continue with the current implementation which protects against a primary trying to equivocate against secondaries. "Self signed" equivocation by primary and secondaries cause garbage in the state.pending, and will be handled when we strengthen garbage collection.

tock-ibm (approved):

PR #993, opened 2026-07-07
BFT deliverer endpoints update 

DorKatzelnick:
change BFT deliverer to update the enpoints only for newer config blocks

### DIFF
--- a/common/deliverclient/blocksprovider/bft_deliverer.go
+++ b/common/deliverclient/blocksprovider/bft_deliverer.go
@@ -102,6 +102,7 @@ type BFTDeliverer struct {
 	lastBlockTime                  time.Time     // last block time
 	fetchFailureCounter            int           // counts the number of consecutive failures to fetch a block
 	fetchFailureTotalSleepDuration time.Duration // the cumulative sleep time from when fetchFailureCounter goes 0->1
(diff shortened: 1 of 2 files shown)


common/deliverclient/blocksprovider/bft_deliverer.go
```diff
@@ -419,18 +421,27 @@ func (d *BFTDeliverer) onBlockProcessingSuccess(blockNum uint64, channelConfig *
...
-		ordererConfig, err := extractOrdererConfig(d.ChannelID, channelConfig, d.CryptoProvider)
-		if err != nil {
-			d.Logger.Panicf("Failed to extract orderer config: %s", err)
-		}
+		seq := channelConfig.GetSequence()
+		if seq <= d.lastConfigSequence {
+			// This is a stale config block replayed while catching up. Applying its (older) endpoints would
+			// clobber the current set with an obsolete membership, so skip the refresh.
+			d.Logger.Debugf("Ignoring stale config block for endpoint refresh: sequence %d <= last applied %d", seq, d.lastConfigSequence)
+		} else {
```
HagarMeir: can you please just do an early return instead of else here?

tock-ibm (approved):

Issue #935, opened 2026-06-18
Verify local config on start


tock-ibm:
- [x] Verify local config every time the node starts
- [x] Enforce defaults on missing parameters
- [x] Enforce reasonable values on all parameters (e.g. no 0 seconds on timeouts, relation between related timeouts)
- [x] Whenever a default is set - print a log Info message

MayRosenbaum:
We should enforce reasonable defaults for all parameters during the `LoadLocalConfig` method, because after that point it is too late.
Move the checks from the `Extract*` methods into `LoadLocalConfig` and make sure we do not repeat the same checks in multiple places.

tock-ibm:
See comments: https://github.com/hyperledger/fabric-x-orderer/pull/998#pullrequestreview-4733795320

PR #998, opened 2026-07-13
Validate local config and apply defaults

moradna:
#935

### DIFF
--- a/config/local_config.go
+++ b/config/local_config.go
@@ -211,10 +211,12 @@ func LoadLocalConfig(filePath string) (*LocalConfig, string, error) {
 		return nil, "", fmt.Errorf("cannot load local node configuration, failed reading config yaml, err: %s", err)
 	}
 
-	role, err := validateNodeLocalConfigParams(nodeLocalConfig)
+	role, err := validateLocalConfig(nodeLocalConfig)
 	if err != nil {
 		return nil, "", err
 	}
+
+	applyLocalConfigDefaults(nodeLocalConfig, role)
 
 	tlsConfig, err := loadTLSCryptoConfig(&nodeLocalConfig.GeneralConfig.TLSConfig)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


config/local_config.go
```diff
@@ -375,3 +406,106 @@ func loadClusterCryptoConfig(cluster *ClusterYaml) (*Cluster, error) {
 		ReplicationPolicy: cluster.ReplicationPolicy,
 	}, nil
 }
+
+func applyLocalConfigDefaults(nodeLocalConfig *NodeLocalConfig, role string) {
+	applyGeneralDefaults(nodeLocalConfig)
+	applyNodeDefaults(nodeLocalConfig, role)
+}
+
+func applyGeneralDefaults(nodeLocalConfig *NodeLocalConfig) {
```
tock-ibm, line 415: where are the General.Cluster defaults?

tock-ibm: Defaults?

MaxRecvMsgSize int32 `yaml:"MaxRecvMsgSize,omitempty"`
MaxSendMsgSize int32 `yaml:"MaxSendMsgSize,omitempty"`
LogSpec string `yaml:"LogSpec,omitempty"`

tock-ibm (approved):
**Approved** with a few comments. Lets take all these comments in the next PR.

### Minor, non-blocking notes (below the reporting bar — optional)

1. The deleted TODO in LoadLocalConfig explicitly asked to "emit a log message indicating that the default value is being used" when filling defaults. The defaulting is implemented but silent — config/local_config.go imports no logger and LoadLocalConfig takes none. Consider adding default-applied log lines (or threading a logger) so operators can see which defaults kicked in. Low severity (observability, not correctness).

2. Doc-only: config/sample/local_config_consenter.yaml sets WALDir: .../smartbft/wal, which diverges from the FileStore-derived default now that the field is honored. Those samples have an empty FileStore and wouldn't pass the new validation anyway, but aligning the sample would avoid confusion.

config/local_config.go
moradna: Addressed in #1036

PR #1015, opened 2026-07-19
Add prefetch index size metric

moradna:
#414

### DIFF
--- a/node/assembler/metrics.go
+++ b/node/assembler/metrics.go
@@ -46,6 +46,13 @@ var (
 		LabelNames: []string{"party_id"},
 		Buckets:    []float64{.0001, .001, .002, .003, .004, .005, .01, .03, .05, .1, .3, .5, 1}, // TODO: adjust buckets after reviewing Grafana
 	}
+
+	prefetchIndexSizeOpts = metrics.GaugeOpts{
+		Namespace:  "assembler",
+		Name:       "prefetch_index_size_bytes",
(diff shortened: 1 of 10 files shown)


node/assembler/metrics.go
```diff
@@ -46,6 +46,13 @@ var (
 		LabelNames: []string{"party_id"},
 		Buckets:    []float64{.0001, .001, .002, .003, .004, .005, .01, .03, .05, .1, .3, .5, 1}, // TODO: adjust buckets after reviewing Grafana
 	}
+
+	prefetchIndexSizeOpts = metrics.GaugeOpts{
+		Namespace:  "assembler",
+		Name:       "prefetch_index_size_bytes",
+		Help:       "The current size of the assembler prefetch index for a shard in bytes.",
+		LabelNames: []string{"party_id", "shard_id"},
```
DorKatzelnick, line 54: is party_id always the assembler's party?
do we want to have this metric per shard, or for each partition (<primary,shard>)? or just have one gauge that sums all of the partitions for that assembler.

moradna: party_id is always the assembler's party ID. We calculate the prefetch index size per shard by summing its partitions, and each assembler reports that value with its own party_id

DorKatzelnick (approved):
add new metrics to Grafana dashboard (w. @moran-abilea )

tock-ibm (approved):

PR #1018, opened 2026-07-20
change batchers channel name

DorKatzelnick:
batchers channel name changed from shardXpartyY to shardXpartyY-\<channelID\>

#1017


node/batcher/batcher_deliver_service.go
```diff
@@ -87,9 +87,9 @@ type chainManager struct {
 }
 
 func (c *chainManager) GetChain(chainID string) deliver.Chain {
-	shardID, partyID, err := ledger.ChannelNameToShardParty(chainID)
+	shardID, partyID, _, err := ledger.ChannelNameToShardPartyChannelID(chainID)
```
tock-ibm: check that the incoming channel-ID matches what we have in config

node/ledger/batch_encoding.go
```diff
@@ -151,32 +152,64 @@ func NewFabricBatchFromBlock(block *common.Block) (*FabricBatch, error) {
...
+// less than 250 characters long.
+const maxChannelIDLength = 250
+
+// channelIDRegexp matches a legal channel ID: it must start with a lower case letter and contain only
+// lower case letters, digits, dots and dashes.
+var channelIDRegexp = regexp.MustCompile("^[a-z][a-z0-9.-]*$")
+
+// ValidateChannelID returns an error unless channelID is legal, i.e. it is all lower case, less than
+// 250 characters long, and matches the regular expression [a-z][a-z0-9.-]*.
+func ValidateChannelID(channelID string) error {
```
tock-ibm: We don't need to do it only here, but on the channelID defined in the genesis block as well (both).
@moradna 

Move this function to the config/verify package. 

Check where it is done in fabric to verify it is good. 
Unit Test it.

node/ledger/batch_encoding.go
```diff
@@ -151,32 +152,64 @@ func NewFabricBatchFromBlock(block *common.Block) (*FabricBatch, error) {
...
-		return 0, 0, errors.Errorf("cannot extract 'partyID' from channel name: %s, err: %s", channelName, err)
+		return 0, 0, "", errors.Errorf("cannot extract 'partyID' from channel name: %s, err: %s", channelName, err)
+	}
+
+	if err := ValidateChannelID(channelID); err != nil {
+		return 0, 0, "", errors.Errorf("cannot extract 'channelID' from channel name: %s, err: %s", channelName, err)
 	}
 
-	return types.ShardID(shardID), types.PartyID(partyID), nil
+	return types.ShardID(shardID), types.PartyID(partyID), channelID, nil
```
tock-ibm, line 195: shardID and partyID come out of atoi as integers, and downcast to uint16, which may cause overflow. Verify that the range is [0,maxuint16] before down casting

node/ledger/batch_encoding.go
DorKatzelnick: will take `ValidateChannelID` from  fabric-x-common/common/configtx/validator.go
which also how it is done in fabric.

node/ledger/batch_encoding.go
DorKatzelnick: now use `strconv.ParseUint` instead of `atoi`

tock-ibm (approved):

PR #1036, opened 2026-07-22
Extend local config defaults

moradna:
#935

### DIFF
--- a/config/sample/local_config_consenter.yaml
+++ b/config/sample/local_config_consenter.yaml
@@ -25,6 +25,7 @@ General:
 #
 ################################################################################
 FileStore:
+  Location: /var/dec-trust/production/orderer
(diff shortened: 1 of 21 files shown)


config/sample/local_config_consenter.yaml
tock-ibm, line 1: Can we have a "sample_validation_test.go" that tries to load all the samples local config files? just checking there is no panic. this way the samples will not diverge if we change anything in the local config structure. Lets do this in a separate PR.

tock-ibm (approved):
LGTM. Approved with one comment for next PR.

config/sample/local_config_consenter.yaml
moradna: Done in this #1044

PR #1037, opened 2026-07-22
equivocation digest leak

### LINKED ISSUE #1035: Consensus can commit a block against an under-attested digest when a primary equivocates
## Summary

When a shard primary equivocates (emits BAFs with different digests for the same `<seq, shard, primary>`), consensus can finalize a data block whose `DataHash`/`TXCount` come from a digest that **never reached the signature threshold**.

Because leader and followers run the same deterministic `Process` and agree on the same fragment, this is a **safety** issue (a block finalized against a digest lacking quorum attestation), not a liveness/view-change failure.

## Mechanism

1. `ExtractBatchAttestationsFromPending` (`node/consensus/state/state.go`) extracts a `<seq, shard, primary>` group if *at least one* of its digests reached threshold, but then copies **every** pending BAF for that key regardless of digest — including fragments carrying minority digests that never met threshold.

2. `aggregateFragments` (`node/consensus/consenter.go`) groups fragments keyed only on `{seq, shard}`, so a single group is an ordered mix of digests.

3. Consensus collapses each group to `ba[0]` on both the proposer path (`AssembleProposal`) and the verifier path (`VerifyProposal`) in `node/consensus/consensus.go`. `ba[0]` is simply the first BAF in `Pending` iteration order, which may be an under-attested digest. The block commits with `DataHash = ba[0].Digest()` and `TXCount = ba[0].TXCount()`.

Nothing downstream ever inspects `ba[1:]`.

## Interaction with `DetectEquivocation`

`DetectEquivocation` (`node/consensus/state/state.go`) does **not** close this window:

- It runs *before* extraction in the same `Process` pass.
- It only increments the shard term — it never removes the conflicting BAFs from `Pending`.
- The term rotation is forward-looking only. The current sequence is still extracted and committed against `ba[0]` in the same round, and the conflicting fragments are then consumed, so there is no later re-examination for that sequence.

So `DetectEquivocation` provides the audit signal and future primary rotation, but the conflicting sequence has already been finalized against a possibly-wrong digest.

## Proposed fix

Rather than the "explicit digest selection" the existing TODO suggests (choosing the digest with the most signatures), produce **one batch-attestation group per `<seq, shard, primary, digest>` that reaches threshold**:

- `ExtractBatchAttestationsFromPending` should extract only the fragments belonging to threshold-reaching digests, keyed per digest.
- `aggregateFragments` should key on `{seq, shard, primary, digest}` so every emitted group is single-digest, and `ba[0]` is guaranteed to be a threshold-attested fragment.

This makes the downstream `ba[0]` assumption correct by construction, and naturally handles the byzantine case where more than one digest reaches threshold — each becomes its own block rather than one silently winning by iteration order.

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/1035

### DIFF
--- a/node/consensus/consenter.go
+++ b/node/consensus/consenter.go
@@ -63,21 +63,35 @@ func (c *Consenter) Index(digests [][]byte) {
 	}
 }
 
+// bafGroupKey is the comparable form of types.BatchID (<shard, primary, seq, digest>): the
+// tuple that identifies a batch. digest is a string because BatchID.Digest() ([]byte) is not
+// map-key-comparable. Fragments are grouped by batch identity so that each block corresponds
+// to exactly one batch. In the honest case there is one batch per <seq, shard> (unchanged
+// behavior); distinct digests (an equivocating primary) or distinct primaries (e.g. the same
+// batch reproduced across a rotation) are different BatchIDs and therefore different blocks.
+// This guarantees every group is a single batch, so consumers that read only the first
+// fragment (ba[0]) always see a threshold-attested batch rather than a stale/minority digest
+// that happened to be first in Pending order.
+type bafGroupKey struct {
+	seq     types.BatchSequence
+	shard   types.ShardID
+	primary types.PartyID
+	digest  string
+}
+
+func bafKey(baf types.BatchAttestationFragment) bafGroupKey {
+	return bafGroupKey{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary(), digest: string(baf.Digest())}
+}
+
 func aggregateFragments(batchAttestationFragments []types.BatchAttestationFragment) [][]types.BatchAttestationFragment {
 	index := indexBAFs(batchAttestationFragments)
 
 	var attestations [][]types.BatchAttestationFragment
 
-	added := make(map[struct {
-		seq   types.BatchSequence
-		shard types.ShardID
-	}]struct{})
+	added := make(map[bafGroupKey]struct{})
 
 	for _, baf := range batchAttestationFragments {
-		key := struct {
-			seq   types.BatchSequence
-			shard types.ShardID
-		}{seq: baf.Seq(), shard: baf.Shard()}
+		key := bafKey(baf)
 
 		if _, added := added[key]; added {
 			continue
@@ -92,23 +106,12 @@ func aggregateFragments(batchAttestationFragments []types.BatchAttestationFragme
 	return attestations
 }
 
-func indexBAFs(batchAttestationFragments []types.BatchAttestationFragment) map[struct {
-	seq   types.BatchSequence
-	shard types.ShardID
-}][]types.BatchAttestationFragment {
-	index := make(map[struct {
-		seq   types.BatchSequence
-		shard types.ShardID
-	}][]types.BatchAttestationFragment)
+func indexBAFs(batchAttestationFragments []types.BatchAttestationFragment) map[bafGroupKey][]types.BatchAttestationFragment {
+	index := make(map[bafGroupKey][]types.BatchAttestationFragment)
 
 	for _, baf := range batchAttestationFragments {
-		key := struct {
-			seq   types.BatchSequence
-			shard types.ShardID
-		}{seq: baf.Seq(), shard: baf.Shard()}
-		fragments := index[key]
-		fragments = append(fragments, baf)
-		index[key] = fragments
+		key := bafKey(baf)
+		index[key] = append(index[key], baf)
 	}
 	return index
 }
--- a/node/consensus/consenter.go
+++ b/node/consensus/consenter.go
@@ -83,37 +83,26 @@ func bafKey(baf types.BatchAttestationFragment) bafGroupKey {
 	return bafGroupKey{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary(), digest: string(baf.Digest())}
 }
 
+// aggregateFragments groups fragments by batch identity (bafGroupKey), preserving the order in
+// which each group's key first appears. Each returned group holds all fragments of one batch.
 func aggregateFragments(batchAttestationFragments []types.BatchAttestationFragment) [][]types.BatchAttestationFragment {
-	index := indexBAFs(batchAttestationFragments)
-
-	var attestations [][]types.BatchAttestationFragment
-
-	added := make(map[bafGroupKey]struct{})
+	groups := make(map[bafGroupKey][]types.BatchAttestationFragment)
+	var order []bafGroupKey // keys in first-seen order (map iteration is unordered)
 
 	for _, baf := range batchAttestationFragments {
 		key := bafKey(baf)
-
-		if _, added := added[key]; added {
-			continue
+		if _, seen := groups[key]; !seen {
+			order = append(order, key)
 		}
+		groups[key] = append(groups[key], baf)
+	}
 
-		added[key] = struct{}{}
-
-		fragments := index[key]
-		attestations = append(attestations, fragments)
+	attestations := make([][]types.BatchAttestationFragment, 0, len(order))
+	for _, key := range order {
+		attestations = append(attestations, groups[key])
 	}
 
 	return attestations
-}
-
-func indexBAFs(batchAttestationFragments []types.BatchAttestationFragment) map[bafGroupKey][]types.BatchAttestationFragment {
-	index := make(map[bafGroupKey][]types.BatchAttestationFragment)
-
-	for _, baf := range batchAttestationFragments {
-		key := bafKey(baf)
-		index[key] = append(index[key], baf)
-	}
-	return index
 }
 
 func requestsToControlEvents(requests [][]byte) ([]state.ControlEvent, error) {
--- a/node/consensus/state/state.go
+++ b/node/consensus/state/state.go
@@ -41,6 +41,15 @@ type batchAttestationVote struct {
 	signer  types.PartyID
 }
 
+// thresholdDigestKey identifies a batch (the comparable form of types.BatchID:
+// <shard, primary, seq> via vote, plus digest) that reached the signature threshold.
+// digest is the hex-encoded digest, matching the inner key produced by
+// batchAttestationVotesByDigests.
+type thresholdDigestKey struct {
+	vote   batchAttestationVote
+	digest string
+}
+
 type State struct {
 	N          uint16
 	Quorum     uint16
@@ -691,28 +700,33 @@ func ExtractBatchAttestationsFromPending(s *State, l *flogging.FabricLogger) []t
 	// <seq, shard, primary> --> { digest -->  signer }
 	m := batchAttestationVotesByDigests(s)
 
-	batchAttestationsWithThreshold := make(map[batchAttestationVote]struct{})
+	// tuplesWithThreshold holds the <seq, shard, primary> tuples for which at least one
+	// digest reached threshold. It governs which pending BAFs are removed: once a tuple is
+	// decided, all of its pending fragments are cleared (including stale minority/equivocating
+	// digests), exactly as before.
+	tuplesWithThreshold := make(map[batchAttestationVote]struct{})
+
+	// thresholdDigests holds the specific <seq, shard, primary, digest> that individually
+	// reached threshold. Only fragments of these digests are extracted as attestations, so an
+	// under-attested (equivocating) digest can never back a committed block. If more than one
+	// digest for the same tuple reaches threshold (byzantine primary), each is extracted and
+	// downstream becomes its own block.
+	thresholdDigests := make(map[thresholdDigestKey]struct{})
 
 	for batchAttestation, digest2signers := range m {
-		var foundThreshold bool
+		l.Debugf("A total of %d digests were found for seq %d in shard %d with primary %d", len(digest2signers), batchAttestation.seq, batchAttestation.shard, batchAttestation.primary)
 
-		l.Debugf("A total of %d digests where found for seq %d in shard %d with primary %d", len(digest2signers), batchAttestation.seq, batchAttestation.shard, batchAttestation.primary)
-
-		for _, signers := range digest2signers {
+		for digest, signers := range digest2signers {
 			if len(signers) >= int(s.Threshold) {
-				foundThreshold = true
-				l.Debugf("Found threshold (%d > %d) of batch attestation fragments for shard %d, seq %d", len(signers), s.Threshold-1, batchAttestation.shard, batchAttestation.seq)
-				break
+				l.Debugf("Found threshold (%d >= %d) of batch attestation fragments for shard %d, seq %d, digest %s", len(signers), s.Threshold, batchAttestation.shard, batchAttestation.seq, digest)
+				tuplesWithThreshold[batchAttestation] = struct{}{}
+				thresholdDigests[thresholdDigestKey{vote: batchAttestation, digest: digest}] = struct{}{}
 			}
 		}
 
-		if !foundThreshold {
+		if _, ok := tuplesWithThreshold[batchAttestation]; !ok {
 			l.Debugf("Could not find a threshold of batch attestation fragments for shard %d, seq %d", batchAttestation.shard, batchAttestation.seq)
-			continue
 		}
-
-		batchAttestationsWithThreshold[batchAttestation] = struct{}{}
-
 	} // for all <seq, shard, primary>
 
 	var extracted []types.BatchAttestationFragment
@@ -721,13 +735,17 @@ func ExtractBatchAttestationsFromPending(s *State, l *flogging.FabricLogger) []t
 
 	// We iterate over the pending because we need deterministic processing
 	for _, baf := range s.Pending {
-		if _, exists := batchAttestationsWithThreshold[batchAttestationVote{
-			seq:     baf.Seq(),
-			shard:   baf.Shard(),
-			primary: baf.Primary(),
-		}]; !exists {
+		vote := batchAttestationVote{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary()}
+
+		if _, decided := tuplesWithThreshold[vote]; !decided {
+			// No digest for this tuple reached threshold yet; keep it pending.
 			newPending = append(newPending, baf)
-		} else {
+			continue
+		}
+
+		// The tuple is decided and thus removed from Pending. Only emit fragments whose digest
+		// actually reached threshold; minority/equivocating digests are dropped, not extracted.
+		if _, ok := thresholdDigests[thresholdDigestKey{vote: vote, digest: hex.EncodeToString(baf.Digest())}]; ok {
 			extracted = append(extracted, baf)
 		}
 	}
--- a/node/consensus/state/state.go
+++ b/node/consensus/state/state.go
@@ -34,20 +34,37 @@ var Rules = []Rule{
 	// CleanupOldAttestations, // TODO: fully test for byzantine failures
 }
 
-type batchAttestationVote struct {
+// batchKey identifies a batch by <shard, primary, seq>: the batch-identity portion of
+// types.BatchID with the digest excluded. It is the unit of equivocation detection and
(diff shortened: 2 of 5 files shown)


node/consensus/state/state.go
```diff
@@ -41,6 +41,15 @@ type batchAttestationVote struct {
...
 }
 
+// thresholdDigestKey identifies a batch (the comparable form of types.BatchID:
+// <shard, primary, seq> via vote, plus digest) that reached the signature threshold.
+// digest is the hex-encoded digest, matching the inner key produced by
+// batchAttestationVotesByDigests.
+type thresholdDigestKey struct {
+	vote   batchAttestationVote
+	digest string
+}
```
tock-ibm: I don't understand why we need both the "Vote" \<shard,primary,seq, signer\> and "Vote-with-Digest".

Moreover looking where we use the Vote, I see in CollectAndDeduplicateEvents that it will not let a BAF with <sh,pr,sq, dg=Y, signer> come it from the events into the state if \<sh,pr,sq, dg=X, signer\> exists in the pending already, as the "Vote" map is only keyed by \<sh,pr,sq, signer\>. That looks suspicious, and it interferes with equivocation detection in the next rule applied.

tock-ibm (commented):
To be continued...

node/consensus/state/state.go
HagarMeir: Thanks — both points were spot on. Addressed in the latest push:

**1. "Why both the Vote `<sh,pr,sq,signer>` and Vote-with-Digest?"**

You were right that one overloaded struct was doing double duty (and often with a silently-zero `signer`). Split into three purpose-named keys:

- `batchKey` = `<shard, primary, seq>` — batch identity; the unit of equivocation detection and extraction.
- `attestationKey` = `batchKey` + `signer` + `digest` — the dedup unit.
- `thresholdDigestKey` = `batchKey` + `digest` — a digest that reached threshold.

Each key now names the concept it encodes, so there's no more guessing what a zero `signer` means.

**2. The dedup interfering with equivocation detection — confirmed, and fixed.**

You were correct: `CollectAndDeduplicateEvents` keyed dedup on `<sh,pr,sq,signer>` *without* the digest, so once signer S had a pending BAF with digest X, a second BAF from S with digest Y was dropped as a "duplicate." That second digest is exactly the evidence `DetectEquivocation` needs, so an equivocating primary could escape rotation.

Fix: `digest` is now part of `attestationKey`, so the dedup unit is `<sh,pr,sq,signer,digest>`. An exact re-send is still dropped, but a signer's genuine second-digest report is retained and reaches `DetectEquivocation`.

This can't be abused to manufacture false equivocation: `verifyCE` requires every BAF with `Signer != Primary` to carry the primary's signature over its digest, and that's verified before the event hits the state rules. So only a genuinely equivocating primary can produce extra distinct digests — and it gets rotated the same round.

To be clear on scope: this was a fault-detection **liveness** gap, not a safety hole — the digest-leak fix already prevents an under-threshold digest from backing a committed block.

Added `TestCollectAndDeduplicateEvents_SameSignerDifferentDigests`, which fails on the old per-signer key and passes with the digest included.

One nuance worth flagging from writing that test: `CollectAndDeduplicateEvents` seeds its BAF dedup set from `s.Pending` once and never updates it inside the loop, so it only dedups incoming BAFs against *already-pending* ones — an exact duplicate arriving twice within the *same* round would not be caught here (the `Complaint` branch does dedup in-round; the BAF branch doesn't). A true "duplicate" is therefore a cross-round re-send.

In the assembled system that same-round case can't actually occur: `ControlEvent.ID()` hashes the full `<seq, configSeq, signer, primary, shard, digest>` tuple, and SmartBFT's request pool rejects a second request carrying the same `RequestInfo` (`ErrReqAlreadyExists`) before it ever reaches a proposal — so two identical BAFs can't both land in one round's events. So this is defense-in-depth resting on the pool's dedup rather than a live double-count, but it does mean the BAF branch leans on that upstream invariant where the `Complaint` branch is self-sufficient. The fix doesn't change any of this; just noting it since it's non-obvious and worth a comment in the code.

node/consensus/consenter.go
```diff
@@ -63,56 +63,48 @@ func (c *Consenter) Index(digests [][]byte) {
...
+// bafGroupKey is the comparable form of types.BatchID (<shard, primary, seq, digest>): the
+// tuple that identifies a batch. digest is a string because BatchID.Digest() ([]byte) is not
+// map-key-comparable. Fragments are grouped by batch identity so that each block corresponds
+// to exactly one batch. In the honest case there is one batch per <seq, shard> (unchanged
+// behavior); distinct digests (an equivocating primary) or distinct primaries (e.g. the same
+// batch reproduced across a rotation) are different BatchIDs and therefore different blocks.
+// This guarantees every group is a single batch, so consumers that read only the first
+// fragment (ba[0]) always see a threshold-attested batch rather than a stale/minority digest
+// that happened to be first in Pending order.
+type bafGroupKey struct {
```
tock-ibm, line 75: Adding `primary` **and** `digest` to the aggregation key can turn one committed batch into two blocks with the *same* `DataHash`. `BatchedRequests.Digest()` (common/types/batched_requests.go:92) hashes only the request payloads — not shard/primary/seq — so two primaries that batch the identical set of requests for the same `<seq, shard>` (the "same batch reproduced across a rotation" case this comment calls out) produce the identical digest. If both `<seq, shard, primary>` tuples independently reach threshold in the same round, `ExtractBatchAttestationsFromPending` emits fragments for both, and `aggregateFragments` groups them into **two** blocks whose `Header.DataHash` is the same digest.

The old code keyed grouping on `<seq, shard>` only and coalesced these into a single block. With the new key, the assembler will materialize the same batch content into two consecutive blocks → the same transactions delivered twice. Worth confirming the downstream assembler dedups by digest, or that two primaries can never both cross threshold at the same seq; otherwise this is a duplicate-delivery regression.

_Reported by Claude Code (opus-4.8[1m]) during code review._

node/consensus/state/state.go
```diff
@@ -444,11 +475,17 @@ func PrimaryRotateDueToComplaints(s *State, configSeq types.ConfigSequence, l *f
...
 
 func CollectAndDeduplicateEvents(s *State, configSeq types.ConfigSequence, l *flogging.FabricLogger, ces ...ControlEvent) {
-	shardsAndSequences := make(map[batchAttestationVote]struct{}, len(s.Pending))
+	// shardsAndSequences dedups incoming BAFs against ones already in Pending. It is seeded from
+	// Pending and, unlike the complaints set below, is not updated as BAFs are appended in this
+	// loop — so two identical BAFs arriving in the same round would both be kept here. That case
+	// is prevented upstream: ControlEvent.ID() hashes the full <seq, configSeq, signer, primary,
+	// shard, digest> tuple, and SmartBFT's request pool rejects a duplicate RequestInfo before it
+	// ever reaches a proposal, so identical BAFs cannot co-occur in one round's events.
+	shardsAndSequences := make(map[attestationKey]struct{}, len(s.Pending))
```
tock-ibm, line 484: Relaxing the dedup key from `<seq, shard, primary, signer>` to `<seq, shard, primary, signer, digest>` means a single signer may now contribute one pending BAF *per distinct digest* instead of exactly one per batch. When a primary equivocates with many digests, none of the minority digests reaches threshold, so the batch is never `decided` and its fragments are never removed by `ExtractBatchAttestationsFromPending`. `DetectEquivocation` only bumps the term; it does not prune these fragments, and `CleanupOldAttestations` is disabled (state.go:34). A byzantine primary colluding with `f` secondaries can therefore accumulate `signers × digests` never-cleaned entries in `s.Pending`, an amplification the old one-per-signer cap prevented. Consider pruning conflicting-digest fragments once equivocation is detected and the shard is rotated.

_Reported by Claude Code (opus-4.8[1m]) during code review._

node/consensus/consenter.go
```diff
@@ -63,56 +63,48 @@ func (c *Consenter) Index(digests [][]byte) {
...
+	seq     types.BatchSequence
+	shard   types.ShardID
+	primary types.PartyID
+	digest  string
+}
 
-	var attestations [][]types.BatchAttestationFragment
+func bafKey(baf types.BatchAttestationFragment) bafGroupKey {
+	return bafGroupKey{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary(), digest: string(baf.Digest())}
```
tock-ibm, line 83: Minor consistency nit: `bafKey` uses the raw digest bytes as the map-key string (`string(baf.Digest())`), whereas every other place in this PR that keys a map by digest (`attestationKeyOf`, `thresholdDigestKey`, `batchAttestationVotesByDigests` in state.go) uses `hex.EncodeToString(baf.Digest())`. Both work in isolation, but the two files now encode the same conceptual key differently; using `hex.EncodeToString` here too would keep the "digest as map key" convention uniform and avoid confusion for future readers comparing the two grouping paths.

_Reported by Claude Code (opus-4.8[1m]) during code review._

node/consensus/state/state.go
```diff
@@ -444,11 +475,17 @@ func PrimaryRotateDueToComplaints(s *State, configSeq types.ConfigSequence, l *f
 }
 
 func CollectAndDeduplicateEvents(s *State, configSeq types.ConfigSequence, l *flogging.FabricLogger, ces ...ControlEvent) {
```
tock-ibm, line 477: Document what is the intent of the function itself, please.
What is the input? What does it change? What are the expected changes?

For future refactoring: These rules would be much more understandable if they are simple methods of `State`, IMO.

node/consensus/state/state.go
```diff
@@ -444,11 +475,17 @@ func PrimaryRotateDueToComplaints(s *State, configSeq types.ConfigSequence, l *f
...
+	// loop — so two identical BAFs arriving in the same round would both be kept here. That case
+	// is prevented upstream: ControlEvent.ID() hashes the full <seq, configSeq, signer, primary,
+	// shard, digest> tuple, and SmartBFT's request pool rejects a duplicate RequestInfo before it
+	// ever reaches a proposal, so identical BAFs cannot co-occur in one round's events.
+	shardsAndSequences := make(map[attestationKey]struct{}, len(s.Pending))
 	complaints := make(map[ShardTerm]map[types.PartyID]struct{})
 
 	for _, baf := range s.Pending {
-		shardsAndSequences[batchAttestationVote{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary(), signer: baf.Signer()}] = struct{}{}
+		shardsAndSequences[attestationKeyOf(baf)] = struct{}{}
```
tock-ibm, line 488: move the seeding right below the definition of `shardsAndSequences`; and `complaints` just above its seeding. 
Define next to usage.

node/consensus/state/state.go
tock-ibm, line 499: Don't we need to dedup ConfigRequests as well?

tock-ibm, line 596: Again, document the function, as state above.

A method of acceptor `State` is more readable, IMO.

node/consensus/state/state.go
```diff
@@ -565,20 +602,19 @@ func DetectEquivocation(s *State, _ types.ConfigSequence, l *flogging.FabricLogg
...
 	// always have Signer != Primary (the primary's own participation is captured by the
 	// empty PrimarySignature field, not by it re-broadcasting its own BAF as a voter).
-	m := batchAttestationVotesByDigestsExcludingSelfSigned(s)
+	m := batchAttestationVotesByDigests(s, isNotSelfSigned)
 
-	// Sort votes for deterministic processing: by shard, then primary, then seq.
-	votes := make([]batchAttestationVote, 0, len(m))
-	for vote := range m {
-		votes = append(votes, vote)
+	// Sort batches for deterministic processing: by shard, then primary, then seq.
```
tock-ibm, line 612: `batchKey` is a struct. Add a compare method to it. This way this is much more easily tested.
(Also, change field order according to the sort order you claim here, just for aesthetic reasons).

node/consensus/state/state.go
```diff
@@ -642,77 +678,66 @@ func getDigestSummary(digest2signers map[string][]types.PartyID) string {
...
-// but skips BAFs where Signer() == Primary().  Such self-signed BAFs cannot serve as
-// secondary attestations and are filtered here to prevent a non-primary from
-// manufacturing equivocation evidence by sending two BAFs with the same Primary() == Signer()
-// but different digests.
-func batchAttestationVotesByDigestsExcludingSelfSigned(s *State) map[batchAttestationVote]map[string][]types.PartyID {
-	m := make(map[batchAttestationVote]map[string][]types.PartyID)
+// batchAttestationVotesByDigests groups pending BAFs as <batchKey> --> { digest --> signers },
+// including only fragments for which include(baf) is true. Digests are hex-encoded so they are
+// map-comparable.
+func batchAttestationVotesByDigests(s *State, include func(baf types.BatchAttestationFragment) bool) map[batchKey]map[string][]types.PartyID {
```
tock-ibm, line 684: This unification is good, except the use of function pointers (style). A boolean would be much clearer, IMO, with the evaluation `if self-signed` within the function.

node/consensus/state/state.go
```diff
@@ -642,77 +678,66 @@ func getDigestSummary(digest2signers map[string][]types.PartyID) string {
...
 			continue
 		}
 
-		currentVote := batchAttestationVote{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary()}
-
-		digests2signers, exists := m[currentVote]
-		if !exists {
-			digests2signers = make(map[string][]types.PartyID)
-			m[currentVote] = digests2signers
+		batch := batchKeyOf(baf)
```
tock-ibm, line 692: batch => bKey; maybe?

node/consensus/state/state.go
```diff
@@ -642,77 +678,66 @@ func getDigestSummary(digest2signers map[string][]types.PartyID) string {
...
 }
 
+// allBAFs includes every pending BAF (used when tallying signatures toward the threshold).
+func allBAFs(_ types.BatchAttestationFragment) bool { return true }
+
+// isNotSelfSigned excludes BAFs where Signer() == Primary(). Such self-signed BAFs cannot serve
+// as secondary attestations; excluding them prevents a non-primary from manufacturing
+// equivocation evidence by sending two BAFs with Primary() == Signer() but different digests.
+func isNotSelfSigned(baf types.BatchAttestationFragment) bool { return baf.Signer() != baf.Primary() }
+
```
tock-ibm, line 710: This is so simple that a function pointer is really not necessary, in my opinion.

node/consensus/state/state.go
```diff
@@ -565,20 +602,19 @@ func DetectEquivocation(s *State, _ types.ConfigSequence, l *flogging.FabricLogg
...
 	// empty PrimarySignature field, not by it re-broadcasting its own BAF as a voter).
-	m := batchAttestationVotesByDigestsExcludingSelfSigned(s)
+	m := batchAttestationVotesByDigests(s, isNotSelfSigned)
 
-	// Sort votes for deterministic processing: by shard, then primary, then seq.
-	votes := make([]batchAttestationVote, 0, len(m))
-	for vote := range m {
-		votes = append(votes, vote)
+	// Sort batches for deterministic processing: by shard, then primary, then seq.
+	batches := make([]batchKey, 0, len(m))
```
tock-ibm, line 613: batches => batchKeys or bKeys

node/consensus/state/state.go
```diff
@@ -642,77 +678,66 @@ func getDigestSummary(digest2signers map[string][]types.PartyID) string {
...
-		var foundThreshold bool
+	// thresholdDigests holds the specific <seq, shard, primary, digest> that individually
+	// reached threshold. Only fragments of these digests are extracted as attestations, so an
+	// under-attested (equivocating) digest can never back a committed block. If more than one
+	// digest for the same batch reaches threshold (byzantine primary), each is extracted and
+	// downstream becomes its own block.
+	thresholdDigests := make(map[thresholdDigestKey]struct{})
 
-		l.Debugf("A total of %d digests where found for seq %d in shard %d with primary %d", len(digest2signers), batchAttestation.seq, batchAttestation.shard, batchAttestation.primary)
+	for batch, digest2signers := range m {
```
tock-ibm, line 727: batch => bKeys or bafKey?

node/consensus/consenter.go
```diff
@@ -63,56 +63,48 @@ func (c *Consenter) Index(digests [][]byte) {
...
+// behavior); distinct digests (an equivocating primary) or distinct primaries (e.g. the same
+// batch reproduced across a rotation) are different BatchIDs and therefore different blocks.
+// This guarantees every group is a single batch, so consumers that read only the first
+// fragment (ba[0]) always see a threshold-attested batch rather than a stale/minority digest
+// that happened to be first in Pending order.
+type bafGroupKey struct {
+	seq     types.BatchSequence
+	shard   types.ShardID
+	primary types.PartyID
+	digest  string
```
tock-ibm, line 79: it would be readable to have the field order same as here: `ypes.BatchID (<shard, primary, seq, digest>): `

node/consensus/consenter.go
```diff
@@ -63,56 +63,48 @@ func (c *Consenter) Index(digests [][]byte) {
...
+}
 
-	var attestations [][]types.BatchAttestationFragment
+func bafKey(baf types.BatchAttestationFragment) bafGroupKey {
```
tock-ibm, line 82: It maybe more prudent to have the input as type.BatchID, which  types.BatchAttestationFragment actually implements.

node/consensus/consenter.go
tock-ibm, line 75: why not call it simply batchIDKey,  or keyBatchID, to reduce the mental load of what is what?

node/consensus/consenter.go
```diff
@@ -63,56 +63,48 @@ func (c *Consenter) Index(digests [][]byte) {
...
+	return bafGroupKey{seq: baf.Seq(), shard: baf.Shard(), primary: baf.Primary(), digest: string(baf.Digest())}
+}
 
-	added := make(map[struct {
-		seq   types.BatchSequence
-		shard types.ShardID
-	}]struct{})
+// aggregateFragments groups fragments by batch identity (bafGroupKey), preserving the order in
+// which each group's key first appears. Each returned group holds all fragments of one batch.
+func aggregateFragments(batchAttestationFragments []types.BatchAttestationFragment) [][]types.BatchAttestationFragment {
```
tock-ibm, line 88: This, like many of the other sorting and list manipulating utilities, should be tested on unit tests directly. Whitebox testing - in the `consensus` package. It would add coverage and confidence to this code.

tock-ibm (approved):
The main motivation of this PR is well covered and solved, and I see no bugs. Therefore, I approve. However I have several style comments that can be handled in a simple follow-up PR.

node/consensus/consenter.go
HagarMeir: This is extremely rare since it is only relevant if both reach threshold **in the same round**, otherwise it is gated by the BADB. 
Anyway if this happens then we will have a duplicate block but the assembler should be able to handle it gracefully.

node/consensus/state/state.go
HagarMeir: Yes a GC for the Pending BAFs is missing and there is an open issue

node/consensus/consenter.go
HagarMeir: will be fixed in the next PR

node/consensus/state/state.go
HagarMeir: Will add a comment
The rules will be refactored in a future PR

node/consensus/state/state.go
HagarMeir: there is a dedup that occurs within SubmitConfig done by SmartBFT's own mem pool
and config reqs are not accumulated in the state between rounds like BAFs and complaints 
so no need to dedup config reqs here

node/consensus/consenter.go
HagarMeir: don't really have a good minimal struct ready to use here

PR #1040, opened 2026-07-23
Config ack integration

MayRosenbaum:
issue #641


node/batcher/batcher.go
```diff
@@ -16,6 +16,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/hyperledger/fabric-x-orderer/common/configack"
+
```
HagarMeir: empty line

node/batcher/batcher.go
```diff
@@ -201,6 +204,8 @@ func (b *Batcher) Stop() {
 		b.opsSystem.Stop()
 	}
 
+	b.configAcker.Stop()
+
```
HagarMeir: why leave a line?

node/consensus/consensus.go
```diff
@@ -919,11 +924,10 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
 		}
 		inLatestDecision = true
 		c.Logger.Infof("Delivering config block number %d", lastBlockNum)
-		// if this is a new config block (with a larger number) then apply (soft stop)
+		// if this is a new config block (with a larger number) then apply
 		if c.lastConfigBlockNum < lastBlockNum {
 			c.decisionNumOfLastConfigBlock = hdr.Num
 			c.lastConfigBlockNum = lastBlockNum
-			c.Logger.Infof("Soft stop: pending restart")
```
HagarMeir, line 942: change comment but still leave one please

node/consensus/consensus.go
```diff
@@ -940,6 +944,22 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
 func (c *Consensus) processNewConfigBlock(configBlock *common.Block) {
 	c.Logger.Infof("Processing new config block number %d", configBlock.Header.Number)
 
+	// wait for acks
```
HagarMeir: why are we doing this before soft stop?

node/consensus/consensus_builder.go
```diff
@@ -13,6 +13,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/hyperledger/fabric-x-orderer/common/configack"
+
```
HagarMeir: empty line

node/assembler/assembler.go
```diff
@@ -379,6 +400,17 @@ func (a *Assembler) ProcessNewConfigBlock(configBlock *common.Block) {
 	if a.configuration == nil {
 		a.logger.Panicf("Current configuration is nil, cannot process new config block")
 	}
+
+	// send ack to the consensus node
+	configSeq, err := common_utils.GetConfigSequenceFromBlock(a.logger, configBlock, a.assemblerNodeConfig.BCCSP)
+	if err != nil {
+		a.logger.Warnf("failed to get config sequence from block %d", configSeq)
+		return
```
tock-ibm: 🔴 **The ack side-channel aborts config application.** If `GetConfigSequenceFromBlock` fails, this `return` skips `NewUpdatedConfigurationFromBlock` and the rest of `ProcessNewConfigBlock` entirely — so the assembler silently never applies the new config, only logging a `Warnf`. Sending an ack should be best-effort and must not gate applying the block.

Suggest: log the error and continue to apply the config (only skip the `SubmitConfigAck` call), rather than returning. Same pattern exists in `router.go` (~L693) and `batcher.go` (~L318) — the batcher case is worse because it has already `SoftStop()`ped (see that comment).

node/batcher/batcher.go
```diff
@@ -306,6 +311,17 @@ func (b *Batcher) replicateDecision() {
 func (b *Batcher) processNewConfigBlock(configBlock *common.Block) {
 	b.SoftStop()
```
tock-ibm, line 310: 🔴 **Early return after `SoftStop()` wedges the batcher.** `SoftStop()` runs here at the top, then a few lines down the `GetConfigSequenceFromBlock` error path `return`s before `ApplyConfig`. Result: on any failure to compute the ack sequence, the batcher is left soft-stopped and never applies the config or reconfigures. As with the assembler, the ack is a side-channel — on error, log and fall through to `ApplyConfig` instead of returning.

node/router/router.go
```diff
@@ -667,6 +687,17 @@ func (r *Router) processNewConfigBlock(configBlock *common.Block) {
 		return
 	}
 
+	// send ack to the consensus node
+	configSeq, err := utils.GetConfigSequenceFromBlock(r.logger, configBlock, r.routerNodeConfig.BCCSP)
+	if err != nil {
+		r.logger.Warnf("failed to get config sequence from block %d", configSeq)
```
tock-ibm: 🔴 Same as the assembler/batcher: this `return` on `GetConfigSequenceFromBlock` failure skips `ApplyConfig` below, so the router won't apply the new config. Log and continue; don't let ack-sequence extraction gate config application.

common/utils/block.go
```diff
@@ -0,0 +1,33 @@
...
+	"fmt"
+
+	"github.com/hyperledger/fabric-lib-go/bccsp"
+	"github.com/hyperledger/fabric-lib-go/common/flogging"
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"github.com/hyperledger/fabric-x-common/common/channelconfig"
+	"github.com/hyperledger/fabric-x-common/protoutil"
+)
+
+func GetConfigSequenceFromBlock(logger *flogging.FabricLogger, configBlock *common.Block, bccsp bccsp.BCCSP) (uint32, error) {
```
tock-ibm: 🟡 A few things in this helper:

1. **Double logging + wrapping**: each error is both `logger.Warn*`'d and returned wrapped. Pick one — callers already log. Prefer returning a wrapped error and letting the caller decide.
2. **Convention**: project uses `github.com/pkg/errors` (`errors.Wrapf`) rather than `fmt.Errorf` for wrapping (see CLAUDE.md / surrounding code).
3. **Inconsistent logger methods**: `Warningf` on L22 vs `Warnf` on L27 — pick one.
4. `uint32(bundle.ConfigtxValidator().Sequence())` truncates a `uint64`; fine given the proto field is `uint32`, just noting.

Also there's a stray blank line inside the second `if` block (before its closing brace).

node/consensus/consensus.go
```diff
@@ -940,6 +944,22 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
...
+	configSeq, err := utils.GetConfigSequenceFromBlock(c.Logger, configBlock, c.Config.BCCSP)
+	if err != nil {
+		c.Logger.Warnf("failed to get config sequence from block %d", configSeq)
```
tock-ibm: 🟡 The log prints `configSeq` (which is `0` on this error path) rather than the actual `err`. Should be e.g. `c.Logger.Warnf("failed to get config sequence from block: %v", err)`. This same `"...from block %d", configSeq` pattern is copy-pasted into the router/batcher/assembler error paths — all log the zero seq and drop `err`.

node/consensus/consensus.go
```diff
@@ -940,6 +944,22 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
...
+		return
+	}
+	c.Logger.Infof("waiting for acknowledgement from router, batchers and assembler on the new configuration on sequence %v\n", configSeq)
+	timeoutCtx, cancelFunc := context.WithTimeout(context.Background(), time.Second*60)
+	defer cancelFunc()
+	allAcksReceived := c.ConfigAckReceiver.WaitForAllAcks(timeoutCtx, uint64(configSeq))
+	if !allAcksReceived {
+		c.Logger.Warnf("consenter did not receive acknowledgments from all nodes on sequence %v\n", configSeq)
+	}
+	c.Logger.Infof("all acknowledgement have been received, is it safe to apply the new configuration")
```
tock-ibm: 🟠 Misleading log on timeout. When `WaitForAllAcks` returns false (timeout / not all acked), the code logs the warning on L958 but then unconditionally logs `"all acknowledgement have been received, is it safe to apply..."` here and proceeds. Move this success log into the `if allAcksReceived` branch so it isn't printed after a timeout. (Also minor wording: "is it safe" → "it is safe".)

node/consensus/consensus.go
```diff
@@ -940,6 +944,22 @@ func (c *Consensus) Deliver(proposal smartbft_types.Proposal, signatures []smart
...
+	// wait for acks
+	configSeq, err := utils.GetConfigSequenceFromBlock(c.Logger, configBlock, c.Config.BCCSP)
+	if err != nil {
+		c.Logger.Warnf("failed to get config sequence from block %d", configSeq)
+		return
+	}
+	c.Logger.Infof("waiting for acknowledgement from router, batchers and assembler on the new configuration on sequence %v\n", configSeq)
+	timeoutCtx, cancelFunc := context.WithTimeout(context.Background(), time.Second*60)
+	defer cancelFunc()
+	allAcksReceived := c.ConfigAckReceiver.WaitForAllAcks(timeoutCtx, uint64(configSeq))
```
tock-ibm: 🟡 The 60s ack-wait timeout is hardcoded here (and the sender's retry timeout is also hardcoded 60s in `sender.go`). Consider sourcing from config, or at least a named `const`, so the two sides can't drift. Non-blocking. Also worth confirming the intended semantics are documented: on timeout the consenter proceeds to apply config anyway (acks are advisory) — good for liveness, just make sure that's the design intent.

node/consensus/consensus_real_reconfig_test.go
```diff
@@ -601,3 +619,28 @@ func waitForPendingAdminState(t *testing.T, consenter *consensus_node.Consensus,
 		return status.State == node_utils.StatePendingAdmin && status.ConfigSequenceNumber == configSeq
 	}, 120*time.Second, 100*time.Millisecond)
 }
+
+func sendConfigAcks(consensusNodes []*consensus_node.Consensus, configSeq types.ConfigSequence) {
+	for _, node := range consensusNodes {
+		// one router ack
+		node.ConfigAckReceiver.AddAck(&protos.ConfigAck{
```
tock-ibm: 🟡 `AddAck` returns an error that's ignored in all three calls here. In a test helper that's arguably fine, but silently dropping it can mask a genuinely rejected ack (e.g. seq <= last) and make a future test failure confusing. Consider `require.NoError(t, node.ConfigAckReceiver.AddAck(...))` — would need threading `t` into the helper.

tock-ibm (commented):
## Review — Config ack integration (#641)

### Overview
Wires the `configack` Sender/Receiver (added in #985) into the live nodes. Router, batcher, and assembler now construct a `configack.Sender` at init and, in their `processNewConfigBlock`, send a `ConfigAck` to their party's consenter. The consenter constructs a `configack.Receiver`, and in `processNewConfigBlock` **blocks (up to 60s) on `WaitForAllAcks` before soft-stopping and applying** the new config. `AckConfig` RPC now validates the caller's TLS cert by node type (router/batcher/assembler) — the cert-matching logic was nicely generalized from the router-only `validateRouterFromContext` into a shared `validateClientFromContext`. `BCCSP` is threaded into the assembler/batcher node configs (router/consenter already had it), and a new `common/utils.GetConfigSequenceFromBlock` helper extracts the config sequence.

Overall the shape is good and the refactor of the cert validator is clean. My main concern is **error handling on the config-apply path**: the ack is a side-channel, but a failure to compute the ack sequence currently aborts the actual config application. Details inline.

### Summary of findings
- 🔴 On `GetConfigSequenceFromBlock` error, all three nodes `return` **before applying the config** — the batcher additionally has already `SoftStop()`ped, wedging it.
- 🟠 Consenter logs "all acknowledgement have been received" unconditionally, even when the wait timed out.
- 🟡 Error logs print the (zero) sequence instead of the actual `err`; `block.go` double-logs and doesn't use `pkg/errors`.
- 🟡 Test helper ignores `AddAck` errors.

### Notes on the `configack` package (from #985, not in this diff — no action required here)
- `receiver.AddAck` does a blocking send on `signalChan` (buffer 100) after releasing the lock; if acks arrive with no in-flight `WaitForAllAcks` they accumulate, and a burst >100 between waits would block the RPC handler. Fine at current N, worth a bounded/non-blocking send later.
- `WaitForAllAcks` proceeds (returns false) on timeout and the consenter applies config anyway — i.e. acks are best-effort/advisory. That seems intentional (avoids deadlock if a party member is down); flagging so it's a conscious choice.


node/assembler/assembler.go
MayRosenbaum: @tock-ibm wdyt ?

node/batcher/batcher.go
MayRosenbaum: @tock-ibm wdyt ?
its the same as above

node/router/router.go
MayRosenbaum: @tock-ibm wdyt ?
its the same as above

node/consensus/consensus.go
MayRosenbaum: I already left a comment above `SoftStop` in `processNewConfigBlock`

node/consensus/consensus.go
MayRosenbaum: no reason, fixed

node/assembler/assembler.go
tock-ibm: This is correct. This actually deserves a panic. The assembler receives a bad config block from a consenter  it trusts, there is not much else to do.

It would be better to augment the Configuration / NewConfigurationFromBlock with a way to get it.

Also, move the reporting after all processing of the new config and let the node a change to panic if the config is bad. IE right before  `if adminRequired {` .

common/utils/block.go
tock-ibm, line 1: This pretty redundant and out of place. 
I think adding to this:
```
func (config *Configuration) NewUpdatedConfigurationFromBlock(block *common.Block) (*Configuration, error)
```

a return value that is the sequence of the new config will be better, 

```
func (config *Configuration) NewUpdatedConfigurationFromBlock(block *common.Block) (*Configuration, uint64, error)
```

or even augmenting `Configuration` with a method that returns the config-sequence.

node/batcher/batcher.go
tock-ibm: This is correct. See comment on assembler, apply the same pattern.

node/router/router.go
```diff
@@ -667,6 +687,17 @@ func (r *Router) processNewConfigBlock(configBlock *common.Block) {
...
 
+	// send ack to the consensus node
+	configSeq, err := utils.GetConfigSequenceFromBlock(configBlock, r.routerNodeConfig.BCCSP)
+	if err != nil {
+		r.logger.Warnf("failed to get config sequence from block: %s", err)
+		return
+	}
+	if err := r.configAcker.SubmitConfigAck(configSeq); err != nil {
+		r.logger.Warnf("failed sending ConfigAck for config sequence %d: %v", configSeq, err)
+	}
```
tock-ibm: Same comment as on assembler and batcher. Failing to get the sequence number from a config is a cause for panic.

node/router/router.go
tock-ibm: This is correct. See comment on assembler.

common/configack/sender.go
```diff
@@ -18,6 +18,8 @@ import (
 	"google.golang.org/grpc"
 )
 
+const SubmitConfigAckTimeout = 60 * time.Second
```
tock-ibm: but a TODO to expose in local config

node/assembler/assembler.go
```diff
@@ -11,6 +11,10 @@ import (
 	"fmt"
 	"sync"
 
+	protos "github.com/hyperledger/fabric-x-orderer/node/protos/comm"
+
```
tock-ibm: remove empty lines

node/consensus/consensus.go
```diff
@@ -44,6 +47,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const ConfigAckReceiverTimeout = time.Second * 60
```
tock-ibm: put a todo to expose it in local config

common/utils/block.go
MayRosenbaum: Note that in the assembler, processNewConfigBlock behaves quite differently from the router, batcher, and consensus components, where NewUpdatedConfigurationFromBlock is called from ApplyConfig.

We have two options:

1. Send the acknowledgment from ApplyConfig in the router, batcher, and consensus.
2. Refactor the router, batcher, and consensus to follow the same ApplyConfig flow as the assembler.

To minimize the scope of changes, I'll go with option 1. However, for the sake of consistency and cleaner design, I think we should eventually refactor the code and adopt option 2.

config/config.go
```diff
@@ -924,26 +936,35 @@ func (config *Configuration) CheckIfAssemblerNodeExistsInSharedConfig() error {
...
+	env, err := protoutil.ExtractEnvelope(block, 0)
+	if err != nil {
+		return nil, 0, errors.Wrapf(err, "failed to extract envelope from new config block")
+	}
+	bundle, err := channelconfig.NewBundleFromEnvelope(env, bccsp)
+	if err != nil {
+		return nil, 0, errors.Wrapf(err, "failed to extract bundle from new config block")
+	}
+
+	return newConfig, uint32(bundle.ConfigtxValidator().Sequence()), nil
```
tock-ibm: why uint32? return the native type in the config

node/assembler/assembler.go
```diff
@@ -379,11 +398,17 @@ func (a *Assembler) ProcessNewConfigBlock(configBlock *common.Block) {
 	if a.configuration == nil {
 		a.logger.Panicf("Current configuration is nil, cannot process new config block")
 	}
-	newConfiguration, err := a.configuration.NewUpdatedConfigurationFromBlock(configBlock)
+
+	newConfiguration, configSeq, err := a.configuration.NewUpdatedConfigurationFromBlock(configBlock)
```
tock-ibm, line 402: return uint64, cast when using. 
I don't like that SubmitConfigAck casts to uint64... that is not the place to save bits. better to use the correct type - uint64.

node/consensus/consensus.go
```diff
@@ -982,11 +989,22 @@ func (c *Consensus) ApplyConfig(lastBlock *common.Block) (bool, error) {
...
+	newConfig, configSeq, err := currentFullConfig.NewUpdatedConfigurationFromBlock(lastBlock)
 	if err != nil {
 		return true, errors.Wrapf(err, "failed to build new configuration")
 	}
 
+	// wait for acks
+	c.Logger.Infof("waiting for acknowledgement from router, batchers and assembler on the new configuration on sequence %d", configSeq)
+	timeoutCtx, cancelFunc := context.WithTimeout(context.Background(), ConfigAckReceiverTimeout)
+	defer cancelFunc()
+	allAcksReceived := c.ConfigAckReceiver.WaitForAllAcks(timeoutCtx, uint64(configSeq))
```
tock-ibm: this is confusing... uint64(configSeq); there is absolutely no need to go back and forth from uint32 to uint64.

known_certs.txt
```diff
@@ -0,0 +1,568 @@
+│ Known Identities Implementation Guide                                                             │
```
tock-ibm: where does this come from? wrong PR :-)

tock-ibm: remove this from here form now

tock-ibm:
@MayRosenbaum 
Remove the known_cert.txt; resubmit, and I'll approve.
The uint32 / uint64 thing - do it in subsequent PR as I assume it needs a change in the proto.

known_certs.txt
MayRosenbaum: right!
My mistake by doing` git add . `
I removed

node/assembler/assembler.go
MayRosenbaum: I changed the proto, kept uint64

config/config_test.go
```diff
@@ -360,7 +360,7 @@ func TestConfigurationNewUpdatedConfigurationFromBlock(t *testing.T) {
 		}),
 	})
 
-	newConfig, err := fullConfig.NewUpdatedConfigurationFromBlock(genesisBlock)
+	newConfig, _, err := fullConfig.NewUpdatedConfigurationFromBlock(genesisBlock)
```
tock-ibm, line 363: check the sequence as well, in unit tests, next PR

common/utils/block.go
tock-ibm, line 1: This is unused, should be removed, next PR

tock-ibm (approved):
Approved with two minor comments. Fix both in the next PR please.

Issue #925, opened 2026-06-15
System testing: Nightly/Weekly runs in CI


tock-ibm:
## Nightly runs
Running small setup (4p2s) at low rate and for a long time. Pull from all assemblers.
Success criteria:
- [x] Basic: Verify that the total number of received transactions is at least the number of transactions that were sent.
- [ ] Advanced: Verify that every transaction that was sent was actually received. The loader maintains a map of transaction IDs (TX-IDs) for all submitted transactions. Each successfully received transaction removes its TX-ID from the map. At the end of the test, the map must be empty. It is enough to check one assembler for the TX arrival. Any remaining TX-IDs indicate transactions that were not received. At le end, also verify all assemblers have all the blocks, ie same block height.

NOTE: the Advanced part please do it last!

### Network startup
- [x] Use the /healthz endpoint to detect when the network is ready before starting the test, instead of relying on a fixed wait period.
### Failures scenarios
- [x] With failures: every several minutes, pick a component (R,B,C,A) from party X, shut it down, and after a minute, restart it. Cycle among components and parties. At the end verify the success criteria. Deterministic order.
- [x] With failures (fully randomized): Randomize both the party and component selection (e.g. p1-R, p2-A, p4-C, ...), ensuring all parties and components are exercised over the course of the run. At the end, verify the success criteria.
### CI integration
- [x] Integrate the test into GitHub Actions and run it on a configurable schedule (daily/weekly or other cron-based schedule).
### Reporting
- [x] Produce the test summary as plain text instead of packaging it into a ZIP archive.
- [x] Send an email/slack notification to the maintainers whenever a scheduled run fails.
- [x] Add a GitHub Actions workflow badge to the repository README to expose the current status of the nightly workflow.
### Make File
- [x] Add the ability to use "make" command for each kind of chaos test, one for the basic deteminstic option and one for the fully random option
### Refactoring
- [x] Squash all the bash files to a single one so it will be easier for all of the above to work smoothly
- [x] Rename the current merged chaos test to "deterministic failure test" instead, and when the real random test will be created call it same idea

PR #1042, opened 2026-07-23
Chaos test improve part 1

moran-abilea:
#925 

Extend Failure Tests in CI part 1:
* rename all "chaos test" to be "deterministic failure test" instead in all relevent files (before refactor them later as well with more changes)
* fixing summary.txt to be just text file and not zipped file
* adding badge to README.MD


.github/workflows/deterministic-failure-test.yml
```diff
@@ -116,42 +116,65 @@ jobs:
...
+                  { "title": "Repository", "value": "${{ github.repository }}", "short": true },
+                  { "title": "Branch",     "value": "${{ github.ref_name }}",   "short": true }
+                ],
+                "actions": [{
+                  "type": "button",
+                  "text": "View Run",
+                  "url": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+                }]
+              }]
+            }'
```
tock-ibm: Lets take this part out of the current PR and add it separately after all the other PRs.

.github/workflows/deterministic-failure-test.yml
```diff
@@ -116,42 +116,65 @@ jobs:
...
-          name: chaos-test-logs-${{ github.run_number }}
+          name: deterministic-failure-test-logs-${{ github.run_number }}
           path: test-results/logs/
           retention-days: 30
-          if-no-files-found: ignore
\ No newline at end of file
+          if-no-files-found: ignore
+
+      - name: Notify Slack on failure
+        if: failure() && (ithub.event_name == 'schedule'
```
HagarMeir: **ithub**

moran-abilea: fixed on the next PR, but will be removed for now as Yoav requested

tock-ibm:
## Code Review

### Overview
Rename/rebrand of the "chaos test" to "deterministic failure test" across the CI workflow, scripts, and docs. Three functional changes ride along: (1) workflow + env vars + marker/signal files renamed consistently, (2) test summary now published to the run's **Summary** tab via `$GITHUB_STEP_SUMMARY` instead of a zipped artifact, (3) a new Slack-on-failure step + README status badge.

The rename is clean and internally consistent — the marker/signal filenames line up across the three cooperating scripts (`*-test.sh` writes `failure_runner_enabled`, monitor reads it; runner writes `failure_runner_party${p}_done`, monitor reads/removes it; monitor writes `failure_runner_stop_signal`, runner reads it). Git `similarity index` confirms true renames, so history is preserved. 👍

---

### 🔴 Blocking: broken `if:` expression on the Slack step

`.github/workflows/deterministic-failure-test.yml`:

```yaml
- name: Notify Slack on failure
  if: failure() && (ithub.event_name == 'schedule'
```

Two defects on one line:
- **`ithub` → `github`** (missing `g`)
- **Unbalanced parenthesis** — the `(` is never closed

This is invalid Actions expression syntax; depending on the parser it can make the whole workflow fail to load. Either way the Slack-on-failure path won't work — and since it only fires on failure, a green CI run won't have exercised it. Presumably intended:

```yaml
if: failure() && github.event_name == 'schedule'
```

### 🟡 Minor: `curl` failure is silent
The Slack step runs `curl -s -X POST ...` with no `--fail` and no guard on `$SLACK_WEBHOOK_URL` being empty. If the secret is unset (e.g. on a fork), `curl` posts to an empty/garbage URL and the step still passes green, masking a broken alerting path. Consider `curl -sS --fail` and/or guarding on the secret being non-empty.

### 🟡 Minor: broadened job `if` may let forks run the heavy job
```yaml
if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.repository == 'hyperledger/fabric-x-orderer'
```
Adding `schedule` means a fork that leaves the schedule enabled would now run this 2–5.5h job; the original `repository ==` guard specifically prevented fork scheduled-runs. Probably harmless, but worth confirming the intent.

---

### Fine as-is
- `$GITHUB_STEP_SUMMARY` change is the correct idiom and handles the missing-file case with an explicit fallback. `collect-results.sh` consistently drops the `summary/` dir and writes `test-results/summary.txt` everywhere.
- README badge points at the renamed workflow. ✅
- SPDX/license headers preserved on renamed scripts; env-var defaulting preserved verbatim.

### Nits
- Nice catch fixing the pre-existing `` ```bashsame idea `` fence → `` ```bash ``.
- PR description has a couple typos ("fialure", "akk relevent files") — cosmetic.

---

**Verdict:** Solid, low-risk rename with good internal consistency. One blocking item — the malformed `if: failure() && (ithub.event_name == 'schedule'` (typo + unbalanced paren) — should be fixed before merge. The silent-`curl` and fork-schedule points are optional hardening.

tock-ibm (approved):

PR #1044, opened 2026-07-26
Add sample local config validation test

moradna:
- Added `sample_validation_test.go` to validate sample local config files.
- Updated sample YAML files to match the generated configuration.
- Removed unused default constants from `local_config_gen.go`.

#935

### DIFF
--- a/config/sample/test-sample/config/party1/local_config_batcher1.yaml
+++ b/config/sample/test-sample/config/party1/local_config_batcher1.yaml
@@ -35,7 +35,7 @@ General:
         Default: SW
         SW:
             Security: 256
-            Hash: SHA2      
+            Hash: SHA2
     LogSpec: info
 FileStore:
     Location: /var/dec-trust/production/orderer/store
--- a/config/sample/test-sample/config/party1/local_config_batcher1.yaml
+++ b/config/sample/test-sample/config/party1/local_config_batcher1.yaml
@@ -34,13 +34,9 @@ General:
     BCCSP:
         Default: SW
         SW:
-            Hash: SHA2
             Security: 256
-            FileKeyStore:
-                KeyStore:
+            Hash: SHA2      
     LogSpec: info
-    ClientSignatureVerificationRequired: false
-    MetricsLogInterval: 0s
 FileStore:
     Location: /var/dec-trust/production/orderer/store
 Batcher:
@@ -48,3 +44,11 @@ Batcher:
     BatchSequenceGap: 10
     MemPoolMaxSize: 1000000
     SubmitTimeout: 500ms
+Operations:
+    ListenAddress: 127.0.0.1
+    TLS:
+        Enabled: false
+        ClientAuthRequired: false
+Metrics:
+    Provider: prometheus
+    MetricsLogInterval: 10s
--- a/config/sample/test-sample/config/party4/local_config_assembler.yaml
+++ b/config/sample/test-sample/config/party4/local_config_assembler.yaml
@@ -34,10 +34,8 @@ General:
     BCCSP:
         Default: SW
         SW:
+            Security: 256   
             Hash: SHA2
-            Security: 256
-            FileKeyStore:
-                KeyStore:
     LogSpec: info
 FileStore:
     Location: /var/dec-trust/production/orderer/store
@@ -48,3 +46,11 @@ Assembler:
     PopWaitMonitorTimeout: 1s
     ReplicationChannelSize: 100
     BatchRequestsChannelSize: 1000
+Operations:
+    ListenAddress: 127.0.0.1
+    TLS:
+        Enabled: false
+        ClientAuthRequired: false
+Metrics:
+    Provider: prometheus
+    MetricsLogInterval: 10s
--- a/config/sample/test-sample/config/party4/local_config_assembler.yaml
+++ b/config/sample/test-sample/config/party4/local_config_assembler.yaml
@@ -34,7 +34,7 @@ General:
     BCCSP:
         Default: SW
         SW:
-            Security: 256   
+            Security: 256
             Hash: SHA2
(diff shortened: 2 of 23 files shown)


config/sample/test-sample/config/party1/local_config_batcher1.yaml
```diff
@@ -34,17 +34,21 @@ General:
     BCCSP:
         Default: SW
         SW:
-            Hash: SHA2
             Security: 256
-            FileKeyStore:
-                KeyStore:
+            Hash: SHA2
```
tock-ibm: Trailing whitespace: `Hash: SHA2` has 6 trailing spaces here. Cosmetic to the parser, but worth cleaning up in a config-hygiene PR — and it may trip the whitespace/format check in `make basic-checks`.

config/sample/test-sample/config/party4/local_config_batcher1.yaml
```diff
@@ -34,17 +34,21 @@ General:
     BCCSP:
         Default: SW
         SW:
-            Hash: SHA2
             Security: 256
-            FileKeyStore:
-                KeyStore:
+            Hash: SHA2
```
tock-ibm: Trailing whitespace on `Hash: SHA2 `.

config/sample/test-sample/config/party4/local_config_assembler.yaml
```diff
@@ -34,10 +34,8 @@ General:
     BCCSP:
         Default: SW
         SW:
+            Security: 256
```
tock-ibm: Two nits on this line: (1) trailing whitespace after `Security: 256`; (2) `Security` is placed before `Hash` here, whereas every other sample file lists `Hash` then `Security`. Gratuitous inconsistency in a PR whose goal is sample consistency — please align the ordering.

config/sample_validation_test.go
```diff
@@ -0,0 +1,93 @@
...
+		if d.IsDir() {
+			return nil
+		}
+
+		matched, err := filepath.Match("local_config*.yaml", d.Name())
+		require.NoError(t, err)
+
+		if !matched {
+			return nil
+		}
```
tock-ibm, line 37: The `require.NoError(t, err)` calls here run in the `WalkDir` callback rather than inside a `t.Run` subtest. It's technically correct (the walk is synchronous on the test goroutine, so `t.FailNow()`'s `runtime.Goexit()` is safe), but mixing top-level `require` with the per-file `t.Run` is a little awkward. Consider `if err != nil { return err }` in the callback and asserting once on the final `filepath.WalkDir` return — which you already do at the end.

config/sample_validation_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		t.Run(path, func(t *testing.T) {
+			data, err := os.ReadFile(path)
+			require.NoError(t, err)
```
tock-ibm, line 41: This pattern also matches the hand-written template `config/sample/local_config.yaml` (full comments + `PKCS11:` block), not just the generated `test-sample` files — good extra coverage. Scope note re #935: the test only checks that samples *decode*, not that they match what `armageddon generate` actually emits, and it doesn't run `applyNodeDefaults`/operations validation (so e.g. `ClientAuthRequired: true` with no `ClientRootCAs` would still pass). A stronger follow-up could generate into a temp dir and decode+validate those. Fine as-is for this PR.

tock-ibm (commented):
## Review: Add sample local config validation test

Good, low-risk hygiene change. The sample files had genuinely drifted from the current schema (`LedgerScanTimeout`/`EvictionTtlL` → `RestartLedgerScanTimeout`/`PrefetchEvictionTtl`, missing `Operations`/`Metrics` sections), and this test locks them to the real `NodeLocalConfig` struct going forward. I checked out the branch: **all 26 subtests pass**, and `go build ./...` / `go vet ./config/...` are clean.

### Strengths
- **Catches real drift.** Strict `KnownFields(true)` decode into the actual struct means any future rename or typo fails CI — exactly the class of bug the old `EvictionTtlL` typo represents.
- **Correct dead-code removal.** The dropped `metricsLogInterval` param of `NewGeneralConfigParams` was accepted but never stored in `GeneralConfigParams`; the four `DefaultXMonitoringPort` / `DefaultClientSignatureVerificationRequired` constants had no other references (grepped). `DefaultMetricsProviderType` / `DefaultMetricsLogInterval` are correctly retained.
- **PKCS11 stripping is well-justified.** The `PKCS11` field on `factory.FactoryOpts` only exists under the `//go:build pkcs11` tag, so in the default build `KnownFields(true)` would reject it. Stripping before decode is the right call.

### Suggestions (see inline comments)
1. Trailing whitespace introduced in a few sample YAMLs, plus an inconsistent `Security`/`Hash` field reorder — cosmetic, but this is a consistency PR. Likely to trip `make basic-checks`.
2. `require.*` inside the `WalkDir` callback is technically fine (synchronous walk) but awkward next to per-file `t.Run`.
3. Scope note on what the test validates vs. what "samples match generated config" (#935) would need.

### Correctness / Convention / Security
- Conventions: `testify/require` matches neighboring config tests; license header present. ✅
- Security: test-only + sample data; TLS `Enabled: false` is appropriate for local-dev samples. ✅
- Performance: test-only, ~0.06s. ✅

### Verdict
Solid, well-scoped improvement. Recommend fixing the whitespace/ordering nits before merge; the rest are optional follow-ups.

config/sample/test-sample/config/party4/local_config_assembler.yaml
moradna: I checked the generated sample files, and they all consistently use `Security` before `Hash`. I only removed the trailing whitespace.

tock-ibm (approved):

PR #1046, opened 2026-07-26
Chaos test improve part 2

moran-abilea:
#925 

Extend Failure Tests in CI part 2:

based on the previous PR of part 1 (https://github.com/hyperledger/fabric-x-orderer/pull/1042) it was extended to include:
- create "make deterministic-failure-test" command that will simpley run 5 minutes short run.
(we can still change the paramaters if needed for example: "make deterministic-failure-test DURATION_MINUTES=10 TX_RATE=500"
- refactor the deterministic failure test bash scripts to squashed into a single file only.
- update the README.md with those changes as well.

only on the next PR we will improve this test even further (using healthz and checking that every tx was really recieved) before moving on to the final part of creating a real chaos test


.github/workflows/deterministic-failure-test.yml
```diff
@@ -116,42 +116,65 @@ jobs:
...
-          name: chaos-test-logs-${{ github.run_number }}
+          name: deterministic-failure-test-logs-${{ github.run_number }}
           path: test-results/logs/
           retention-days: 30
-          if-no-files-found: ignore
\ No newline at end of file
+          if-no-files-found: ignore
+
+      - name: Notify Slack on failure
+        if: failure() && (github.event_name == 'schedule'
```
tock-ibm: 🔴 **Blocking: malformed `if:` expression.** The `ithub`→`github` typo from #1042 is fixed here, but the `(` is opened and never closed. This is invalid GitHub Actions expression syntax and will break the step (or the whole workflow parse). Should be:

```yaml
if: failure() && github.event_name == 'schedule'
```

Note this path only runs on failure, so a green CI run won't have exercised it.

.github/workflows/deterministic-failure-test.yml
```diff
@@ -116,42 +116,65 @@ jobs:
...
+        run: |
+          curl -s -X POST "${{ secrets.SLACK_WEBHOOK_URL }}" \
```
tock-ibm: 🟡 Minor: `curl -s` has no `--fail` and no guard on `SLACK_WEBHOOK_URL` being empty. If the secret is unset (e.g. on a fork), curl posts to an empty/garbage URL and the step still passes green, masking a broken alerting path. Consider `curl -sS --fail` and/or guarding on the secret being non-empty.

.github/workflows/deterministic-failure-test.yml
```diff
@@ -75,11 +75,11 @@ permissions:
...
 
 jobs:
-  chaos-test:
-    name: Chaos Test
+  deterministic-failure-test:
+    name: Deterministic Failure Test
     runs-on: ubuntu-latest
     timeout-minutes: 350
-    if: github.repository == 'hyperledger/fabric-x-orderer' || github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule' || github.repository == 'hyperledger/fabric-x-orderer'
```
tock-ibm, line 82: 🟡 Minor (carried over from #1042): adding `github.event_name == 'schedule'` to the job guard means a fork that leaves the schedule enabled would now run this heavy 2–5.5h job. The original `repository ==` guard specifically prevented fork scheduled-runs. Probably harmless, but worth confirming the intent.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -0,0 +1,891 @@
...
+
+    local PID=$(cat ${pid_file} 2>/dev/null)
+
+    # Check if process is still running
+    if [ -z "$PID" ] || ! kill -0 $PID 2>/dev/null; then
+      echo "[$(date '+%Y-%m-%d %H:%M:%S')] WARNING: ${component} (party ${party}${shard:+ shard ${shard}}) not running (PID: ${PID:-unknown})"
+      # Try to restart anyway
+    else
+      # Kill the process
+      kill $PID 2>/dev/null
```
tock-ibm: 🟠 **Medium: this `kill` can silently kill the whole failure runner under `set -e`.** Previously this loop lived in `chaos-runner.sh`, a separate process with no `set -e`. Now `run_failure_runner` is a backgrounded subshell that inherits the script-level `set -e` (line 20), and this is a bare statement.

Race: the `kill -0 $PID` guard just above can succeed, the process then exits, and this `kill $PID` returns non-zero → `set -e` terminates the failure-runner subshell silently. Over a multi-hour scheduled run with hundreds of kills this will eventually fire, and the test keeps "passing" with failure injection quietly stopped — defeating the purpose.

Fix options:
- `kill $PID 2>/dev/null || true`, or
- wrap the runner body in `set +e` / `set -e`, or
- a `trap`-based cleanup instead of a global `set -e` in the injection loop.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -0,0 +1,891 @@
...
+#   deterministic-failure-runner.sh (failure injection loop)
+#   monitor-completion.sh           (progress monitor)
+#   collect-results.sh              (result collection)
+#
+# All functions are now defined here and called from the main() function at the
+# bottom of this file.  The GitHub Actions workflow and local usage are
+# unchanged — just run this script directly.
+
+# Exit on error
+set -e
```
tock-ibm, line 20: Context for the `kill` comment below: after consolidation everything runs under this single `set -e`. Note `local X=$(...)` masks command-substitution failures (bash quirk — `local` always returns 0), while bare statements like the `kill` do not — so `set -e` behavior is now inconsistent across the merged file. Worth a deliberate pass.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -0,0 +1,891 @@
...
+    fi
+  fi
+
+  # Collect results
+  echo "Collecting results..."
+  collect_results "${TEST_DIR}" "${NUM_PARTIES}" "${DURATION}"
+
+  # Cleanup processes
+  echo "Cleaning up processes..."
+  pkill -f "arma " || true
```
tock-ibm, line 887: 🟡 Minor: under `set -e`, any error before these final `pkill` lines leaves orphaned `arma`/`armageddon` processes. Now that everything is one process this bites harder. Consider `trap 'pkill -f "arma "; pkill -f armageddon' EXIT` near the top for robust cleanup.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -0,0 +1,891 @@
...
+      fi
+    else
+      echo "  ❌ No statistics available" >> test-results/summary.txt
+    fi
+  done
+
+  # Overall statistics: each party independently receives all sent txs, so the
+  # meaningful metric is per-party success rate, not a cross-party sum.
+  local PARTY_SUCCESS_RATE=0
+  REPRESENTATIVE_RECEIVED=${RECEIVER_STATS[1]:-0}
```
tock-ibm, line 673: 👍 Nice: `REPRESENTATIVE_RECEIVED` is now initialized unconditionally here before use. The original only set it inside the `if [ "$SENT" -gt 0 ]` block, so the `Received per party:` summary line could reference an unset var. Good catch during the refactor.

tock-ibm (commented):
Re-posting my earlier review as inline comments so each point is anchored and resolvable.

Overview: Part 2 on top of #1042's rename — adds a `make deterministic-failure-test` target and consolidates the five `test/chaos/*.sh` scripts into one `deterministic-failure-test.sh` (each former script becomes a function). The extraction is a faithful transcription of the deleted originals.

**Blocking:** the Slack `if:` still has an unbalanced paren.
**Substantive:** the failure runner now inherits `set -e` after the merge and a racy `kill` can silently stop failure injection.

Process note: this PR and #1042 both target `main` and conflict (rename vs. delete+recreate) — they can't both merge cleanly, so a merge-order decision is needed.

test/deterministic-failure-test/README.md
```diff
@@ -0,0 +1,210 @@
...
+└── pids/            # PID files for all started processes
+```
+
+Result artifacts are written to (cleaned at the start of each run):
+
+```text
+test-results/
+├── logs/            # component and loader/receiver logs
+├── statistics/      # per-party statistics CSV files
+└── summary/         # summary.txt with pass/fail and tx counts
```
tock-ibm: 🟡 Stale doc: this tree still lists `test-results/summary/` as a subdirectory, but the code now writes `test-results/summary.txt` at the top level (the `mkdir -p test-results/summary` was removed in `collect_results`). Please update the tree to reflect the flat file.

.github/workflows/deterministic-failure-test.yml
moran-abilea: this and any other comments that are related to slack will be moved to the last PR then

tock-ibm (approved):

PR #1048, opened 2026-07-27
Chaos test improve part3

moran-abilea:
https://github.com/hyperledger/fabric-x-orderer/issues/925

Extend Failure Tests in CI part 3:

this will be the major improvment by using "/healthz" endpoint of each component and also send send relevant error messages through the dedicated slack channel

based on the previous PR of part 2 (https://github.com/hyperledger/fabric-x-orderer/pull/1046) it was extended to include:

- checking "/healthz" of each component instead of waiting configuratble amount of time (including removing the previous approach from all the code)
- send messaging slack correctly if the network could not be initiated completly or somehow the github job not having enough resource/busy. in any other case just send the link to the job itself for more details.


test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -714,10 +773,6 @@ EOF
 # main — orchestrates the full test run
 # ---------------------------------------------------------------------------
 main() {
-  # Ensure arma/armageddon processes are cleaned up even if the script exits early
-  # (e.g. due to set -e tripping on an unexpected error before the final pkill lines).
-  trap 'pkill -f "arma " 2>/dev/null || true; pkill -f armageddon 2>/dev/null || true' EXIT
-
   # Calculate total transactions
```
tock-ibm, line 780: **Regression:** this PR removes the `trap '... pkill ...' EXIT` that used to be the first line of `main()`. Its own comment said it existed to catch "`set -e` tripping on an unexpected error before the final `pkill` lines" — and this same PR *adds* new early-exit paths via `wait_for_healthz`'s `exit 1`. So a health-check timeout now leaves every already-started component (all consenters + some batchers, etc.) running as orphaned background processes with no cleanup. On CI the runner is ephemeral so it's bounded, but for the documented local usage this leaks processes and can wedge the next run via LevelDB lock conflicts. Please restore the trap.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -199,7 +258,7 @@ run_failure_runner() {
       # Try to restart anyway
     else
       # Kill the process
-      kill $PID 2>/dev/null || true
+      kill $PID 2>/dev/null
```
tock-ibm: Dropping `|| true` here is risky under `set -e`. `run_failure_runner` runs as a backgrounded function, and there's a TOCTOU window: the PID is checked live with `kill -0` above, but the process can exit before this `kill` fires (the runner deliberately kills components that may already be dying). A failed `kill` returns non-zero and, under `set -e`, terminates the failure-runner subshell — silently ending fault injection mid-test. The `|| true` bought real safety; I'd keep it. Same applies to the `kill -9` below.

test/deterministic-failure-test/deterministic-failure-test.sh
```diff
@@ -212,7 +271,7 @@ run_failure_runner() {
       # Force kill if still running
       if kill -0 $PID 2>/dev/null; then
         echo "[$(date '+%Y-%m-%d %H:%M:%S')] Force killing ${component} (party ${party}${shard:+ shard ${shard}})"
-        kill -9 $PID 2>/dev/null || true
+        kill -9 $PID 2>/dev/null
```
tock-ibm: Same as the `kill $PID` above — restore `|| true`. The force-kill is the *more* likely one to race (process already gone after the 10s wait loop), so a non-zero exit here under `set -e` will kill the runner subshell.

test/deterministic-failure-test/README.md
```diff
@@ -174,7 +174,7 @@ Result artifacts are written to (cleaned at the start of each run):
 test-results/
 ├── logs/            # component and loader/receiver logs
 ├── statistics/      # per-party statistics CSV files
-└── summary.txt      # pass/fail verdict and tx counts
+└── summary/         # summary.txt with pass/fail and tx counts
```
tock-ibm: Doc/code mismatch: the script still writes a flat file (`cat > test-results/summary.txt`, ~10 sites) and never creates a `summary/` directory. Either revert this line to `summary.txt` or actually nest the file under `summary/` in the script.

.github/workflows/deterministic-failure-test.yml
```diff
@@ -157,3 +151,44 @@ jobs:
...
+          fi
+
+      - name: Notify Slack on failure
+        if: failure() && github.event_name == 'schedule'
+        run: |
+          if [ -z "${{ secrets.SLACK_WEBHOOK_URL }}" ]; then
+            echo "SLACK_WEBHOOK_URL is not set — skipping notification"
+            exit 0
+          fi
+          REASON="${{ steps.failure_reason.outputs.reason }}"
```
tock-ibm: Interpolating `${{ steps.failure_reason.outputs.reason }}` directly into a `run:` shell body is the GitHub Actions script-injection antipattern. Today the reason is effectively trusted (hardcoded labels + `127.0.0.1` URLs), so real risk is low — but prefer passing it via `env:` (`env:\n  REASON: ${{ steps.failure_reason.outputs.reason }}`) so a future change to what lands in `startup_failure.txt` can't break out. Related: `$REASON`/`$TEXT` are spliced into the JSON `--data` unescaped, so a `"`, `\`, or newline would produce invalid JSON. Fine for today's single-line reasons, but brittle.

tock-ibm (commented):
## Review — Chaos test improve part3

Solid direction: replacing blind `sleep`s with real `/healthz` polling is a clear improvement, and the `if:` condition fix (`(dispatch || schedule) && repo == …`) corrects a previously always-true guard. A few issues below — two I'd treat as blocking regressions to test robustness, plus a doc mismatch and some workflow hardening.

**Blocking**
- **Removed `EXIT` cleanup trap** — the new `wait_for_healthz` adds `exit 1` paths that now leak every already-started component (orphaned processes, LevelDB lock conflicts on the next local run). The trap that was deleted existed precisely for this case.
- **`kill … || true` removed under `set -e`** — a TOCTOU race (process dies between `kill -0` and `kill`) can now terminate the backgrounded failure-runner subshell early, silently ending fault injection mid-test.

**Non-blocking**
- README `summary/` claim doesn't match the script (still writes flat `summary.txt`).
- Slack step interpolates `${{ }}` into the shell body and splices unescaped text into JSON — pass via `env:` and it's both safer and more robust.
- Dead `else` branch in the notify step (`reason` is always set by the detect step).
- `/healthz` is liveness-only (all checkers `return nil`) — it confirms the ops server bound, not that the network can order TXs. The removed `sleep 10 # stabilize` gave the mesh time to connect; worth validating a few CI runs that the loader isn't now firing before the network is functionally ready.

Details inline.

tock-ibm:
🤖 The review above was generated with [Claude Code](https://claude.com/claude-code) at the request of @Yoav Tock.

test/deterministic-failure-test/deterministic-failure-test.sh
moran-abilea: fixed as suggested

test/deterministic-failure-test/deterministic-failure-test.sh
moran-abilea: fixed as suggested

test/deterministic-failure-test/deterministic-failure-test.sh
moran-abilea: fixed as suggested

test/deterministic-failure-test/README.md
moran-abilea: fixed as suggested.
the folder itself of "summary/" is removed since last part, but fixing this part in the readme.md was missed

.github/workflows/deterministic-failure-test.yml
moran-abilea: fixed as suggested

tock-ibm (approved):

PR #1051, opened 2026-07-27
move config req out of state

HagarMeir:


tock-ibm:
## Code Review

### Overview
Pure refactor extracting the `ConfigRequest` type and its methods (`ConfigSequence`, `toProto`, `fromProto`, `Bytes`, `FromBytes`, `String`) out of `state.go` into a dedicated `config_request.go`, plus a new `config_request_test.go`. No behavior change — the code is moved verbatim, following the established pattern from `move complaint out of state (#1026)`.

### Verification (on the PR branch)
- ✅ `go build ./node/consensus/...` — passes; no dangling imports left in `state.go` (removed `common`, `configtx`, `protoutil`, `pkg/errors`; remaining `fmt`/`proto` still used).
- ✅ `go vet ./node/consensus/state/` — clean.
- ✅ `go test -run TestConfigRequest ./node/consensus/state/` — passes.
- ✅ `gofmt -l` — clean.
- ✅ Type stays in `package state`, so external references (`consensus.go`, `consensus_config_applier.go`, mocks) and the unexported `toProto`/`fromProto` cross-file calls still resolve.

### Strengths
- **Net improvement in test coverage.** The moved code previously had no dedicated tests; this PR adds serialization round-trip, invalid-bytes, sequence extraction, error paths, and `String()` cases. Test comments clearly explain intent.
- **Consistent with conventions** — matches sibling files, keeps a struct's methods collated, uses `package state_test` for black-box testing.

### Minor observations (non-blocking nits)
- **Mixed error idiom.** The moved code mixes `errors.Wrap`/`errors.New` (`pkg/errors`) in `ConfigSequence` with `fmt.Errorf` in `fromProto`/`FromBytes`. Pre-existing (moved verbatim), so out of scope here — but if touching it, the `fmt.Errorf(... %v)` calls could become `errors.Wrapf`.
- **`assert` vs `require`.** In the round-trip tests, `require.NoError` before dereferencing `cr2.Envelope` would be marginally safer than `assert.NoError` (prevents a nil-panic if an earlier step ever fails). Currently safe since the inputs are valid.

### Risks
None material. Zero logic change, verified compiling and passing.

**Verdict:** LGTM — clean, well-scoped refactor that adds coverage. The two nits are optional and mostly pre-existing.

tock-ibm (approved):

PR #1055, opened 2026-07-28
guard consensus RPC handlers against soft-stop without racing softStopCh

HagarMeir:
Add soft-stop guards to OnConsensus and OnSubmit, and read softStopCh through a new isSoftStopped helper under a read lock (lock changed to RWMutex) so handler reads no longer race the softStopCh reassignment in Start

https://github.com/hyperledger/fabric-x-orderer/issues/830

### DIFF
--- a/node/consensus/consensus.go
+++ b/node/consensus/consensus.go
@@ -98,7 +98,7 @@ type Consensus struct {
 	Arma         Arma
 	PartyID      arma_types.PartyID
 
-	lock                         sync.Mutex
+	lock                         sync.RWMutex
 	State                        *state.State
 	status                       node_utils.NodeStatus
 	fullConfig                   *config.Configuration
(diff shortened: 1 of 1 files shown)


node/consensus/consensus.go
```diff
@@ -244,7 +244,26 @@ func (c *Consensus) SoftStop() {
...
+	case <-ch:
+		return true
+	default:
+		return false
+	}
+}
+
 func (c *Consensus) OnConsensus(channel string, sender uint64, request *orderer.ConsensusRequest) error {
+	if c.isSoftStopped() {
+		return errors.New("consensus is soft-stopped")
```
tock-ibm, line 265: Optional (non-blocking, pre-existing): `errors.New("consensus is soft-stopped")` is now duplicated across five methods (this new guard plus `OnSubmit`, `NotifyEvent`, `SubmitConfig`, `SubmitRequest`). Consider a package-level `var errSoftStopped = errors.New("consensus is soft-stopped")` and returning that everywhere — cheaper and keeps the message in one place.

tock-ibm (commented):
## Review — race fix looks correct ✅

**What this does:** fixes a data race on `softStopCh` (reassigned in `Start` under the write lock, previously read lock-free in the handlers) by reading it through the new `isSoftStopped()` helper under an `RLock`, and extends the soft-stop guard to `OnConsensus`/`OnSubmit`.

**Correctness**
- `Start()` reassigns `c.softStopCh` under the write lock while `NotifyEvent`/`SubmitConfig`/`SubmitRequest` read the field with no lock — a genuine read/write race on the field. Copying the pointer under `RLock` and running the non-blocking `select` after releasing closes it. Writes/closes (`Start`, `Stop`, `SoftStop`) are all under the write lock, so field access is now fully synchronized.
- `select` semantics hold: closed channel → returns `true`; `nil` channel (pre-`Start` zero value) → both cases block, `default` wins → `false` (safe).
- Still a point-in-time (TOCTOU) guard — `SoftStop` can close the channel right after the check returns `false`. That's inherent and unchanged; real shutdown safety comes from `BFT.Stop()`. Fine for a best-effort guard.

**Style**
- `RWMutex` is a clean drop-in for all existing `Lock/Unlock` sites. Helper is small and its doc comment correctly states the invariant. Converting the three existing raw `select` checks to the helper is a nice DRY cleanup.

**Testing**
- No test added. A `-race` regression test running `Start` concurrent with a handler would lock this in. Make sure the exercising test isn't only in `TestConsensusFullReplacement` (excluded from the default `unit-tests-consensus` target).

**Risk:** Low — scoped to one file. The one new outward behavior is `OnConsensus`/`OnSubmit` now returning an error during soft-stop; since the other three handlers already returned the same error, callers almost certainly handle it consistently.

tock-ibm (approved):

PR #1063, opened 2026-07-29
cert id msp

MayRosenbaum:
issue #638

### DIFF
--- a/test/basic/sigverify_test.go
+++ b/test/basic/sigverify_test.go
@@ -230,10 +230,7 @@ func TestSubmitTXWithKnownCertID(t *testing.T) {
 	configPath := filepath.Join(dir, "config.yaml")
 	numOfParties := 4
 	numOfShards := 2
-	parties := []types.PartyID{1, 2, 3, 4}
-	for partyID := 1; partyID <= numOfParties; partyID++ {
-		parties = append(parties, types.PartyID(partyID))
-	}
+
 	netInfo := testutil.CreateNetwork(t, configPath, numOfParties, numOfShards, "mTLS", "mTLS")
 	defer netInfo.CleanUp()
 	require.NotNil(t, netInfo)
@@ -262,7 +259,7 @@ func TestSubmitTXWithKnownCertID(t *testing.T) {
 	defer broadcastClient.Stop()
 
 	t.Log("Submit TX with known certID – expect success")
-	mspID := fmt.Sprintf("org%d", 1)
+	mspID := "org1"
 	txContent := tx.PrepareTxWithTimestamp(1, 64, []byte("knownCertIDTx"))
 	envWithCertID := tx.CreateSignedStructuredEnvelopeWithCertID(txContent, signer, certBytes, mspID)
 	require.NotNil(t, envWithCertID)
@@ -275,8 +272,9 @@ func TestSubmitTXWithKnownCertID(t *testing.T) {
 	unknownClientPair, err := unknownCA.NewClientCertKeyPair()
 	require.NoError(t, err)
 	block, _ := pem.Decode(unknownClientPair.Key)
-	reLabeled := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: block.Bytes})
-	unknownPrivKey, err := tx.CreateECDSAPrivateKey(reLabeled)
+	require.NotNil(t, block)
+	privKeyBytes := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: block.Bytes})
+	unknownPrivKey, err := tx.CreateECDSAPrivateKey(privKeyBytes)
 	require.NoError(t, err)
 	unknownSigner := (*crypto.ECDSASigner)(unknownPrivKey)
 
--- a/test/basic/sigverify_test.go
+++ b/test/basic/sigverify_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package basic
 
 import (
+	"encoding/pem"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -16,6 +17,8 @@ import (
 	"github.com/hyperledger/fabric-protos-go-apiv2/common"
 	"github.com/hyperledger/fabric-x-orderer/common/tools/armageddon"
 	"github.com/hyperledger/fabric-x-orderer/common/types"
+	"github.com/hyperledger/fabric-x-orderer/node/comm/tlsgen"
+	"github.com/hyperledger/fabric-x-orderer/node/crypto"
 	test_utils "github.com/hyperledger/fabric-x-orderer/test/utils"
 	"github.com/hyperledger/fabric-x-orderer/testutil"
 	"github.com/hyperledger/fabric-x-orderer/testutil/client"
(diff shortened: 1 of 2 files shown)


test/basic/sigverify_test.go
```diff
@@ -203,3 +206,96 @@ func TestSubmitTXWithVerification(t *testing.T) {
...
+	require.NotNil(t, armaBinaryPath)
+
+	t.Log("Generate the configuration with clientSignatureVerificationRequired = True")
+	dir := t.TempDir()
+	configPath := filepath.Join(dir, "config.yaml")
+	numOfParties := 4
+	numOfShards := 2
+	parties := []types.PartyID{1, 2, 3, 4}
+	for partyID := 1; partyID <= numOfParties; partyID++ {
+		parties = append(parties, types.PartyID(partyID))
```
tock-ibm: **Bug + dead code.** `parties` is pre-seeded with `{1,2,3,4}` and then appended to in the loop, producing `[1,2,3,4,1,2,3,4]`. Worse, the variable is never used afterward — `PullFromAssemblers` below uses `Parties: []types.PartyID{1}`. The sibling `TestSubmitTXWithVerification` starts from `[]types.PartyID{}`. Delete this block entirely (or fix the seeding and actually use it). As written it's confusing and will likely trip the linter (unused/ineffassign).

test/basic/sigverify_test.go
```diff
@@ -203,3 +206,96 @@ func TestSubmitTXWithVerification(t *testing.T) {
...
+	signer, certBytes, err := testutil.LoadCryptoMaterialsFromDir(t, uc.MSPDir)
+	require.NoError(t, err)
+	require.NotNil(t, signer)
+	require.NotNil(t, certBytes)
+
+	broadcastClient := client.NewBroadcastTxClient(uc, 10*time.Second)
+	defer broadcastClient.Stop()
+
+	t.Log("Submit TX with known certID – expect success")
+	mspID := fmt.Sprintf("org%d", 1)
```
tock-ibm: Nit: `fmt.Sprintf("org%d", 1)` over a literal `1` is odd — just use `"org1"`.

test/basic/sigverify_test.go
```diff
@@ -203,3 +206,96 @@ func TestSubmitTXWithVerification(t *testing.T) {
...
+	err = broadcastClient.SendTx(envWithCertID)
+	require.NoError(t, err, "TX with a known cert-ID should be accepted")
+
+	t.Log("Submit TX with unknown certID – expect rejection")
+	unknownCA, err := tlsgen.NewCA()
+	require.NoError(t, err)
+	unknownClientPair, err := unknownCA.NewClientCertKeyPair()
+	require.NoError(t, err)
+	block, _ := pem.Decode(unknownClientPair.Key)
+	reLabeled := pem.EncodeToMemory(&pem.Block{Type: "PRIVATE KEY", Bytes: block.Bytes})
```
tock-ibm: **Fragile PEM relabel + missing nil check.** `block` from `pem.Decode` is used without a nil check (nil → panic on `block.Bytes`). The relabel to `"PRIVATE KEY"` only works because `tlsgen` marshals PKCS8 but labels the block `"EC PRIVATE KEY"`, while `CreateECDSAPrivateKey` requires type `"PRIVATE KEY"` + PKCS8 — a silent dependency on tlsgen internals. Please add a comment explaining *why* the relabel is needed, or extract a small `testutil` helper so this coupling lives in one place.

test/basic/sigverify_test.go
```diff
@@ -203,3 +206,96 @@ func TestSubmitTXWithVerification(t *testing.T) {
...
+	unknownPrivKey, err := tx.CreateECDSAPrivateKey(reLabeled)
+	require.NoError(t, err)
+	unknownSigner := (*crypto.ECDSASigner)(unknownPrivKey)
+
+	txContent2 := tx.PrepareTxWithTimestamp(2, 64, []byte("unknownCertIDTx"))
+	envUnknown := tx.CreateSignedStructuredEnvelopeWithCertID(txContent2, unknownSigner, unknownClientPair.Cert, mspID)
+	require.NotNil(t, envUnknown)
+	err = broadcastClient.SendTx(envUnknown)
+	require.Error(t, err, "TX with an unknown cert-ID should be rejected")
+	require.ErrorContains(t, err, "signature did not satisfy policy /Channel/Writers")
```
tock-ibm, line 286: Consider whether an *unknown cert-ID* actually surfaces `signature did not satisfy policy /Channel/Writers` (reused from the fake-signature test) rather than an identity-deserialization/lookup error. If it fails earlier, the assertion could pass for the wrong reason — worth confirming the exact error path.

testutil/tx/tx_utils.go
```diff
@@ -230,11 +232,57 @@ func PrepareSignedEnvelopeWithCertificate(txNumber int, envSize int, sessionNumb
...
+	}
+	return hex.EncodeToString(digest), nil
+}
+
+// createSignedStructuredPayloadWithCertID builds a Payload whose creator identity uses a certificate-ID
+// (hash) instead of the full certificate PEM.
+func createSignedStructuredPayloadWithCertID(data []byte, certBytes []byte, org string) *common.Payload {
+	payloadChannelHeader := createChannelHeader(common.HeaderType_MESSAGE, 0, "channelID", 0)
+
+	certID, err := certID(certBytes)
```
tock-ibm: **Name shadowing:** the local `certID` variable shadows the `certID` function on the same line. It compiles (RHS evaluated first) but is needlessly confusing. Rename the function to `computeCertID` (or rename the variable).

testutil/tx/tx_utils.go
```diff
@@ -230,11 +232,57 @@ func PrepareSignedEnvelopeWithCertificate(txNumber int, envSize int, sessionNumb
...
 func PrepareSignedEnvelopeWithCertificateID(txNumber int, envSize int, sessionNumber []byte, signer *crypto.ECDSASigner, certBytes []byte, org string) *common.Envelope {
-	return nil
+	data := PrepareTxWithTimestamp(txNumber, envSize, sessionNumber)
+	return CreateSignedStructuredEnvelopeWithCertID(data, signer, certBytes, org)
+}
+
+// CreateSignedStructuredEnvelopeWithCertID creates an envelope whose SignatureHeader.Creator carries a
+// certificate-ID (hex SHA256 of the DER-encoded cert) rather than the full PEM certificate.
+// The verifying MSP must have the corresponding certificate registered in its known-identities list.
+func CreateSignedStructuredEnvelopeWithCertID(data []byte, signer *crypto.ECDSASigner, certBytes []byte, org string) *common.Envelope {
```
tock-ibm, line 244: Minor: inconsistent error handling in the two new functions — `CreateSignedStructuredEnvelopeWithCertID` returns `nil` on sign error (silent), while `createSignedStructuredPayloadWithCertID` `panic`s on a `certID` error. The panic matches the existing `deterministicMarshall` pattern in this file so it's acceptable, but pick one convention.

tock-ibm (commented):
## Code Review — PR #1063 (cert id msp)

**Overview:** Test-only change (issue #638). Implements the previously-stubbed `PrepareSignedEnvelopeWithCertificateID` and adds `TestSubmitTXWithKnownCertID`, exercising the "short" (known-identity) signing mode where `SignatureHeader.Creator` carries a cert-ID (`hex(SHA256(cert.DER))`) instead of the full PEM cert.

The approach is sound: it matches how `fabric-x-common`'s MSP derives the known-cert identifier (`msp/identities.go` — `hex.EncodeToString(digest)` over `cert.Raw`, SHA256 default), and armageddon's cryptogen populates `knowncerts` with each node/user signcert, so the accept path is well-founded. Good coverage of both accept (known) and reject (unknown) cases plus a ledger-pull confirmation.

Main things to address before merge are inline. The `parties` block is a bug + dead code and may trip the strict linter — worth running `make linter` locally.

testutil/tx/tx_utils.go
```diff
@@ -230,11 +232,57 @@ func PrepareSignedEnvelopeWithCertificate(txNumber int, envSize int, sessionNumb
...
+	}
+}
+
+func certID(certBytes []byte) (string, error) {
+	digest, err := certificate.DigestPemContent(certBytes, bccsp.SHA256)
+	if err != nil {
+		return "", fmt.Errorf("failed computing cert ID: %w", err)
+	}
+	return hex.EncodeToString(digest), nil
+}
```
tock-ibm: Isn't there already a method in the MSP that does that?

testutil/tx/tx_utils.go
```diff
@@ -230,11 +232,57 @@ func PrepareSignedEnvelopeWithCertificate(txNumber int, envSize int, sessionNumb
...
+
+// createSignedStructuredPayloadWithCertID builds a Payload whose creator identity uses a certificate-ID
+// (hash) instead of the full certificate PEM.
+func createSignedStructuredPayloadWithCertID(data []byte, certBytes []byte, org string) *common.Payload {
+	payloadChannelHeader := createChannelHeader(common.HeaderType_MESSAGE, 0, "channelID", 0)
```
tock-ibm, line 269: Isn't channel-ID called "arma"

MayRosenbaum: I will change it, but I didnt see any effect under another channel name

tock-ibm: That is not a good sign because the router should check that the channel ID in the channel header matches that of the genesis

testutil/tx/tx_utils.go
MayRosenbaum: I followed the error handling in`CreateSignedStructuredEnvelope`

testutil/tx/tx_utils.go
MayRosenbaum: I followed what is used in `fabric-x-committer`.
In `fabric-x-common` we do have methods to create a `SignatureHeader` with a given `msp.SigningIdentity`.
For this, we need to make some changes. 
Maybe we should keep this in a separate PR?

testutil/tx/tx_utils.go
MayRosenbaum: The channel ID check is currently commented out, so we are not validating it
If we uncomment it, then the test fails.

tock-ibm: Resolved: channelID is ignored on data TXs by design. Enforced only on config-update TX.

testutil/tx/tx_utils.go
MayRosenbaum: I opened issue #1076

tock-ibm (approved):

Issue #1029, opened 2026-07-21
Extend armageddon load to sign with Cert-ID


tock-ibm:
Armageddon load should support generating TXs signed by Cert-ID - "short" mode.

PR #1064, opened 2026-07-29
load command with cert id

MayRosenbaum:
issue #1029

### DIFF
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -612,8 +612,15 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
 			broadcastClient.SendTxToAllRouters(env)
 		}
 	case "short":
-		fmt.Fprintf(os.Stderr, "Short signing mode is not yet supported")
-		os.Exit(3)
+		for i := 0; i < numOfTxs; i++ {
+			env = tx.PrepareSignedEnvelopeWithCertificateID(i, txSize, sessionNumber, signer, certBytes, org)
+			status := rl.GetToken()
+			if !status {
+				fmt.Fprintf(os.Stderr, "failed to send tx %d in signed mode %s", i+1, signedMode)
+				os.Exit(3)
+			}
+			broadcastClient.SendTxToAllRouters(env)
+		}
 	default:
 		for i := 0; i < numOfTxs; i++ {
 			env = tx.PrepareUnsignedEnvelope(i, txSize, sessionNumber)
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -604,6 +604,10 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
 	case "full":
 		for i := 0; i < numOfTxs; i++ {
 			env = tx.PrepareSignedEnvelopeWithCertificate(i, txSize, sessionNumber, signer, certBytes, org)
+			if env == nil {
+				fmt.Fprintf(os.Stderr, "failed to prepare envelope %d in signed mode %s", i+1, signedMode)
+				os.Exit(3)
+			}
 			status := rl.GetToken()
 			if !status {
 				fmt.Fprintf(os.Stderr, "failed to send tx %d in signed mode %s", i+1, signedMode)
@@ -614,6 +618,10 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
 	case "short":
 		for i := 0; i < numOfTxs; i++ {
 			env = tx.PrepareSignedEnvelopeWithCertificateID(i, txSize, sessionNumber, signer, certBytes, org)
+			if env == nil {
+				fmt.Fprintf(os.Stderr, "failed to prepare envelope %d in signed mode %s", i+1, signedMode)
+				os.Exit(3)
+			}
 			status := rl.GetToken()
 			if !status {
 				fmt.Fprintf(os.Stderr, "failed to send tx %d in signed mode %s", i+1, signedMode)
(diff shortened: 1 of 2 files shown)


common/tools/armageddon/armageddon.go
```diff
@@ -587,8 +587,15 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
 			broadcastClient.SendTxToAllRouters(env)
 		}
 	case "short":
-		fmt.Fprintf(os.Stderr, "Short signing mode is not yet supported")
-		os.Exit(3)
+		for i := 0; i < numOfTxs; i++ {
+			env = tx.PrepareSignedEnvelopeWithCertificateID(i, txSize, sessionNumber, signer, certBytes, org)
```
tock-ibm, line 620: **Blocking:** this calls `tx.PrepareSignedEnvelopeWithCertificateID`, which is still an unimplemented stub returning `nil` in `testutil/tx/tx_utils.go` (not modified in this PR). As a result short mode builds `nil` envelopes; `SendTxToAllRouters(nil)` then transmits an empty `Envelope`, not a signed one. The cert-by-ID builder must be implemented here (or land first) for this feature to do anything.

common/tools/armageddon/armageddon_test.go
```diff
@@ -276,6 +276,63 @@ func TestLoadAndReceive(t *testing.T) {
...
+// Scenario:
+//  1. Create a config YAML file to be an input to armageddon
+//  2. Run armageddon generate command to create config files in a folder structure
+//  3. Run arma with the generated config files to run each of the nodes for all parties
+//  4. Run armageddon load command with --signedMode=short
+//  5. Run armageddon receive command to verify transactions are received
+//
+// This test verifies that the --signedMode flag works correctly with the load and receive commands.
+// It also verifies cryptographic signature correctness.
+func TestLoadAndReceiveSignedModeAsFull(t *testing.T) {
```
tock-ibm: Name is inverted vs. the body: this function sets `signedMode := "short"` (line 318) yet is named `...AsFull`. Its doc comment above also says step 4 uses `--signedMode=short` while claiming it verifies signature correctness — which nothing here does. Swap the two function names so name/comment/body agree.

common/tools/armageddon/armageddon_test.go
```diff
@@ -285,8 +342,7 @@ func TestLoadAndReceive(t *testing.T) {
 //
 // This test verifies that the --signedMode flag works correctly with the load and receive commands.
 // It does NOT verify cryptographic signature correctness, and test only the full option
-// TODO: add another test for "--signedMode=short" as well.
-func TestLoadAndReceiveSignedModeAsFull(t *testing.T) {
+func TestLoadAndReceiveSignedModeAsShort(t *testing.T) {
```
tock-ibm: Mirror of the naming issue: this renamed function runs `signedMode := "full"` (line 375) but is now called `...AsShort`, and still carries the old comment "test only the full option / does NOT verify cryptographic signature correctness." The `Full`/`Short` names are reversed relative to their bodies.

common/tools/armageddon/armageddon_test.go
```diff
@@ -276,6 +276,63 @@ func TestLoadAndReceive(t *testing.T) {
...
+	defer os.RemoveAll(dir)
+
+	// 1. Create network config
+	configPath := filepath.Join(dir, "config.yaml")
+	netInfo := testutil.CreateNetwork(t, configPath, 4, 2, "none", "none")
+	defer netInfo.CleanUp()
+
+	// 2. Generate armageddon config files
+	armageddonCLI := armageddon.NewCLI()
+	armageddonCLI.Run([]string{"generate", "--config", configPath, "--output", dir, "--clientSignatureVerificationRequired"})
```
tock-ibm, line 300: This test generates with `--clientSignatureVerificationRequired`, so routers will reject the empty/unsigned envelopes produced by the stub. `load` doesn't check acks and returns, but the parallel `receive` blocks (`BLOCK_UNTIL_READY`) waiting for 200 txs that never reach the assembler, so `waitForLoadAndReceive.Wait()` never completes → the test hangs until the package timeout. Also: it asserts nothing about how many txs were received, so it can only hang, never meaningfully fail.

tock-ibm (commented):
## Review summary

This PR wires up the `--signedMode=short` path in `armageddon load`, adds an integration test for it, and renames the existing signed-mode test.

**Requesting changes** — three blocking issues:

1. **The feature does not work.** The `short` case calls `tx.PrepareSignedEnvelopeWithCertificateID(...)`, but that function is still an unimplemented stub returning `nil` (in `testutil/tx/tx_utils.go`, not touched by this PR). Short mode therefore builds `nil` envelopes and sends empty `Envelope`s to the routers — not signed ones. The cert-by-ID builder needs to land in the same PR (or first).

2. **Test names are swapped against their bodies.** `TestLoadAndReceiveSignedModeAsFull` runs `signedMode := "short"`, and the renamed `TestLoadAndReceiveSignedModeAsShort` runs `signedMode := "full"`. The doc comments are inverted too.

3. **The new short test will hang in CI.** It uses `generate --clientSignatureVerificationRequired`, so routers reject the empty/unsigned envelopes; `load` doesn't check acks, but `receive` blocks on `BLOCK_UNTIL_READY` waiting for 200 txs that never arrive, so `WaitGroup.Wait()` never returns → package timeout. Worth confirming the CI run actually executed these tests.

Minor: the tests assert nothing beyond `Wait()` — no received-tx count, no error checks — so they can't fail on regression, only hang. Consider asserting on the receive output. The two mode branches are also near-duplicates; once both work, factor the loop into a helper parameterized by the envelope builder.

🤖 Posted via Claude Code on behalf of Yoav Tock

common/tools/armageddon/armageddon_test.go
MayRosenbaum: dont understand.

tock-ibm (approved):

PR #1072, opened 2026-07-30
Integration tests (rejoin party) - reconfiguration

gengur:
Verifies that a party that has been stopped from can rejoin the network through a configuration update
issue #247

### DIFF
--- a/common/utils/net.go
+++ b/common/utils/net.go
@@ -110,21 +110,29 @@ func CertificateToString(cert *x509.Certificate) string {
 }
 
 func GetPortFromEndpoint(endpoint string) uint32 {
-	_, portS, err := net.SplitHostPort(endpoint)
-	if err != nil || len(portS) == 0 {
-		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: %v", endpoint, err))
-	}
-	port, err := strconv.ParseUint(portS, 10, 32)
-	if err != nil {
-		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: %v", endpoint, err))
-	}
-	return uint32(port)
+	_, port := ParseEndpoint(endpoint)
+	return port
 }
 
 func TrimPortFromEndpoint(endpoint string) string {
-	host, _, err := net.SplitHostPort(endpoint)
-	if err != nil || len(host) == 0 {
+	host, _ := ParseEndpoint(endpoint)
+	return host
+}
+
+func ParseEndpoint(endpoint string) (string, uint32) {
+	host, portS, err := net.SplitHostPort(endpoint)
+	if err != nil {
 		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: %v", endpoint, err))
 	}
-	return host
+	if host == "" {
+		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: host is empty", endpoint))
+	}
+	port, err := strconv.ParseUint(portS, 10, 16) // bit-size 16 rejects ports > 65535
+	if err != nil {
+		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: invalid port %s: %v", endpoint, portS, err))
+	}
+	if port == 0 {
+		panic(fmt.Sprintf("endpoint %s is not a valid host:port string: port must be greater than 0", endpoint))
+	}
+	return host, uint32(port)
 }
--- a/test/reconfig/membership/config_membership_update_test.go
+++ b/test/reconfig/membership/config_membership_update_test.go
@@ -35,6 +35,7 @@ import (
 	"github.com/onsi/gomega/gexec"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
 )
 
 // TestRemovePartyRunAll verifies that removing a party via a config update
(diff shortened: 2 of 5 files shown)


tock-ibm (commented):
## Code Review

Adds an integration test `TestReJoinSingleParty` verifying that a party stopped from the network can rejoin via a config update, plus supporting CI/Makefile wiring, a new `utils.ParseEndpoint` helper, a `UpdateConsenterEndpoint` config-update builder, and a `RestartArmaNode` change that rewrites the bootstrap block path.

The CI job, Makefile target, and skip-regex wiring are correct and consistent with the existing full-parties-replacement pattern, and the production `RestartArmaNode` change is properly gated. No blocking correctness bug in the production change. Main things I'd want addressed before merge:

1. **Replace the 60s fixed sleep with a poll** — fixed sleeps are the classic source of flaky/slow integration tests.
2. **Justify or eliminate the `UpdateConsenterEndpoint` / `UpdateOrderingEndpoint` duplication** — the new method is a near-verbatim copy of the existing one, differing only by a `syncBlockValidationPolicy` call whose necessity for an endpoint-only change is unclear.
3. **Check the ignored `WriteToYAML` errors and dead `Stop()` call.**

Inline comments below. Note: reviewed statically; I did not run the new suite — worth confirming it's green and non-flaky across a couple of CI runs given the fixed-sleep timing.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

test/reconfig/membership/config_membership_update_test.go
```diff
@@ -1353,6 +1354,290 @@ func TestJoinMultipleParties(t *testing.T) {
...
+	for _, net := range netToRejoin {
+		net.ConfigBlockPath = configBlockPath
+	}
+
+	// Restart the rejoining party's nodes with the updated config block
+	armaNetwork.StopParties([]types.PartyID{partyToRejoin})
+	armaNetwork.RestartParties(t, []types.PartyID{partyToRejoin}, readyChan)
+	testutil.WaitReady(t, readyChan, nodesPerParty, 10)
+
+	time.Sleep(60 * time.Second) // give the rejoining party some time to finish syncing
```
tock-ibm: **Fixed 60s sleep for sync.** Fixed sleeps are the classic source of flaky/slow integration tests: too short under CI load leads to flakes, too long wastes time. Prefer polling a readiness/height condition — the suite already has `WaitReady`, `WaitForNetworkRelaunch`, and assembler pulls with timeouts you could reuse. If a sleep is truly unavoidable here, a comment explaining why polling isn't possible would help.

test/reconfig/membership/config_membership_update_test.go
```diff
@@ -1353,6 +1354,290 @@ func TestJoinMultipleParties(t *testing.T) {
...
+		err = configtxgen.WriteOutputBlock(lastConfigBlock, configBlockPath)
+		require.NoError(t, err)
+
+		// Update the nodes endpoints in the local config files to use the new endpoints from the config update
+		consenterNodeConfigPath := filepath.Join(dir, "config", fmt.Sprintf("party%d", partyID), "local_config_consenter.yaml")
+		localConfig, _, err := config.LoadLocalConfig(consenterNodeConfigPath, logger)
+		require.NoError(t, err)
+		localConfig.NodeLocalConfig.GeneralConfig.ListenAddress = newConsenterHostName
+		localConfig.NodeLocalConfig.GeneralConfig.ListenPort = newConsenterPortUint
+		utils.WriteToYAML(localConfig.NodeLocalConfig, consenterNodeConfigPath)
```
tock-ibm: The returned `error` from `utils.WriteToYAML` is dropped here (and at line 1522). The strict `.golangci.yml` `errcheck` set may flag these even in test files. Wrap with `require.NoError(t, err)`.

test/reconfig/membership/config_membership_update_test.go
```diff
@@ -1353,6 +1354,290 @@ func TestJoinMultipleParties(t *testing.T) {
...
+
+		netToRelaunch := maps.Collect(filterMap(netInfo, func(k testutil.NodeName, v *testutil.ArmaNodeInfo) bool {
+			return k.PartyID != partyID || (k.PartyID == partyID && (k.NodeType == testutil.Router || k.NodeType == testutil.Assembler))
+		}))
+
+		testutil.WaitForNetworkRelaunch(t, netToRelaunch, configSeq)
+
+		// Read the last config block to get the updated config
+		logger := flogging.MustGetLogger("TestReJoinSingleParty")
+		routerNodeConfigPath := filepath.Join(dir, "config", fmt.Sprintf("party%d", partyID), "local_config_consenter.yaml")
```
tock-ibm: `routerNodeConfigPath` is assigned `local_config_consenter.yaml` and `config.ReadConfig` is called on it — the name is misleading. It's also the same path re-read a few lines below as `consenterNodeConfigPath` (line 1510). Consider computing it once and naming it `consenterNodeConfigPath`.

test/reconfig/membership/config_membership_update_test.go
```diff
@@ -1353,6 +1354,290 @@ func TestJoinMultipleParties(t *testing.T) {
...
+		builder.UpdateBatcherEndpoint(t, partyID, types.ShardID(1), newBatcherHostName, int(newBatcherPortUint))
+
+		env := configutil.CreateConfigTX(t, dir, remainingParties, int(submittingParty), builder.ConfigUpdatePBData(t))
+		require.NotNil(t, env)
+
+		// Send the config tx to the remaining parties
+		err = broadcastTxWithRetry(t, uc, env, remainingParties, 5)
+		require.NoError(t, err)
+
+		broadcastClient.Stop()
```
tock-ibm: This `broadcastClient.Stop()` acts on the outer client that was already stopped before the loop (the per-attempt client inside `broadcastTxWithRetry` cleans up via its own `t.Cleanup`). `Stop()` is idempotent so it's harmless, but it's dead/confusing code — remove it.

test/reconfig/membership/config_membership_update_test.go
```diff
@@ -1353,6 +1354,290 @@ func TestJoinMultipleParties(t *testing.T) {
...
+
+	netToRejoin := maps.Collect(filterMap(netInfo, func(k testutil.NodeName, v *testutil.ArmaNodeInfo) bool {
+		return k.PartyID == partyToRejoin
+	}))
+	maps.DeleteFunc(netInfo, func(nodeName testutil.NodeName, _ *testutil.ArmaNodeInfo) bool {
+		return nodeName.PartyID == partyToRejoin
+	})
+
+	configBlockPath := filepath.Join(dir, "bootstrap", "bootstrap.block")
+	submittingParty := types.PartyID(2)
```
tock-ibm: `submittingParty := types.PartyID(2)` duplicates `submittingPartyID := types.PartyID(2)` from line 1370 — same value. Collapse to a single variable and reuse it for the `require.NotEqual` guard.

common/utils/net.go
```diff
@@ -128,3 +128,21 @@ func TrimPortFromEndpoint(endpoint string) string {
 	}
 	return host
 }
+
+func ParseEndpoint(endpoint string) (string, uint32) {
```
tock-ibm, line 122: `ParseEndpoint` largely duplicates `GetPortFromEndpoint` + `TrimPortFromEndpoint` in this same file. It does add real value (bit-size 16 correctly rejects ports > 65535, plus empty-host and zero-port checks), but consider having `GetPortFromEndpoint`/`TrimPortFromEndpoint` delegate to it rather than keeping three near-identical panic-on-parse helpers. Also note this lands in production `common/utils` but is currently consumed only by tests.

testutil/configutil/config_update_utils.go
```diff
@@ -846,6 +846,45 @@ func (c *ConfigUpdateBuilder) UpdateRouterEndpoint(t *testing.T, partyID types.P
 	return c.createConfigUpdate(t, c.configData)
 }
 
+func (c *ConfigUpdateBuilder) UpdateConsenterEndpoint(t *testing.T, partyID types.PartyID, host string, port int) []byte {
```
tock-ibm: `UpdateConsenterEndpoint` is a near-verbatim copy of the existing `UpdateOrderingEndpoint` (both mutate `ConsenterConfig` host/port and the `consenter_mapping` host/port). The only difference is the added `syncBlockValidationPolicy` call. Two questions: (1) Why a new method rather than adding the sync call to the existing one, or delegating? (2) `syncBlockValidationPolicy` recomputes the block-validation policy from `msp_id`/`identity` certs, which an endpoint (host/port) change does not touch — is that call actually required here, or copied defensively? A short comment would justify it.

testutil/arma_network.go
```diff
@@ -205,6 +206,12 @@ func (armaNetwork *ArmaNetwork) GetBatcher(t *testing.T, partyID types.PartyID,
 func (armaNodeInfo *ArmaNodeInfo) RestartArmaNode(t *testing.T, readyChan chan string) {
 	require.FileExists(t, armaNodeInfo.RunInfo.NodeConfigPath)
 	nodeConfig := ReadNodeConfigFromYaml(t, armaNodeInfo.RunInfo.NodeConfigPath)
+	if len(armaNodeInfo.ConfigBlockPath) > 0 && nodeConfig.GeneralConfig.Bootstrap.File != armaNodeInfo.ConfigBlockPath {
+		nodeConfig.GeneralConfig.Bootstrap.File = armaNodeInfo.ConfigBlockPath
```
tock-ibm: `RestartArmaNode` is used by every restart-based test. This new block rewrites `Bootstrap.File` whenever `ConfigBlockPath` is set and differs. It's gated correctly, but worth confirming no existing test sets `ConfigBlockPath` and relies on the previously-written YAML bootstrap path (previously `ConfigBlockPath` was only consumed at initial start via `EditDirectoryInNodeConfigYAML`).

test/reconfig/membership/config_membership_update_test.go
gengur: fixed as suggested

test/reconfig/membership/config_membership_update_test.go
gengur: fixed as suggested

test/reconfig/membership/config_membership_update_test.go
gengur: fixed as suggested

test/reconfig/membership/config_membership_update_test.go
gengur: fixed as suggested

common/utils/net.go
gengur: fixed as suggested

testutil/configutil/config_update_utils.go
gengur: removed the UpdateConsenterEndpoint method as obsolete

test/reconfig/membership/config_membership_update_test.go
gengur: fixed as suggested

testutil/arma_network.go
gengur: fixed

tock-ibm (approved):

Issue #857, opened 2026-04-27
Cluster parameters in local config


tock-ibm:
Add the following cluster parameters to the local config `General.Cluster`
```
// TODO add the fields:
//
//      RPCTimeout                     time.Duration // The timeout for RPC calls between orderers.
//		ReplicationBufferSize          int           // The size of the buffer that is used to store blocks that are being replicated from other orderers.
//		ReplicationMaxRetryInterval    time.Duration // The maximal amount of time to wait before retrying to connect.
//	    ReplicationMinRetryInterval    time.Duration // The minimal amount of time to wait before retrying to connect.
//	    ReplicationMaxRetryDuration    time.Duration // The maximal amount of time to retry connecting before giving up.
//      CertExpirationWarningThreshold time.Duration // The amount of time before TLS certificate expiration to start warning about impending expiration.
//      TLSHandshakeTimeShift          time.Duration // The duration by which to shift TLS handshakes time sampling to the past
```

Use these parameters in 

`node/consensus/synchronizer/synchronizer_bft.go` 

`node/consensus/synchronizer/height_detector_factory.go`

`node/consensus/consensus_builder.go` when building 

`c.ClusterService`, 
`commAuth := &comm.AuthCommMgr{` and 
`c.BFT.Comm = &comm.Egress{`

PR #1074, opened 2026-08-02
Add cluster parameters to local config

moradna:
#857

### DIFF
--- a/common/synchronizer/block_puller.go
+++ b/common/synchronizer/block_puller.go
@@ -85,10 +85,10 @@ func NewBlockPuller(
 		MyOwnTLSCert:        myCert,
 		VerifyBlockSequence: verifyBlockSequenceNoOp,
 		Logger:              logger,
-		RetryTimeout:        time.Second * 5,  // clusterConfig.ReplicationRetryTimeout,
-		MaxTotalBufferBytes: 20 * 1024 * 1024, // clusterConfig.ReplicationBufferSize,
-		FetchTimeout:        time.Second * 5,  // clusterConfig.ReplicationPullTimeout,
-		Endpoints:           endpoints,        // TODO the block puller is not party aware yet
+		RetryTimeout:        time.Second * 5, // clusterConfig.ReplicationRetryTimeout,
+		MaxTotalBufferBytes: clusterConfig.ReplicationBufferSize,
+		FetchTimeout:        time.Second * 5, // clusterConfig.ReplicationPullTimeout,
+		Endpoints:           endpoints,       // TODO the block puller is not party aware yet
 		Signer:              support,
 		TLSCert:             der.Bytes,
 		Channel:             support.ChannelID(),
--- a/common/synchronizer/block_puller.go
+++ b/common/synchronizer/block_puller.go
@@ -85,10 +85,10 @@ func NewBlockPuller(
 		MyOwnTLSCert:        myCert,
 		VerifyBlockSequence: verifyBlockSequenceNoOp,
 		Logger:              logger,
-		RetryTimeout:        time.Second * 5, // clusterConfig.ReplicationRetryTimeout,
+		RetryTimeout:        time.Second * 5,
 		MaxTotalBufferBytes: clusterConfig.ReplicationBufferSize,
-		FetchTimeout:        time.Second * 5, // clusterConfig.ReplicationPullTimeout,
-		Endpoints:           endpoints,       // TODO the block puller is not party aware yet
+		FetchTimeout:        time.Second * 5,
+		Endpoints:           endpoints, // TODO the block puller is not party aware yet
 		Signer:              support,
 		TLSCert:             der.Bytes,
 		Channel:             support.ChannelID(),
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -17,8 +17,14 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
 	GeneralConfig: &GeneralConfig{
 		LogSpec: "info",
 		Cluster: ClusterYaml{
-			SendBufferSize:    100,
-			ReplicationPolicy: "consensus",
+			SendBufferSize:                 100,
+			ReplicationPolicy:              "consensus",
+			RPCTimeout:                     time.Minute,
+			ReplicationBufferSize:          20971520,
+			ReplicationMaxRetryInterval:    10 * time.Second,
+			ReplicationMinRetryInterval:    200 * time.Millisecond,
+			ReplicationMaxRetryDuration:    time.Minute,
+			CertExpirationWarningThreshold: time.Hour * 24 * 7,
 		},
 	},
 	OperationsConfig: &Operations{
--- a/config/local_config.go
+++ b/config/local_config.go
@@ -136,6 +136,20 @@ type ClusterYaml struct {
 	ClientPrivateKey string `yaml:"ClientPrivateKey,omitempty"`
 	// ReplicationPolicy defines how blocks are replicated between orderers.
 	ReplicationPolicy string `yaml:"ReplicationPolicy,omitempty"`
+	// RPCTimeout is the timeout for RPC calls between orderers.
+	RPCTimeout time.Duration `yaml:"RPCTimeout,omitempty"`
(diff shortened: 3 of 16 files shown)


tock-ibm:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

config/defaults.go
```diff
@@ -17,8 +17,14 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
...
-			SendBufferSize:    100,
-			ReplicationPolicy: "consensus",
+			SendBufferSize:                 100,
+			ReplicationPolicy:              "consensus",
+			RPCTimeout:                     time.Minute,
+			ReplicationBufferSize:          20971520,
+			ReplicationMaxRetryInterval:    10 * time.Second,
+			ReplicationMinRetryInterval:    200 * time.Millisecond,
+			ReplicationMaxRetryDuration:    time.Minute,
+			CertExpirationWarningThreshold: time.Hour * 24 * 7,
```
tock-ibm, line 27: Default behavior change: this becomes the default for `ClusterService.CertExpWarningThreshold` (see `node/consensus/consensus_builder.go:409`), which was previously hardcoded to `time.Hour`. So the cert-expiration warning window shifts from 1h to 7 days for default deployments.

Every other new field's default was chosen to preserve the old hardcoded value (RPCTimeout=1m, ReplicationBufferSize=20MiB, retry intervals 200ms/10s/1m). This one intentionally diverges to match Fabric's upstream default (`time.Hour*24*7`). Looks deliberate and reasonable, but worth calling out in the PR description / release notes since it's the one behavior change here.

common/synchronizer/block_puller.go
```diff
@@ -85,10 +85,10 @@ func NewBlockPuller(
 		MyOwnTLSCert:        myCert,
 		VerifyBlockSequence: verifyBlockSequenceNoOp,
 		Logger:              logger,
-		RetryTimeout:        time.Second * 5,  // clusterConfig.ReplicationRetryTimeout,
-		MaxTotalBufferBytes: 20 * 1024 * 1024, // clusterConfig.ReplicationBufferSize,
-		FetchTimeout:        time.Second * 5,  // clusterConfig.ReplicationPullTimeout,
-		Endpoints:           endpoints,        // TODO the block puller is not party aware yet
+		RetryTimeout:        time.Second * 5, // clusterConfig.ReplicationRetryTimeout,
+		MaxTotalBufferBytes: clusterConfig.ReplicationBufferSize,
+		FetchTimeout:        time.Second * 5, // clusterConfig.ReplicationPullTimeout,
```
tock-ibm: `MaxTotalBufferBytes` is now wired to `clusterConfig.ReplicationBufferSize`, but `RetryTimeout` and `FetchTimeout` are still hardcoded to `5s` with trailing comments referencing `clusterConfig.ReplicationRetryTimeout` / `clusterConfig.ReplicationPullTimeout` — field names that were never added (the PR added `ReplicationMaxRetryInterval` / `ReplicationMinRetryInterval` / `ReplicationMaxRetryDuration` instead, wired only in the consensus synchronizer). No behavior change, but the comments now point at non-existent config fields. Consider either wiring these up or updating/removing the stale comments so the next person isn't misled.

node/consensus/synchronizer/synchronizer_bft.go
```diff
@@ -266,11 +268,11 @@ func (s *BFTSynchronizer) createBFTDeliverer(startHeight uint64, myParty arma_ty
 	clientConfig.SecOpts.VerifyCertificate = nil
 
 	// The maximal amount of time to wait before retrying to connect.
-	maxRetryInterval := 10 * time.Second // TODO s.LocalConfigCluster.ReplicationRetryTimeout
+	maxRetryInterval := s.LocalConfigCluster.ReplicationMaxRetryInterval
 	// The minimal amount of time to wait before retrying. The retry interval doubles after every unsuccessful attempt.
-	minRetryInterval := maxRetryInterval / 50
+	minRetryInterval := s.LocalConfigCluster.ReplicationMinRetryInterval
```
tock-ibm, line 273: Now that min/max retry intervals and max duration are independently configurable, the invariant that was previously guaranteed structurally (`minRetryInterval = maxRetryInterval / 50`, so `min <= max`) is no longer enforced. Nothing in `applyGeneralDefaults` or `validateLocalConfig` checks `Min <= Max` or `MaxRetryInterval <= MaxRetryDuration`. Defaults are safe, but a misconfiguration would silently break the exponential-backoff-with-cap semantics. A sanity check would be nice to have (optional / follow-up).

test/utils/utils.go
```diff
@@ -927,3 +927,24 @@ func SendTransactions(t *testing.T, routers []*router.Router, assembler *assembl
 
 	t.Logf("%f (totalTxn / elapsed)\n", float32(totalTxn)/float32(elapsed))
 }
+
+// configurationWithDefaultCluster returns a configuration with the default cluster settings.
+func configurationWithDefaultCluster() *config.Configuration {
```
tock-ibm: This helper is duplicated byte-for-byte in `node/consensus/setup_test.go:362`. Since CLAUDE.md notes test helpers live in `testutil/`, consider hoisting a single exported helper there and calling it from both sites rather than copy-pasting. Minor.

tock-ibm (commented):
### Code review

A few things worth confirming before merge. None are hard blockers — the defaults preserve existing behavior in most cases — but a couple could use a second look.

🤖 Generated with [Claude Code](https://claude.ai/code)

common/synchronizer/block_puller.go
tock-ibm: True

config/defaults.go
tock-ibm: This is good

node/consensus/synchronizer/synchronizer_bft.go
tock-ibm: True

test/utils/utils.go
tock-ibm: true

config/defaults.go
moradna: I used Fabric's default, would you prefer to keep the previous 1h default?

node/consensus/synchronizer/synchronizer_bft.go
moradna: Good point. I'll add the validation in the next local config validation PR

config/defaults.go
tock-ibm: I meant what you did is good :-)

config/defaults.go
```diff
@@ -17,8 +17,14 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
 	GeneralConfig: &GeneralConfig{
 		LogSpec: "info",
 		Cluster: ClusterYaml{
-			SendBufferSize:    100,
-			ReplicationPolicy: "consensus",
+			SendBufferSize:                 100,
+			ReplicationPolicy:              "consensus",
```
tock-ibm, line 21: ReplicationPolicy:  empty string ""

config/defaults.go
```diff
@@ -17,8 +17,14 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
...
+			RPCTimeout:                     time.Minute,
+			ReplicationBufferSize:          20971520,
```
tock-ibm, line 23: Lets increase the default to 50MB

config/local_config.go
```diff
@@ -136,6 +136,20 @@ type ClusterYaml struct {
 	ClientPrivateKey string `yaml:"ClientPrivateKey,omitempty"`
 	// ReplicationPolicy defines how blocks are replicated between orderers.
 	ReplicationPolicy string `yaml:"ReplicationPolicy,omitempty"`
+	// RPCTimeout is the timeout for RPC calls between orderers.
+	RPCTimeout time.Duration `yaml:"RPCTimeout,omitempty"`
+	// ReplicationBufferSize is the size of the buffer that is used to store blocks that are being replicated from other orderers.
```
tock-ibm, line 141: add that it is measured in bytes, increase the default to 50MB

tock-ibm (approved):

Issue #1069, opened 2026-07-30
validateBlockValidationPolicy does not validate the signature policy rule tree


DorKatzelnick:
The orderer rules verifier validates the BlockValidation policy's type, top-level NOutOf.N (BFT quorum), and that Identities is exactly a permutation of the consenters. It never inspects NOutOf.Rules.

 a config update with:
Rule:       NOutOf{N: 3, Rules: [SignedBy(0), NOutOf(0, []), NOutOf(0, [])]}
Identities: <all consenters, correct>
passes validation today but is satisfied by a single signature. 

Validate the rule tree against the canonical shape emitted by policies.EncodeBFTBlockVerificationPolicy — NOutOf(quorum, [SignedBy(0)…SignedBy(n-1)]):

- [ ]  Check that len(nOutOf.Rules) == len(consenters)
- [ ]  Check thaat every rule is a SignaturePolicy_SignedBy — reject nested NOutOf and nil rules
- [ ]  SignedBy indices within [0, len(Identities)) and pairwise distinct, so each identity is referenced exactly once (this is the check that closes the bypass; the first two alone still permit duplicate indices)
- [ ]  Fix reported error "quorum mismatch expected %d got %d" for the relevant one.
- [ ] Add negative tests in TestValidateBlockValidationPolicy, extending the existing buildBlockValidationPolicy helper: duplicate SignedBy, nested NOutOf(0, []), short rule list, out-of-range index

PR #1077, opened 2026-08-02
Validate BlockValidationPolicy rule tree

moradna:
#1069

### DIFF
--- a/config/verify/orderer_rules.go
+++ b/config/verify/orderer_rules.go
@@ -549,8 +549,37 @@ func validateBlockValidationPolicy(policy *common.ConfigPolicy, consenters []*co
 	expectedQuorum := int32(policies.ComputeBFTQuorum(n, f))
 
 	nOutOf := envelope.GetRule().GetNOutOf()
-	if nOutOf == nil || nOutOf.N != expectedQuorum {
-		return errors.Errorf("quorum mismatch expected %d got %d", expectedQuorum, nOutOf.GetN())
+	if nOutOf == nil {
+		return errors.New("block validation policy rule is not NOutOf")
+	}
+	if nOutOf.N != expectedQuorum {
+		return errors.Errorf("quorum mismatch expected %d got %d", expectedQuorum, nOutOf.N)
+	}
+
+	// verify that the policy has one SignedBy rule for each consenter
+	if len(nOutOf.Rules) != n {
+		return errors.Errorf("unexpected number of policy rules: expected %d got %d", n, len(nOutOf.Rules))
+	}
+
+	seenSignedBy := make([]bool, n)
+	for i, rule := range nOutOf.Rules {
+		if rule == nil {
+			return errors.Errorf("policy rule %d is nil", i)
+		}
+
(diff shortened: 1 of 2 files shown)


DorKatzelnick (approved):

DorKatzelnick:
check failed test in CI

config/verify/orderer_rules.go
```diff
@@ -549,8 +549,37 @@ func validateBlockValidationPolicy(policy *common.ConfigPolicy, consenters []*co
...
+	}
+
+	// verify that the policy has one SignedBy rule for each consenter
+	if len(nOutOf.Rules) != n {
+		return errors.Errorf("unexpected number of policy rules: expected %d got %d", n, len(nOutOf.Rules))
+	}
+
+	seenSignedBy := make([]bool, n)
+	for i, rule := range nOutOf.Rules {
+		if rule == nil {
```
tock-ibm, line 566: This `rule == nil` branch is effectively dead when reached through a marshaled proto: protobuf-go decodes a nil repeated-message element as an empty message, so unmarshal yields a non-nil `*SignaturePolicy` with a nil `Type`, which the type assertion below catches instead. Correspondingly, the `"nil policy rule is rejected"` test asserts on `"policy rule 3 is not SignedBy"` — i.e. it exercises the type-assertion path, not this branch. Harmless defensive code worth keeping, but note this branch has no real coverage.

config/verify/orderer_rules_internal_test.go
```diff
@@ -194,15 +268,12 @@ func (o *testOrdererOrg) MSPID() string       { return "" }
 func (o *testOrdererOrg) MSP() fabmsp.MSP     { return o.mspImpl }
```
tock-ibm, line 268: Optional: the SignedBy indices intentionally need not be ordered — e.g. `SignedBy(2), SignedBy(1), SignedBy(0), SignedBy(3)` should still pass. A happy-path permutation case would lock in that ordering is not required and guard against a future "tighten" that accidentally assumes it. Similarly, a `SignedBy(-1)` case would give real coverage to the `idx < 0` guard.

tock-ibm (commented):
## Validate BlockValidationPolicy rule tree — review

Genuine security/correctness hardening, and cleanly done. Previously only the top-level `NOutOf.N` and the identity set were checked, so a config block could carry the correct quorum + identities but an arbitrary rule subtree (nested policies, duplicate `SignedBy` indices, or fewer rules than consenters) — satisfiable by the wrong/fewer signers. The PR now enforces a bijection: exactly `n` rules, each a `SignedBy` with a distinct in-range index.

**Strengths**
- Splitting `nOutOf == nil` into its own branch fixes the previously misleading `"quorum mismatch expected N got 0"` message when the rule simply wasn't `NOutOf`.
- Negative-index guard (`idx < 0`) is worthwhile since `SignedBy` is an `int32`.
- Combined with the pre-existing identity check, the `SignedBy idx` → `Identities[idx]` mapping is now fully coherent and bounds-safe.
- Good negative-test coverage; the variadic `buildBlockValidationPolicy` refactor is clean and reusable.

**Verdict: approve.** Nitpicks are inline and all optional.

PR #1083, opened 2026-08-05
Propagate original tx_id into CONFIG envelope

elishambadi:
Extract the original tx_id from the CONFIG_UPDATE request and propagate it into the resulting CONFIG envelope for downstream consumers.

### DIFF
--- a/common/policy/policy.go
+++ b/common/policy/policy.go
@@ -113,10 +113,43 @@ func (cp *DefaultConfigUpdateProposer) AuthorizeAndVerifyConfigUpdate(envelope *
 }
 
 func BuildVerifiedConfigRequest(request *protos.Request, configEnvelope *cb.ConfigEnvelope, bundle channelconfig.Resources, signer identity.SignerSerializer, verifier *requestfilter.RulesVerifier) (*protos.Request, error) {
+	// Extract tx_id from the original CONFIG_UPDATE request so we can
+	// propagate it into the resulting CONFIG envelope.  Downstream
+	// consumers (e.g. the committer sidecar) use the tx_id to
+	// correlate config blocks with the original submission.
+	var originalTxID string
+	if request != nil && len(request.Payload) > 0 {
+		payload, err := protoutil.UnmarshalPayload(request.Payload)
+		if err == nil && payload.Header != nil {
+			chdr, err := protoutil.UnmarshalChannelHeader(payload.Header.ChannelHeader)
+			if err == nil && chdr != nil {
+				originalTxID = chdr.TxId
+			}
+		}
+	}
+
 	// Wrap the config envelope and sign, prepare a matching request
 	config, err := protoutil.CreateSignedEnvelope(cb.HeaderType_CONFIG, bundle.ConfigtxValidator().ChannelID(), signer, configEnvelope, int32(0), 0)
 	if err != nil {
 		return nil, fmt.Errorf("error creating a signed envelope, err: %s", err)
+	}
+
+	// Propagate the original tx_id into the new CONFIG envelope.
+	if originalTxID != "" {
+		payload, err := protoutil.UnmarshalPayload(config.Payload)
+		if err == nil && payload.Header != nil {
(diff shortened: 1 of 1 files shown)


tock-ibm (changes requested):
@elishambadi Thanks for your contribution!

First, a procedural request: when submitting a PR, always accompany it with an issue that explains what is the problem this PR solves. It is not enough to have it in a comment in the code.

Now to the PR itself. If understand your intent correctly, you want to be able to extract the TX-ID of the CONFIG_UPDATE transaction from the config block that pops out of the Assemblers via the Delivery service. Note that this info already exists, as the `ConfigEnvelope`, which is stored in the config block, contains both the full `Config` and the `LastUpdate`:

```go
type ConfigEnvelope struct {
	Config     *Config   `protobuf:"bytes,1,opt,name=config,proto3" json:"config,omitempty"`                           // A marshaled Config structure
	LastUpdate *Envelope `protobuf:"bytes,2,opt,name=last_update,json=lastUpdate,proto3" json:"last_update,omitempty"` // The last CONFIG_UPDATE message which generated this current configuration
}
```

This is exactly like Fabric:

https://github.com/hyperledger/fabric/blob/13792d5c2d38f9c6d3b1d86692e7aa20359b572b/orderer/common/multichannel/chainsupport.go#L133

In Fabric-X it is here:

https://github.com/hyperledger/fabric-x-orderer/blob/8ec46debd753f10dc824d891f087b65f34d1106c/common/policy/policy.go#L99

The `LastUpdate` should have the TX-ID you are looking for.

Therefore, it seems to me that this PR is not needed. 
What do you think?

elishambadi:
Thanks for the response. I'll be sure to create an issue.

Meanwhile this issue, is caused by the verifier downstream rejecting envelopes without TX_ID on top-level of envelope in `mapMessage`.

Take a look here: https://github.com/hyperledger/fabric-x-committer/blob/260027471523dc2383127b9e47fe1b1356038f9a/service/sidecar/mapping.go#L115

The TxID exists but not where the verifier requires it.

I think this change can be here or in protos.CreateSignedEnvelope().

tock-ibm:
> Thanks for the response. I'll be sure to create an issue.
> 
> Meanwhile this issue, is caused by the verifier downstream rejecting envelopes without TX_ID on top-level of envelope in `mapMessage`.
> 
> Take a look here: https://github.com/hyperledger/fabric-x-committer/blob/260027471523dc2383127b9e47fe1b1356038f9a/service/sidecar/mapping.go#L115
> 
> The TxID exists but not where the verifier requires it.
> 
> I think this change can be here or in protos.CreateSignedEnvelope().

@elishambadi  Can you post here what are the errors you are getting from the downstream verifier?

elishambadi:
@tock-ibm My bad. I meant sidecar as the `mapping.go` file path shows, not the verifier. Find debug logs from sidecar below on submitting a config update:

```
2026-08-06 13:06:53.085 UTC [msp.identity] Verify -> DEBU Verify: signer identity (certificate subject=CN=admin,OU=client issuer=CN=ca,OU=Fabric,O=Hyperledger,L=,C=US serialnumber=384686640402911678117279081645295933254170743553)
2026-08-06 13:06:53.088 UTC [policies] SignatureSetToValidIdentities -> DEBU signature for identity 0 validated
2026-08-06 13:06:53.090 UTC [sidecar] debugTx -> DEBU ID [<no-id>]: excluded: MALFORMED_MISSING_TX_ID (no TX ID) #missing tx id here
2026-08-06 13:06:53.092 UTC [msp] DeserializeIdentity -> DEBU Obtaining identity
2026-08-06 13:06:53.092 UTC [msp.identity] newIdentity -> DEBU Creating identity instance for cert -----BEGIN CERTIFICATE-----
```

The error is indicated.

tock-ibm:
@elishambadi See an issue I opened for this in the committer: https://github.com/hyperledger/fabric-x-committer/issues/752

elishambadi:
Thanks for all the guidance @tock-ibm !

I'll probably now move my contributions to the committer issue.

Issue #1079, opened 2026-08-04
Assembler: `BatchHeap.Pop` leaks its index entry, causing panics in `PartitionPrefetchIndex`


DorKatzelnick:
## Summary

`batchHeap.Pop` truncates `data` but never deletes the entry from `batchToIndex`. Because `heap.Pop` first `Swap`s the item to the tail, which sets its index to `n-1`, a popped batch stays in the mapper pointing at an index that no longer exists.

`saveOrdinaryBatch` then discards the `Push` error . So when a batch previously removed via `heap.Pop` — i.e. by `evictOldestBatch`  or `Stop` — is re-delivered (if possible), `cache.Put` succeeds while `sequenceHeap.Push` silently fails, leaving a batch in the cache the heap does not know about.

## Tasks

- [ ] Delete the popped batch from `batchToIndex` in `batchHeap.Pop`.
- [ ] Stop ignoring the `sequenceHeap.Push` error in `saveOrdinaryBatch`: on failure `timer.Stop()` and roll back `cache.Put`.
- [ ] Nil-check `sequenceHeap.Pop()` in `evictOldestBatch`.
- [ ] Guard `heap.Remove` in `batchHeap.Remove` against an out-of-range index.
- [ ] Tests — `batch_heap_test.go`: assert `Has()` is false after `Pop()`, re-`Push` a popped item,      and `Remove` after a `Pop` (the untested triple this bug exploits).
- [ ] Tests — `partition_prefetch_index` regression test for `Put(a), Put(b), Put(a)` at      `maxSizeBytes` (evict → re-put → subsequent `Put`/`PopOrWait`), plus one asserting the TTL      timer is stopped by `Stop()`.

PR #1088, opened 2026-08-05
assembler BatchHeap Pop, Push

DorKatzelnick:
#1079

### DIFF
--- a/node/assembler/batch_heap.go
+++ b/node/assembler/batch_heap.go
@@ -70,6 +70,7 @@ func (h *batchHeap[T]) Pop() interface{} {
 	n := len(h.data)
 	lastItem := h.data[n-1]
 	h.data = h.data[:n-1]
+	_, _ = h.batchToIndex.Remove(lastItem.Batch)
 	return lastItem
 }
 
@@ -87,11 +88,14 @@ func (h *batchHeap[T]) Remove(batchId types.BatchID) (*BatchHeapItem[T], error) 
 	if err != nil {
 		return nil, ErrBatchDoesNotExist
 	}
-	item := heap.Remove(h, idx).(*BatchHeapItem[T])
-	_, err = h.batchToIndex.Remove(batchId)
-	if err != nil {
+	if idx < 0 || idx >= len(h.data) {
+		// A mapper entry pointing outside the heap means the mapper and the data went out of
+		// sync; heap.Remove would panic on it. Drop the stale entry so the batch can be pushed again.
+		_, _ = h.batchToIndex.Remove(batchId)
 		return nil, ErrBatchDoesNotExist
 	}
+	// heap.Remove ends up calling h.Pop, which removes the batch from the mapper.
+	item := heap.Remove(h, idx).(*BatchHeapItem[T])
 	return item, nil
 }
 
--- a/node/assembler/partition_prefetch_index.go
+++ b/node/assembler/partition_prefetch_index.go
@@ -284,9 +284,7 @@ func (pi *PartitionPrefetchIndex) saveOrdinaryBatch(batch types.Batch) error {
 	if err != nil {
 		return err
 	}
-	var timer StoppableTimer
-	pi.lastPutSeq = batch.Seq()
-	timer = pi.timerFactory.Create(pi.defaultTtl, func() {
+	timer := pi.timerFactory.Create(pi.defaultTtl, func() {
 		pi.logger.Infof("TTL handler executed for batch %s", BatchToString(batch))
 
 		pi.stateCond.L.Lock()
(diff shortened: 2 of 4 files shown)


node/assembler/batch_heap.go
```diff
@@ -70,6 +70,7 @@ func (h *batchHeap[T]) Pop() interface{} {
 	n := len(h.data)
 	lastItem := h.data[n-1]
 	h.data = h.data[:n-1]
+	_, _ = h.batchToIndex.Remove(lastItem.Batch)
```
tock-ibm, line 73: This is the core fix — good. Swallowing the mapper `Remove` error (`_, _ =`) is fine here since the entry is guaranteed present and `heap.Interface.Pop` can't propagate an error anyway; just flagging that a would-be stale entry would pass silently.

node/assembler/batch_heap.go
```diff
@@ -87,11 +88,14 @@ func (h *batchHeap[T]) Remove(batchId types.BatchID) (*BatchHeapItem[T], error)
 	if err != nil {
 		return nil, ErrBatchDoesNotExist
 	}
-	item := heap.Remove(h, idx).(*BatchHeapItem[T])
-	_, err = h.batchToIndex.Remove(batchId)
-	if err != nil {
+	if idx < 0 || idx >= len(h.data) {
```
tock-ibm, line 91: Defensive branch: now that `Pop` keeps `batchToIndex` in sync, a valid `idx` from `Get` that is out of range for `data` shouldn't occur in normal operation, so this is effectively a safety net. Fine to keep, but note it isn't exercised by the tests (the `RemovingAPoppedBatch` test hits the `Get`-error path above, not this branch). A one-word "defensive" note in the comment would make the intent clear.

node/assembler/partition_prefetch_index.go
```diff
@@ -300,7 +298,16 @@ func (pi *PartitionPrefetchIndex) saveOrdinaryBatch(batch types.Batch) error {
 
 		pi.metrics.updatePrefetchIndexSize(pi.partition.Shard, -batchSizeBytes(batch))
 	})
-	pi.sequenceHeap.Push(&BatchHeapItem[StoppableTimer]{Batch: batch, Value: timer})
+	if err := pi.sequenceHeap.Push(&BatchHeapItem[StoppableTimer]{Batch: batch, Value: timer}); err != nil {
```
tock-ibm, line 301: Nice — rolling back the cache insert and stopping the timer when `Push` fails keeps the cache and heap in sync. Note the `Push` rejection path (`ErrBatchAlreadyExists`) is now impossible in normal flow given the `Pop` fix, but handling it here is the right defensive choice.

node/assembler/partition_prefetch_index.go
```diff
@@ -367,6 +374,9 @@ func (bs *PartitionPrefetchIndex) PutForce(batch types.Batch) error {
 
 func (pi *PartitionPrefetchIndex) evictOldestBatch() error {
 	batchAndTimer := pi.sequenceHeap.Pop()
+	if batchAndTimer == nil {
```
tock-ibm, line 377: Good guard — converts the former nil-deref panic into an error. Unreachable in normal flow now (an empty heap implies an empty cache, so the eviction loop wouldn't run), but sound.

tock-ibm (commented):
## Review: assembler BatchHeap Pop, Push (#1079)

### Overview
Fixes the `sequenceHeap` / `cache` desync in the assembler prefetch index. The root cause was `batchHeap.Pop()` removing the item from `data` but not from `batchToIndex`; the stale mapper entry then (a) made `Has()` report a popped batch as present, so a re-delivered batch's `Push()` returned `ErrBatchAlreadyExists` and the batch stayed in the cache but never in the heap (unevictable), and (b) pointed out of bounds, so a later `Remove()` would panic inside `container/heap`. The fix is correct and minimal.

### Correctness — verified
- **`Pop` mapper cleanup**: both `heap.Pop(h)` and `heap.Remove(h, i)` funnel through our `Pop()`, and in both paths the target ends up at the last slot with its mapper entry updated by `Swap` before deletion — consistent in every path.
- **`Remove` reorder**: the old explicit `batchToIndex.Remove` after `heap.Remove` is now redundant (Pop handles it) and correctly dropped — no double-removal.
- **`lastPutSeq` moved after `Push`**: subtle but correct — it now advances only when the batch actually lands, matching the documented "incremented after insertion" contract, and correctly stays put on rollback.
- **Rollback path**: cache is popped and the timer stopped on `Push` failure; the metrics `+batchSize` update happens in `Put` only after `saveOrdinaryBatch` returns, so there is nothing to roll back there. Consistent.

### Test coverage — excellent
Strong suite: pop-clears-mapper, re-push after pop, re-push/re-pop cycles, remove-of-popped-returns-error, heap ordering after pop+remove, the full re-delivered-evicted-batch scenarios, and that `Stop` stops each remaining TTL timer exactly once. The `raiseMaxPoppedSeq` helper is well-documented and the timer-stop-count assertions nicely prove no timers leak.

### Risk
Low. Changes are confined to a single lock-protected data structure and its one consumer. No API/signature changes. The only behavioral change is `lastPutSeq` timing, which is an improvement toward the documented contract.

**Recommendation: looks good to merge.** A couple of minor, non-blocking notes inline.

---

tock-ibm (approved):

PR #1089, opened 2026-08-06
Wire cluster local config and update defaults

moradna:
#857

### DIFF
--- a/common/synchronizer/block_puller.go
+++ b/common/synchronizer/block_puller.go
@@ -68,19 +68,7 @@ func NewBlockPuller(
 		logger.Warnf("Failed parsing my own TLS certificate: %v, therefore we may connect to our own endpoint when pulling blocks", err)
 	}
 
-	// TODO Fabric defaults. Extend the config to have these values, and use the config values instead of hardcoding them here.
-	// Cluster: Cluster{
-	// 	ReplicationMaxRetries:          12,
-	// 	RPCTimeout:                     time.Second * 7,
-	// 	DialTimeout:                    time.Second * 5,
-	// 	ReplicationBufferSize:          20971520,
-	// 	SendBufferSize:                 100,
-	// 	ReplicationRetryTimeout:        time.Second * 5,
-	// 	ReplicationPullTimeout:         time.Second * 5,
-	// 	CertExpirationWarningThreshold: time.Hour * 24 * 7,
-	// 	ReplicationPolicy:              "consensus", // BFT default; on etcdraft it is ignored
-	// },
-
+	// TODO: RetryTimeout and FetchTimeout are hardcoded; add matching Cluster fields if needed.
 	bp := &comm.BlockPuller{
 		MyOwnTLSCert:        myCert,
 		VerifyBlockSequence: verifyBlockSequenceNoOp,
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -18,9 +18,9 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
 		LogSpec: "info",
 		Cluster: ClusterYaml{
 			SendBufferSize:                 100,
-			ReplicationPolicy:              "consensus",
+			ReplicationPolicy:              "",
 			RPCTimeout:                     time.Minute,
-			ReplicationBufferSize:          20971520,
+			ReplicationBufferSize:          50 * 1024 * 1024,
 			ReplicationMaxRetryInterval:    10 * time.Second,
 			ReplicationMinRetryInterval:    200 * time.Millisecond,
 			ReplicationMaxRetryDuration:    time.Minute,
--- a/config/defaults.go
+++ b/config/defaults.go
@@ -17,7 +17,7 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
 	GeneralConfig: &GeneralConfig{
 		LogSpec: "info",
 		Cluster: ClusterYaml{
-			SendBufferSize:                 100,
+			SendBufferSize:                 2000,
 			ReplicationPolicy:              "",
 			RPCTimeout:                     time.Minute,
 			ReplicationBufferSize:          50 * 1024 * 1024,
--- a/config/local_config.go
+++ b/config/local_config.go
@@ -618,5 +618,10 @@ func applyNodeDefaults(nodeLocalConfig *NodeLocalConfig, role string, logger *fl
 			nodeLocalConfig.AssemblerParams.BatchRequestsChannelSize = DefaultAssemblerParams.BatchRequestsChannelSize
 			logger.Infof("Assembler.BatchRequestsChannelSize is not set, using default value: %d", nodeLocalConfig.AssemblerParams.BatchRequestsChannelSize)
 		}
+
+		if nodeLocalConfig.GeneralConfig.Cluster.ReplicationPolicy == "" {
+			nodeLocalConfig.GeneralConfig.Cluster.ReplicationPolicy = "assemblerSync"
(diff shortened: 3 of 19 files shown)


node/consensus/consensus_builder.go
```diff
@@ -443,7 +437,7 @@ func setupComm(c *Consensus) {
 	commAuth := &comm.AuthCommMgr{
 		Logger:         c.Logger,
 		Signer:         c.Signer,
-		SendBufferSize: 2000,
+		SendBufferSize: c.fullConfig.LocalConfig.ClusterConfig.SendBufferSize,
 		NodeIdentity:   selfID,
 		Connections:    comm.NewConnectionMgr(c.clientConfig()),
 	}
```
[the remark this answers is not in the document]
moradna: @tock-ibm what do you think? It used to be hardcoded to `2000`, and now it uses the config default (`100`)

node/consensus/consensus_builder.go
```diff
@@ -443,7 +431,7 @@ func setupComm(c *Consensus) {
 	commAuth := &comm.AuthCommMgr{
 		Logger:         c.Logger,
 		Signer:         c.Signer,
-		SendBufferSize: 2000,
+		SendBufferSize: c.fullConfig.LocalConfig.ClusterConfig.SendBufferSize,
```
tock-ibm, line 434: **`SendBufferSize` default disagrees with prior behavior and with the generator.**

This changes the `AuthCommMgr` egress buffer from a hardcoded `2000` to `ClusterConfig.SendBufferSize`. The problem is the *default* for that field is **100** (`defaults.go:20`, applied at `local_config.go:462` when unset), while:
- the config generator uses **2000** (`DefaultSendBufferSize`, `local_config_gen.go:26`), and
- the test-sample consenter YAMLs set **2000**.

So generated deployments are unaffected, but any hand-written or default-relying consenter config now gets a 20x smaller consensus egress buffer (100 vs 2000). Also, this single knob now drives both this buffer (was 2000) and the synchronizer `SendBufferSize` (was hardcoded 100), so they can no longer be tuned independently and the shared default matches only the old synchronizer value.

Please confirm 100 is intended here, and consider aligning `defaults.go` with the generator (`DefaultSendBufferSize = 2000`) so the unset behavior matches what real nodes ship with.

config/local_config.go
```diff
@@ -623,5 +618,10 @@ func applyNodeDefaults(nodeLocalConfig *NodeLocalConfig, role string, logger *fl
 			nodeLocalConfig.AssemblerParams.BatchRequestsChannelSize = DefaultAssemblerParams.BatchRequestsChannelSize
 			logger.Infof("Assembler.BatchRequestsChannelSize is not set, using default value: %d", nodeLocalConfig.AssemblerParams.BatchRequestsChannelSize)
 		}
+
+		if nodeLocalConfig.GeneralConfig.Cluster.ReplicationPolicy == "" {
+			nodeLocalConfig.GeneralConfig.Cluster.ReplicationPolicy = "assemblerSync"
```
tock-ibm: `"assemblerSync"` is a magic string set here, matched by a literal in the assembler factory switch (`synchronizer_factory.go:82`), and repeated in `testutil/config_utils.go`. Same for `"consensus"`. Consider extracting named constants (e.g. `ReplicationPolicyAssemblerSync`, `ReplicationPolicyConsensus`) so the default-setter, dispatcher, and tests cannot drift apart.

node/assembler/assembler.go
```diff
@@ -294,7 +294,7 @@ func (a *Assembler) initLedger(configBlock *common.Block, nodeConfig *node_confi
 		synchronizer := synchronizerFactory.CreateSynchronizer(
 			a.logger,
 			uint64(nodeConfig.PartyId),
-			config.Cluster{SendBufferSize: 100, ClientCertificate: nodeConfig.TLSCertificateFile, ClientPrivateKey: nodeConfig.TLSPrivateKeyFile, ReplicationPolicy: "assemblerSync"},
+			*a.configuration.LocalConfig.ClusterConfig,
```
tock-ibm, line 297: This now dereferences `*a.configuration.LocalConfig.ClusterConfig`. In production this is populated via `ReadConfig`→`LoadLocalConfig`, and the test was correctly updated away from the empty `&Configuration{}` (which would panic here). Worth double-checking no other caller constructs an `Assembler` with a bare `Configuration`. Also note the client cert now comes from `ClusterConfig.ClientCertificate/PrivateKey` (populated by the `General.TLS` fallback at `local_config.go:502-507`) rather than `nodeConfig.TLSCertificateFile` — equivalent, but an implicit dependency on that default step having run.

config/defaults.go
```diff
@@ -18,9 +18,9 @@ var DefaultNodeLocalConfig = NodeLocalConfig{
 		LogSpec: "info",
 		Cluster: ClusterYaml{
 			SendBufferSize:                 100,
-			ReplicationPolicy:              "consensus",
+			ReplicationPolicy:              "",
 			RPCTimeout:                     time.Minute,
-			ReplicationBufferSize:          20971520,
+			ReplicationBufferSize:          50 * 1024 * 1024,
```
tock-ibm, line 23: Buffer-size bump to 50 MiB looks fine and the samples/generated consenter YAMLs were updated to match (52428800). Nit: consider sharing this with the sample generation via a named constant so the YAML literal and this default stay in sync.

tock-ibm (commented):
## Code Review

Good, well-scoped refactor: it removes duplicated hardcoded `Cluster` structs at the consenter and assembler call sites and threads real values from `LocalConfig.ClusterConfig`. The policy wiring is coherent — `""` is accepted by the consensus synchronizer factory (`case "consensus", "":`) and `"assemblerSync"` by the assembler factory. Tests were updated appropriately (notably the assembler test now passes a real `ConfigurationWithDefaultCluster()` instead of an empty `&Configuration{}`, which the new dereference requires).

One thing I'd like to resolve before merge is the `SendBufferSize` default mismatch (inline comment). The rest are minor.

_Posted by Claude Code on behalf of @Yoav Tock_

common/synchronizer/block_puller.go
```diff
@@ -68,19 +68,7 @@ func NewBlockPuller(
...
-	// 	DialTimeout:                    time.Second * 5,
-	// 	ReplicationBufferSize:          20971520,
-	// 	SendBufferSize:                 100,
-	// 	ReplicationRetryTimeout:        time.Second * 5,
-	// 	ReplicationPullTimeout:         time.Second * 5,
-	// 	CertExpirationWarningThreshold: time.Hour * 24 * 7,
-	// 	ReplicationPolicy:              "consensus", // BFT default; on etcdraft it is ignored
-	// },
-
+	// TODO: RetryTimeout and FetchTimeout are hardcoded; add matching Cluster fields if needed.
```
tock-ibm, line 71: Nice cleanup replacing the large commented-out defaults block with a concise TODO. Just confirming the intent: `RetryTimeout` and `FetchTimeout` remain hardcoded at 5s here, which the TODO now correctly acknowledges.

node/consensus/consensus_builder.go
tock-ibm: lets go with 2000

node/consensus/consensus_builder.go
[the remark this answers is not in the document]
tock-ibm: lets go with 2000

config/local_config.go
tock-ibm: True. Lets call the options "consensus" | "assembler".

tock-ibm (approved):

Issue #1090, opened 2026-08-06
Compute batch digest once in batcher


DorKatzelnick:
## Problem: Batcher computes the batch digest 2–3 times per batch

`BatchedRequests.Digest()` is a SHA-256 over every byte of the batch payload. The batcher computes
it repeatedly over the same bytes, for the same result.

**Primary — 2 times per batch**: once for the BAF when the batch is cut, then again inside the
ledger append path.

**Secondary — 3 times per batch**: 1. `verifyBatch` hashes the payload to check it against the digest the primary claimed in the block header.  2. The ledger append recomputes it. 3. The BAF is created with `requests.Digest()`, called inline as an argument right after the append.

## Proposed fix

Thread the already-computed digest through the batch-ledger append path, computing it only when the
caller supplies none; the digest value is unchanged, so there is no protocol or wire-format impact.

## Tasks

- [ ] Add an optional digest parameter to `BatchLedgerArray.Append`, `BatchLedgerPart.Append` and
      `NewFabricBatchFromRequests`, falling back to `BatchedRequests.Digest()` when it is `nil`.
- [ ] Update the `BatchLedger` interface in `node/batcher` and the generated mocks.
- [ ] Primary: pass the digest already computed for the BAF into the ledger append.
- [ ] Secondary: reuse the digest computed in `verifyBatch` for both the ledger append and the BAF.

PR #1093, opened 2026-08-08
Remove duplicate batch digest computations

DorKatzelnick:
Compute the batch digest once per batch in the batcher
Add and update unit tests
#1090

### DIFF
--- a/node/batcher/batcher_role.go
+++ b/node/batcher/batcher_role.go
@@ -83,7 +83,8 @@ type BAFCreator interface {
 }
 
 type BatchLedgerWriter interface {
-	Append(partyID types.PartyID, batchSeq types.BatchSequence, configSeq types.ConfigSequence, batchedRequests types.BatchedRequests, primarySignature []byte)
+	// Append adds a batch to the ledger. The `digest` must be the digest of `batchedRequests`.
+	Append(partyID types.PartyID, batchSeq types.BatchSequence, configSeq types.ConfigSequence, batchedRequests types.BatchedRequests, digest []byte, primarySignature []byte)
 }
 
 type BatchLedgeReader interface {
@@ -354,7 +355,8 @@ func (b *BatcherRole) runPrimary() {
 		// Once the BAF reached the consenters it is considered safe to remove the requests from the mem pool
 
 		appendStart := time.Now()
-		b.Ledger.Append(b.ID, b.seq, b.ConfigSequenceGetter.ConfigSequence(), currentBatch, baf.Signature())
+		// The digest was already computed for the BAF above. pass it so the ledger does not recompute it.
+		b.Ledger.Append(b.ID, b.seq, b.ConfigSequenceGetter.ConfigSequence(), currentBatch, digest, baf.Signature())
 		b.Metrics.batchLedgerAppendLatency.Observe(time.Since(appendStart).Seconds())
 		sendBAFDone := make(chan struct{})
 		ctx, sendBafCancel := context.WithCancel(b.stopCtx)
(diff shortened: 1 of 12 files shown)


node/batcher/batcher_role.go
```diff
@@ -83,7 +83,10 @@ type BAFCreator interface {
 }
 
 type BatchLedgerWriter interface {
-	Append(partyID types.PartyID, batchSeq types.BatchSequence, configSeq types.ConfigSequence, batchedRequests types.BatchedRequests, primarySignature []byte)
+	// Append adds a batch to the ledger. The `digest` is the digest of `batchedRequests`; when it is nil
```
HagarMeir: let's always pass a pre computed digest and not let the ledger compute it 
do not check if it is nil

DorKatzelnick: fixed

node/batcher/batcher_role.go
```diff
@@ -354,7 +355,8 @@ func (b *BatcherRole) runPrimary() {
 		// Once the BAF reached the consenters it is considered safe to remove the requests from the mem pool
 
 		appendStart := time.Now()
-		b.Ledger.Append(b.ID, b.seq, b.ConfigSequenceGetter.ConfigSequence(), currentBatch, baf.Signature())
+		// The digest was already computed for the BAF above. pass it so the ledger does not recompute it.
```
tock-ibm, line 358: Nit: comment reads `...for the BAF above. pass it...` — capitalize "Pass" after the period.

node/ledger/batch_ledger_part_test.go
```diff
@@ -68,6 +70,34 @@ func TestBatchLedgerPart(t *testing.T) {
 	require.NotNil(t, part.RetrieveBatchByNumber(0))
 }
 
+// TestBatchLedgerPart_AppendWithDigest verifies that Append persists the digest it was given rather than
+// recomputing one over the payload;.
```
tock-ibm, line 74: Nit: stray `;.` at the end — `...recomputing one over the payload;.` → `...over the payload.`

tock-ibm (commented):
## Review: Remove duplicate batch digest computations

**Overview** — The batch digest (SHA-256 over the length-prefixed requests) was computed multiple times per batch: once for the BAF, then again inside the ledger-append path via `NewFabricBatchFromRequests` → `batchedRequests.Digest()`. This PR threads the already-computed `digest` through the write chain (`BatchLedgerWriter.Append` → `BatchLedgerArray.Append` → `BatchLedgerPart.Append` → `NewFabricBatchFromRequests`) so it's computed once per batch. Nice hot-path win.

**Correctness — ✓ behavior preserved.** The invariant is that the passed `digest` equals `batchedRequests.Digest()` in every production path:
- Primary (`runPrimary`): `digest = currentBatch.Digest()` is computed once and reused for both `CreateBAF` and `Append`. ✓
- Secondary (`runSecondary`): `digest := batch.Digest()` is reused, and `verifyBatch` already asserts `slices.Equal(batch.Digest(), br.Digest())` *before* this point, so reuse is safe. ✓

All production call sites of the two changed signatures are updated — only `batch_ledger_part.go` calls `NewFabricBatchFromRequests`, and only the two batcher sites call `Ledger.Append`. Mocks/stubs (`FakeBatchLedger`, `noopLedger`, `naiveReplication`) all match. Note the stored `DataHash` is the custom `BatchedRequests.Digest()` rather than `protoutil.ComputeBlockDataHash` — but that was already true pre-PR, so no downstream semantics change.

**Tests — ✓ strong.** The new tests pass a deliberately *wrong* digest and assert it's persisted as-is and differs from the recomputed one — exactly the "store, don't recompute" property the refactor relies on.

**Suggestion (non-blocking):** Since the digest-must-match invariant is now a caller contract enforced only by convention/comments, a future caller could silently corrupt the ledger's `DataHash`. Consider a cheap debug-level consistency check in `Append`, or a comment pointing at `verifyBatch` as the guard. Not required — the two current callers are provably correct.

**Verdict: LGTM** — clean, well-documented, well-tested optimization on a genuine hot path. Two nits inline.

tock-ibm (approved):

node/batcher/batcher_role.go
HagarMeir: ledger will not recompute it anyway...

Issue #1091, opened 2026-08-06
Improve metrics for accurate ledger append measurements


tock-ibm:
In the batcher, the current measurement for batch ledger append includes the calculation of the digest and header hash.
Separate these measurements to separate metrics: 1) the hash calculations; and 2) ledger append separately.

PR #1097, opened 2026-08-11
separate batch ledger latency metrics

moradna:
#1091


node/ledger/batch_ledger_part.go
```diff
@@ -75,11 +80,15 @@ func (b *BatchLedgerPart) Append(seq types.BatchSequence, configSeq types.Config
 	block := NewFabricBatchFromRequests(b.shardID, b.primaryPartyID, seq, batchedRequests, digest, configSeq, b.prevHash, primarySignature)
 
 	// Note: We do this only because we reuse the Fabric ledger, we don't really need a hash chain here.
+	hashingStart := time.Now()
 	b.prevHash = protoutil.BlockHeaderHash(block.Header)
+	b.metrics.HeaderHashingLatency.Observe(time.Since(hashingStart).Seconds())
```
tock-ibm, line 85: Fragile metrics initialization / nil-deref risk. `Append` unconditionally dereferences `b.metrics.HeaderHashingLatency`/`b.metrics.AppendLatency`, but `NewBatchLedgerArray` accepts a `*BatchLedgerMetrics` whose histogram fields are still `nil` (in `configureBatcher` it receives `&BatchLedgerMetrics{}` and is only populated later, in-place, by `NewBatcherMetrics` via the shared pointer). This only works because of hidden temporal coupling: the ledger is constructed first, then the batcher metrics constructor mutates the same pointer, and no `Append` happens in between. Any future caller that creates a `BatchLedgerArray` and appends without first routing the same pointer through `NewBatcherMetrics` (or that reorders these two calls) gets a nil-pointer panic on the hot path. Consider having the ledger own/initialize its metrics in its own constructor.

node/ledger/batch_ledger_metrics.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+)
+
+type BatchLedgerMetrics struct {
+	HeaderHashingLatency metrics.Histogram
+	AppendLatency        metrics.Histogram
+	logger               *flogging.FabricLogger
+}
+
+func (bl *BatchLedgerMetrics) NewBatchLedgerMetrics(p metrics.Provider, partyID, shardID string, logger *flogging.FabricLogger) {
```
tock-ibm: Un-idiomatic constructor. `NewBatchLedgerMetrics` is a *method* on `*BatchLedgerMetrics` that returns nothing and mutates the receiver, forcing the awkward two-step `m := &BatchLedgerMetrics{}; m.NewBatchLedgerMetrics(...)` at every call site. The `New...` prefix conventionally denotes a constructor that returns a new value. Prefer a plain `func NewBatchLedgerMetrics(p, party, shard, logger) *BatchLedgerMetrics`, which also lets the ledger initialize its own metrics and removes the nil-field window flagged in `batch_ledger_part.go`.

node/ledger/batch_ledger_metrics.go
```diff
@@ -0,0 +1,42 @@
...
+		Help:       "The latency to append a batch to the ledger.",
+		LabelNames: []string{"party_id", "shard_id"},
+		Buckets:    []float64{.0001, .001, .002, .003, .004, .005, .01, .03, .05, .1, .3, .5, 1}, // TODO: adjust buckets after reviewing Grafana
+	}
+)
+
+type BatchLedgerMetrics struct {
+	HeaderHashingLatency metrics.Histogram
+	AppendLatency        metrics.Histogram
+	logger               *flogging.FabricLogger
```
tock-ibm: Dead field: `logger` is stored (`bl.logger = logger`) but never read anywhere in the package. Drop the field and the `logger` parameter unless a use is intended.

node/ledger/batch_ledger_array.go
```diff
@@ -98,6 +100,10 @@ func (bla *BatchLedgerArray) ChannelID() string {
 	return bla.channelID
 }
 
+func (bla *BatchLedgerArray) Metrics() *BatchLedgerMetrics {
```
tock-ibm: Dead code: the new `Metrics()` accessor on `BatchLedgerArray` is never called. The batcher path keeps a separate `ledgerMetrics` local and passes it to both `NewBatchLedgerArray` and `NewBatcherMetrics` rather than going through this accessor (unlike the `AssemblerLedger.Metrics()` pattern it appears to mirror). Remove it, or use it in `configureBatcher` for consistency.

docs/monitoring/metrics.md
```diff
@@ -109,7 +109,15 @@ It queries Prometheus for metric data and displays it as dashboards, graphs, and
   **Help:** "The latency from receiving a batch on the secondary until it is verified."
 
 - **Namespace:** "batcher"  
-  **Name:** "batch_ledger_append_latency_seconds"  
+  **Name:** "batch_hashing_latency_seconds"
```
tock-ibm: Markdown formatting inconsistency: this `**Name:**` line lacks the trailing two-space hard break that the other two new entries (lines 116, 120) have, so the `Name` and `Help` lines for `batch_hashing_latency_seconds` render on one line.

```suggestion
  **Name:** "batch_hashing_latency_seconds"  
```

node/ledger/batch_ledger_part.go
moradna: It’s initialized as part of the batcher builder, before the role starts and any appends happen. So the current flow is safe.

node/ledger/batch_ledger_metrics.go
moradna: good point, removed it here and from `AssemblerLedgerMetrics` as well

node/batcher/batcher_role.go
DorKatzelnick: observe `br.Digest()` calculation time in  batchHashingLatency.

save it first: `computed := br.Digest()`, and then compare `batch.Digest()` with `computed` so the hash is calculated just once.

DorKatzelnick (approved):

tock-ibm (approved):

PR #1103, opened 2026-08-11
Fix data race on c.BFT.Synchronizer during dynamic reconfiguration

### LINKED ISSUE #1101: Data race on c.BFT.Synchronizer during dynamic reconfiguration
## Data race on `c.BFT.Synchronizer` during dynamic reconfiguration

Raised in review of #959 (see https://github.com/hyperledger/fabric-x-orderer/pull/959#discussion_r3751292687).

### The race
On the dynamic reconfiguration path (`configureConsensus` called with `createNewBFT=false`, from `stopAndReconfigure`), the existing BFT instance keeps running, and `configureConsensus` reassigns `c.BFT.Synchronizer` (and arma's own `c.Synchronizer`) in `node/consensus/consensus_builder.go`.

SmartBFT reads that field **without a lock** in `pkg/consensus.Sync()` (`c.Synchronizer.Sync()`), and the viewchanger goroutine can drive a sync independently of the controller goroutine that is parked in `Deliver`. So the unlocked read can race the reconfiguration goroutine's write.

`SoftStop`'s `Synchronizer.Stop()` stops the old synchronizer's goroutines but does not stop SmartBFT from *calling* `Sync()` again, so it does not close the window. This is a genuine data race by the Go memory model and would be flagged by `-race`.

### Recommended fix — a `synchronizerHolder`
The `bft.Synchronizer` interface is just `Sync() bft.SyncResponse`, so we can insert a stable indirection that SmartBFT holds once and never sees reassigned:

```go
type synchronizerHolder struct {
    mu    sync.RWMutex
    inner SynchronizerWithStop
}
func (h *synchronizerHolder) Sync() smartbft_types.SyncResponse {
    h.mu.RLock(); s := h.inner; h.mu.RUnlock()
    return s.Sync()
}
func (h *synchronizerHolder) Stop() {
    h.mu.RLock(); s := h.inner; h.mu.RUnlock()
    s.Stop()
}
func (h *synchronizerHolder) swap(s SynchronizerWithStop) {
    h.mu.Lock(); h.inner = s; h.mu.Unlock()
}
```

- `c.BFT.Synchronizer` (and arma's `c.Synchronizer`) is set to the holder **once**, at BFT creation, and never reassigned.
- On the dynamic path, `configureConsensus` calls `holder.swap(newSynchronizer)` instead of reassigning the field.
- Reads (`Sync()`/`Stop()`) and the swap are serialized by the holder's own `RWMutex`, so there's no unsynchronized access — and no change to vendored SmartBFT.

### Acceptance
- [ ] `c.BFT.Synchronizer` is never reassigned after BFT creation
- [ ] Dynamic reconfiguration swaps the inner synchronizer under a lock
- [ ] Covered by a test run under `-race`

HagarMeir:
## What

Fixes the data race on `c.BFT.Synchronizer` during dynamic reconfiguration, described in #1101 (raised in review of #959).

### The race

On the dynamic reconfiguration path (`configureConsensus` with `createNewBFT=false`, from `stopAndReconfigure`), the still-running BFT instance had its `Synchronizer` field reassigned. SmartBFT reads that field **without a lock** in `pkg/consensus.Sync()`, and the viewchanger goroutine can drive a sync independently of the reconfiguration goroutine's write — a genuine data race under the Go memory model, flagged by `-race`.

### The fix

Introduce a `synchronizer.Holder`: a stable indirection over `SynchronizerWithStop` guarded by an `RWMutex`.

- SmartBFT (`c.BFT.Synchronizer`) and arma's `c.Synchronizer` hold the `Holder` **once**, at BFT creation, and never see it reassigned.
- The dynamic path calls `holder.Swap(newSynchronizer)` instead of reassigning the field.
- `Sync()`/`Stop()` reads and the swap are serialized by the holder's own lock.
- No change to vendored SmartBFT.

## Acceptance

- [x] `c.BFT.Synchronizer` is never reassigned after BFT creation
- [x] Dynamic reconfiguration swaps the inner synchronizer under a lock
- [x] Covered by a test run under `-race`

## Testing

- New `TestHolderSwapDoesNotRaceSync` / `TestHolderSwapReplacesInner` — pass under `-race`.
- `TestConsensusWithRealConfigUpdate` (full dynamic-reconfig e2e) — passes under `-race`.
- `make basic-checks` — clean.

Fixes #1101

### DIFF
--- a/node/consensus/consensus_builder.go
+++ b/node/consensus/consensus_builder.go
@@ -167,8 +167,18 @@ func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConf
 	)
 	c.Logger.Info("Created a BFT Synchronizer")
 
-	c.BFT.Synchronizer = bftSynch
-	c.Synchronizer = bftSynch
+	if createNewBFT {
+		// Initial creation: build the holder once and hand it to SmartBFT (and keep it as
+		// c.Synchronizer). It is never reassigned afterwards.
+		c.synchronizerHolder = bft_synch.NewHolder(bftSynch)
+		c.BFT.Synchronizer = c.synchronizerHolder
+		c.Synchronizer = c.synchronizerHolder
+	} else {
+		// Dynamic reconfig: the existing BFT keeps running and still holds the same holder, so we
+		// only swap the inner synchronizer under the holder's lock rather than reassigning the
+		// BFT.Synchronizer field (which SmartBFT reads without a lock).
+		c.synchronizerHolder.Swap(bftSynch)
+	}
 }
 
 func (c *Consensus) InitOperationSystem() {
--- a/node/consensus/synchronizer/synchronizer_holder.go
+++ b/node/consensus/synchronizer/synchronizer_holder.go
@@ -1,0 +1,57 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package synchronizer
+
+import (
+	"sync"
+
+	smartbft_types "github.com/hyperledger-labs/SmartBFT/pkg/types"
+)
+
+// Holder is a stable indirection over a SynchronizerWithStop.
+//
+// SmartBFT reads its Synchronizer field without a lock in pkg/consensus.Sync(), and the
+// viewchanger goroutine can drive a Sync() independently of the controller goroutine. On the
+// dynamic reconfiguration path the inner synchronizer must be replaced while the BFT instance
+// keeps running, which would race that unlocked read if we reassigned the field directly.
+//
+// Instead, SmartBFT (and arma's own c.Synchronizer) holds the Holder once, at BFT creation, and
+// never sees it reassigned. Reconfiguration calls Swap to replace the inner synchronizer under
+// the Holder's own lock, so Sync()/Stop() reads and the swap are serialized.
+type Holder struct {
+	mu    sync.RWMutex
+	inner SynchronizerWithStop
+}
+
+// NewHolder returns a Holder wrapping the given inner synchronizer.
+func NewHolder(inner SynchronizerWithStop) *Holder {
+	return &Holder{inner: inner}
+}
+
+func (h *Holder) Sync() smartbft_types.SyncResponse {
+	h.mu.RLock()
+	s := h.inner
+	h.mu.RUnlock()
+	return s.Sync()
+}
+
+func (h *Holder) Stop() {
+	h.mu.RLock()
+	s := h.inner
+	h.mu.RUnlock()
+	s.Stop()
+}
+
+// Swap replaces the inner synchronizer under the Holder's lock. It is safe to call concurrently
+// with Sync()/Stop().
+func (h *Holder) Swap(inner SynchronizerWithStop) {
+	h.mu.Lock()
+	h.inner = inner
+	h.mu.Unlock()
+}
+
+var _ SynchronizerWithStop = (*Holder)(nil)
--- a/node/consensus/consensus.go
+++ b/node/consensus/consensus.go
@@ -123,7 +123,12 @@ type Consensus struct {
 	ReconfigAbort chan struct{}
 
 	synchronizerFactory bft_synch.SynchronizerFactory // Builds a BFT synchronizer
-	Synchronizer        SynchronizerStopper           // The BFT synchronizer built by the factory
+	// synchronizerHolder is the stable indirection handed to SmartBFT (as BFT.Synchronizer) and
+	// held as c.Synchronizer. It is created once at BFT creation and never reassigned; the dynamic
+	// reconfiguration path swaps its inner synchronizer under the holder's lock instead. This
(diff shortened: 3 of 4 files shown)


node/consensus/consensus_builder.go
```diff
@@ -167,8 +167,18 @@ func (c *Consensus) configureConsensus(nodeConfig *node_config.ConsenterNodeConf
...
+		// Initial creation: build the holder once and hand it to SmartBFT (and keep it as
+		// c.Synchronizer). It is never reassigned afterwards.
+		c.synchronizerHolder = bft_synch.NewHolder(bftSynch)
+		c.BFT.Synchronizer = c.synchronizerHolder
+		c.Synchronizer = c.synchronizerHolder
+	} else {
+		// Dynamic reconfig: the existing BFT keeps running and still holds the same holder, so we
+		// only swap the inner synchronizer under the holder's lock rather than reassigning the
+		// BFT.Synchronizer field (which SmartBFT reads without a lock).
+		c.synchronizerHolder.Swap(bftSynch)
```
tock-ibm, line 180: **Low severity — unguarded `Swap` relies on an unenforced ordering invariant.**

`c.synchronizerHolder` is assigned only in the `createNewBFT == true` branch. If `configureConsensus` is ever invoked with `createNewBFT == false` before an initial `createNewBFT == true` run (a future refactor, or a test constructing `Consensus` directly and driving `stopAndReconfigure`), `c.synchronizerHolder` is nil and this `Swap()` panics.

Currently unreachable because `Start` always runs the `createNewBFT == true` path first, but nothing enforces or documents that precondition at this call site. A nil guard (or a comment stating the invariant) would make the dependency explicit.

node/consensus/synchronizer/synchronizer_holder.go
```diff
@@ -0,0 +1,57 @@
...
+// SmartBFT reads its Synchronizer field without a lock in pkg/consensus.Sync(), and the
+// viewchanger goroutine can drive a Sync() independently of the controller goroutine. On the
+// dynamic reconfiguration path the inner synchronizer must be replaced while the BFT instance
+// keeps running, which would race that unlocked read if we reassigned the field directly.
+//
+// Instead, SmartBFT (and arma's own c.Synchronizer) holds the Holder once, at BFT creation, and
+// never sees it reassigned. Reconfiguration calls Swap to replace the inner synchronizer under
+// the Holder's own lock, so Sync()/Stop() reads and the swap are serialized.
+type Holder struct {
+	mu    sync.RWMutex
```
tock-ibm, line 26: **Low severity — `sync.RWMutex` reimplements what `sync/atomic` provides more simply.**

`Sync()`/`Stop()` take `RLock`, copy `inner`, `RUnlock`, then call the copy; `Swap` takes the write lock to store. Because the lock is released *before* the actual call, the mutex provides nothing beyond an atomic load/store of the interface value.

An `atomic.Pointer[SynchronizerWithStop]` (or `atomic.Value`) would be a lock-free, shorter equivalent with identical semantics. Not a correctness issue — just less machinery to maintain.

node/consensus/synchronizer/synchronizer_holder.go
```diff
@@ -0,0 +1,57 @@
...
+func (h *Holder) Stop() {
+	h.mu.RLock()
+	s := h.inner
+	h.mu.RUnlock()
+	s.Stop()
+}
+
+// Swap replaces the inner synchronizer under the Holder's lock. It is safe to call concurrently
+// with Sync()/Stop().
+func (h *Holder) Swap(inner SynchronizerWithStop) {
```
tock-ibm, line 51: **Low severity — `Swap` silently orphans the previous inner without stopping it.**

`Swap` replaces `h.inner` and discards the old synchronizer without calling its `Stop()`. Correctness depends entirely on the reconfig path having already called `SoftStop` (which stops the current inner) before `Swap`.

The `Holder` API neither stops the old inner nor documents this requirement, so any future use of `Swap` outside the `SoftStop`-first sequence would leak the old synchronizer's `Sync()` goroutines / BFT deliverer. Worth a doc comment on `Swap` stating the caller must stop the previous inner first.

tock-ibm (commented):
Traced the whole PR. The fix is sound and complete: SmartBFT's only unlocked read of the field is `pkg/consensus.Sync()`, controller/viewchanger delegate through the top-level `Consensus.Sync()` rather than caching the raw field, and the reconfig flow stops the old inner (`SoftStop`) before swapping, so no leak in normal operation. Clean, well-reasoned change.

Three low-severity observations left inline — all optional/defensive, none blocking.

_Posted by Claude Code on behalf of @Yoav Tock._

node/consensus/consensus_builder.go
tock-ibm: Add documentation at the call to Swap()

node/consensus/synchronizer/synchronizer_holder.go
tock-ibm: Add documentation at the call to Swap()

node/consensus/synchronizer/synchronizer_holder.go
tock-ibm: We will live with that

tock-ibm (approved):

PR #1107, opened 2026-08-12
Invalidate authorized cluster streams on node identity change

### LINKED ISSUE #1102: ConfigureNodeCerts does not invalidate authorized streams on node identity change
## `ConfigureNodeCerts` does not invalidate authorized streams when a node's identity changes

Raised in review of #959 (see https://github.com/hyperledger/fabric-x-orderer/pull/959#discussion_r3757089036).

### Behavior today
In `node/comm/clusterservice.go`, an incoming cluster stream is authenticated **once** at establishment (`VerifyAuthRequest` checks `Membership.MemberMapping[FromId]` identity + signature) and then recorded in `AuthorizedStreams` (streamID → nodeID). Subsequent messages on that stream are accepted based only on `AuthorizedStreams.Load(streamID)` — the identity is not re-verified per message.

`ConfigureNodeCerts(newNodes)` rebuilds `MemberMapping` and then prunes authorized streams:

```go
c.Membership.AuthorizedStreams.Range(func(streamID, nodeID interface{}) bool {
    if _, exists := c.Membership.MemberMapping[nodeID.(uint64)]; !exists {
        c.Membership.AuthorizedStreams.Delete(streamID.(uint64))
    }
    return true
})
```

The prune only removes streams whose **node ID is no longer present**. It does **not** remove streams whose **identity (cert) changed** for a still-present ID.

### The gap
If an existing node's ID (`uint64`) keeps the same value but its `Identity` (cert) is rotated via a config update, any stream that was already authorized under the **old** cert survives the reconfiguration and continues to be accepted, even though it authenticated with a now-invalid identity. A new stream would be forced to authenticate against the new cert, but the pre-existing one is never invalidated.

### Question to resolve / acceptance
- [ ] Confirm the above (a cert change for an existing ID leaves its already-authorized incoming stream valid).
- [ ] If confirmed, `ConfigureNodeCerts` should also drop authorized streams whose stored nodeID's identity changed (not just IDs that were removed), forcing re-authentication against the new cert.
- [ ] Add a test covering cert rotation for an existing ID, asserting the old stream is de-authorized.

HagarMeir:
## Problem

Fixes #1102.

In `node/comm/clusterservice.go`, an incoming cluster stream is authenticated once at establishment (`VerifyAuthRequest` checks the sender's identity + signature) and then recorded in `AuthorizedStreams` as `streamID → nodeID`. Subsequent messages are accepted based only on stream authorization, without re-verifying identity per message.

`ConfigureNodeCerts` rebuilds `MemberMapping` and prunes authorized streams, but only removed streams whose **node ID was no longer present**. It did **not** remove streams whose **identity (cert) changed** for a still-present ID. As a result, if an existing node kept its `uint64` ID but rotated its cert via a config update, any stream already authorized under the **old** cert survived reconfiguration and kept being accepted — even though it authenticated with a now-invalid identity.

## Fix

- Record the identity each stream authenticated under. `AuthorizedStreams` now maps `streamID → authorizedStream{nodeID, identity}`.
- `VerifyAuthRequest` now also returns the validated `fromIdentity`, so the stored identity is the exact value that was verified (single source of truth, no second map lookup).
- The prune in `ConfigureNodeCerts` now de-authorizes a stream if its node ID was removed **or** its stored identity no longer matches the current `MemberMapping` entry — forcing re-authentication against the rotated cert.

## Tests

- `TestConfigureNodeCerts_PruneOnIdentityChange` — table-driven: unchanged identity keeps the stream authorized; a rotated identity for the same ID de-authorizes it (the core acceptance criterion).
- `TestConfigureNodeCerts_PruneOnRemovedID` — the pre-existing removed-ID pruning still holds.

Verified locally: `go test -race ./node/comm/`, `go build ./...`, `go vet`, and `gofmt` all clean.

### DIFF
--- a/node/comm/clusterservice.go
+++ b/node/comm/clusterservice.go
@@ -37,8 +37,16 @@ type ClusterStepStream interface {
 
 type MembersConfig struct {
 	MemberMapping     map[uint64][]byte
-	AuthorizedStreams sync.Map // Stream ID --> node identifier
-	nextStreamID      uint64
+	AuthorizedStreams sync.Map // Stream ID --> authorizedStream
+	nextStreamID      atomic.Uint64
+}
+
+// authorizedStream records the node and the identity (cert) a stream
+// authenticated under, so a later identity rotation for that node can
+// invalidate the stream and force re-authentication.
+type authorizedStream struct {
+	nodeID   uint64
+	identity []byte
 }
 
 // ClusterService implements the server API for ClusterNodeService service
@@ -80,15 +88,18 @@ func (s *ClusterService) Step(stream orderer.ClusterNodeService_StepServer) erro
 	}
 
 	s.Lock.RLock()
-	authReq, err := s.VerifyAuthRequest(stream, request)
+	authReq, fromIdentity, err := s.VerifyAuthRequest(stream, request)
 	if err != nil {
 		s.Lock.RUnlock()
 		s.Logger.Warnf("service authentication of %s failed with error: %v", addr, err)
 		return status.Errorf(codes.Unauthenticated, "access denied")
 	}
 
-	streamID := atomic.AddUint64(&s.Membership.nextStreamID, 1)
-	s.Membership.AuthorizedStreams.Store(streamID, authReq.FromId)
+	streamID := s.Membership.nextStreamID.Add(1)
+	s.Membership.AuthorizedStreams.Store(streamID, authorizedStream{
+		nodeID:   authReq.FromId,
+		identity: fromIdentity,
+	})
 	s.Lock.RUnlock()
 
 	defer s.Logger.Debugf("Closing connection from %s(%s)", commonName, addr)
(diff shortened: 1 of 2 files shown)


node/comm/clusterservice.go
```diff
@@ -242,9 +253,14 @@ func (c *ClusterService) ConfigureNodeCerts(newNodes []*common.Consenter) error
...
 
-	// Iterate over existing streams and prune those that should not be there anymore
-	c.Membership.AuthorizedStreams.Range(func(streamID, nodeID interface{}) bool {
-		if _, exists := c.Membership.MemberMapping[nodeID.(uint64)]; !exists {
+	// Iterate over existing streams and prune those that should not be there
+	// anymore: either the node ID was removed, or its identity (cert) was
+	// rotated. In the latter case the stream authenticated under the old cert
+	// and must re-authenticate against the new one.
+	c.Membership.AuthorizedStreams.Range(func(streamID, stream any) bool {
+		as := stream.(authorizedStream)
```
tock-ibm: Unchecked type assertion `stream.(authorizedStream)` will `panic` if a value of any other type is ever stored in `AuthorizedStreams`. This runs inside `Range` while holding `c.Lock` (write), so a mistyped `Store` anywhere would take down `ConfigureNodeCerts` at runtime rather than failing safely. Low severity given all current `Store` sites are consistent, but a comma-ok guard would be more robust:

```suggestion
	c.Membership.AuthorizedStreams.Range(func(streamID, stream any) bool {
		as, ok := stream.(authorizedStream)
		if !ok {
			c.Membership.AuthorizedStreams.Delete(streamID.(uint64))
			return true
		}
		currentIdentity, exists := c.Membership.MemberMapping[as.nodeID]
		if !exists || !bytes.Equal(currentIdentity, as.identity) {
			c.Membership.AuthorizedStreams.Delete(streamID.(uint64))
		}
```

tock-ibm (approved):

Issue #505, opened 2026-01-07
Batcher apply config - unit test


HagarMeir:
- [x] To test config changes, a **status** should be added to the batcher (similar to the rest of the nodes) See #1032 

Unit test config changes:

- [x] Batching parameter change (e.g. BatchSize, BatchTimeouts parameter)

- [x] Endpoint of batcher

- [x] Certificate of batcher

- [x] CACerts and TLSCACerts

- [x] Add party - make sure the joined batcher can communicate with the primary

- [x] Remove party - make sure the removed batcher cannot communicate with the primary

- [ ] Validity of config changes (bad configurations should be rejected)

- [ ] Timing scenarios: Pulling batches from a primary and pulling configs from consensus are asynch!
        So a primary could verify a batch on conf-sqn K and send a BAF
        Secondaries may pull said batch and verify it with either K, K-1, or K+1;
        We should make sure no TXs are lost in this situation;

- [ ] Memory pool pruning - make sure memory pool is pruned as expected

- [x] RequestMaxBytes change - make sure the batcher rejects requests that exceed the maximal size

HagarMeir:
**Findings on "Validity of config changes (bad configurations should be rejected)"**

While scoping a batcher unit test for this item, I found that the batcher is not the component that validates config *changes* — so where a "bad config rejected" test belongs depends on the layer:

- **Validation happens before ordering, not in the batcher.** Config updates are validated by the **router** ([`config_submitter.go`](https://github.com/hyperledger/fabric-x-orderer/blob/main/node/router/config_submitter.go#L132-L138)) and the **consenter** ([`consensus.go`](https://github.com/hyperledger/fabric-x-orderer/blob/main/node/consensus/consensus.go#L1297-L1300)) via `verify.DefaultOrdererRules.ValidateNewConfig` + `ValidateTransition`, *before* the config block is ordered.
- **The batcher does not re-validate reconfig blocks.** `Batcher.ApplyConfig` → `Configuration.NewUpdatedConfigurationFromBlock` ([`config.go`](https://github.com/hyperledger/fabric-x-orderer/blob/main/config/config.go#L938-L968)) only rebuilds a bundle; there are no business-rule checks. Consequences for a bad block reaching the batcher via consensus:
  - **Unparseable block** → `ApplyConfig` errors → `processNewConfigBlock` calls `Panicf` (crash, not graceful rejection).
[...]

PR #1108, opened 2026-08-12
Add NodeStatus to batcher

### LINKED ISSUE #1032: Batcher: add status
Add status to batcher in order to better test and monitor re-config flows.

HagarMeir:
## What

Adds a lifecycle status to the batcher, mirroring the other node roles (assembler, router, consensus).

The batcher was the only role not using the shared `node/utils.NodeStatus` type: it tracked lifecycle with two ad-hoc booleans (`isStopped`, `isSoftStopped`) and `GetStatus()` returned a bespoke `string`. This made it impossible to observe the applied config sequence — the signal the batcher reconfiguration tests need.

## Changes

- Add a `status node_utils.NodeStatus` field, initialized to `StateInitializing` with the bootstrap config sequence in `CreateBatcher`.
- Transition through `StateRunning` (carrying the current config sequence, so a dynamic reconfig updates it), `StateSoftStopped`, `StatePendingAdmin` (when a config change requires an admin action), and `StateStopped`.
- `GetStatus()` now returns `node_utils.NodeStatus` instead of `string`.
- Update the reconfig test to the new API and add status assertions (`Running` after `Run`, `Stopped` after `Stop`, `SoftStopped` after `SoftStop`) to the batcher unit tests.

No new proto/gRPC/HTTP endpoint — `NodeStatus` is an in-memory field read via `GetStatus()`, exactly as the other roles do it.

## Testing

`make unit-tests-batcher` passes (with `-race`).

Fixes #1032

### DIFF
--- a/node/batcher/batcher.go
+++ b/node/batcher/batcher.go
@@ -87,8 +87,7 @@ type Batcher struct {
 	running                            sync.WaitGroup // maybe change the name, it is only for state replicator
 	stopChan                           chan struct{}
 	mainExitChan                       chan struct{}
-	isStopped                          bool
-	isSoftStopped                      bool
+	status                             node_utils.NodeStatus
 	configAcker                        configack.Sender
 
 	primaryLock sync.RWMutex
@@ -144,9 +143,6 @@ func (b *Batcher) Run() {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	b.isStopped = false
-	b.isSoftStopped = false
-
 	b.stopChan = make(chan struct{})
 	b.stateChan = make(chan *state.State, 1)
 
@@ -166,30 +162,27 @@ func (b *Batcher) Run() {
 	b.logger.Infof("Health check serving on URL: %s", operations.HealthCheckServiceURL(b.opsSystem, b.logger))
 	b.logger.Infof("Logging spec service serving on URL: %s", operations.LogSpecServiceURL(b.opsSystem, b.logger))
 	b.logger.Infof("Version info serving on URL: %s", operations.VersionInfoServiceURL(b.opsSystem, b.logger))
+
+	b.status.Set(node_utils.StateRunning, b.config.Bundle.ConfigtxValidator().Sequence())
 }
 
-func (b *Batcher) GetStatus() string {
+func (b *Batcher) GetStatus() node_utils.NodeStatus {
 	b.lock.Lock()
 	defer b.lock.Unlock()
-	if b.isSoftStopped && !b.isStopped {
-		return "Soft Stopped"
-	}
-	if b.isSoftStopped && b.isStopped {
-		return "Stopped"
-	}
-	return "Running"
+	return b.status
 }
 
 func (b *Batcher) Stop() {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if b.isStopped {
+	state := b.status.GetState()
+	if state == node_utils.StateStopped {
 		return
 	}
 
 	b.logger.Infof("Stopping batcher node")
-	if !b.isSoftStopped {
+	if state != node_utils.StateSoftStopped && state != node_utils.StatePendingAdmin {
 		close(b.stopChan)
 		b.controlEventBroadcaster.Stop()
 		b.batcher.Stop()
@@ -210,19 +203,19 @@ func (b *Batcher) Stop() {
 
 	close(b.mainExitChan)
 
-	b.isStopped = true
-	b.isSoftStopped = true
+	b.status.SetState(node_utils.StateStopped)
 }
 
 func (b *Batcher) SoftStop() {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if b.isSoftStopped || b.isStopped {
+	state := b.status.GetState()
+	if state == node_utils.StateSoftStopped || state == node_utils.StateStopped || state == node_utils.StatePendingAdmin {
 		return
 	}
 
-	b.isSoftStopped = true
+	b.status.SetState(node_utils.StateSoftStopped)
 
 	b.logger.Infof("Soft stopping batcher node")
 	close(b.stopChan)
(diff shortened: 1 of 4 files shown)


MayRosenbaum (approved):

node/batcher/batcher_reconfig_test.go
```diff
@@ -117,7 +117,7 @@ func TestBatcherReceivesConfigBlockFromConsensusAndApplyConfig_ChangeBatchTimeou
 	//// wait for the batcher to initialize
 	//for j := range parties {
 	//	require.Eventually(t, func() bool {
-	//		return batchers[j].GetStatus() == "Running"
+	//		return batchers[j].GetStatus().GetState() == node_utils.StateRunning
```
tock-ibm, line 120: **Coverage gap for the headline capability.** The PR message motivates this change with the batcher reconfiguration tests needing to observe the applied config sequence after reconfiguration, and `stopAndReconfigure`→`Run()` is what advances `ConfigSequenceNumber` to the new seq. But the only live reconfig assertion checks `StateSoftStopped` (line 99); the block here that would verify the batcher returns to `StateRunning` with the *new* sequence stays commented out. A regression that leaves the batcher stuck in `SoftStopped`, or that fails to advance `ConfigSequenceNumber` on reconfig, would pass CI. The new `batcher_test.go` assertions only cover the bootstrap sequence value `0`, never a post-reconfig increment. Consider enabling this assertion (which also closes the dead-code finding below).

node/batcher/batcher_reconfig_test.go
```diff
@@ -117,7 +117,7 @@ func TestBatcherReceivesConfigBlockFromConsensusAndApplyConfig_ChangeBatchTimeou
 	//// wait for the batcher to initialize
```
tock-ibm, line 117: **Dead commented-out test code.** This is a commented-out `require.Eventually` block, and the diff edits the disabled assertion (line 120) from the old string API to the new `GetState()` API — so it now carries maintenance cost for code that never runs and can silently rot. Per the simplicity guideline it should either be deleted or actually re-enabled (re-enabling would also close the coverage gap flagged above).

node/batcher/batcher.go
```diff
@@ -166,30 +162,27 @@ func (b *Batcher) Run() {
 	b.logger.Infof("Health check serving on URL: %s", operations.HealthCheckServiceURL(b.opsSystem, b.logger))
 	b.logger.Infof("Logging spec service serving on URL: %s", operations.LogSpecServiceURL(b.opsSystem, b.logger))
 	b.logger.Infof("Version info serving on URL: %s", operations.VersionInfoServiceURL(b.opsSystem, b.logger))
+
+	b.status.Set(node_utils.StateRunning, b.config.Bundle.ConfigtxValidator().Sequence())
```
tock-ibm: **Divergence from the sibling roles' status lifecycle.** Router/consensus/assembler set the config sequence once via `Set(StateInitializing, seq)` — including on reconfig (e.g. `consensus.go:1139`, `router.go:129`) — and then transition with a bare `SetState(StateRunning)`. The batcher instead only sets the sequence here inside `Run()` via `Set(StateRunning, b.config.Bundle...Sequence())`, and `stopAndReconfigure` goes `SoftStopped → Run` with no `Initializing` phase.

Consequence: during the whole reconfig window `GetStatus()` reports `StateSoftStopped` carrying a *stale* `ConfigSequenceNumber` (`SoftStop` only calls `SetState`, never updating the seq), so observers can't distinguish "soft-stopped, config not yet applied" from an initializing reconfig the way they can for the other three roles. Not a crash — but it undermines the observability goal and makes the four roles harder to reason about together.

node/batcher/batcher.go
```diff
@@ -166,30 +162,27 @@ func (b *Batcher) Run() {
...
-	return "Running"
+	return b.status
 }
 
 func (b *Batcher) Stop() {
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	if b.isStopped {
+	state := b.status.GetState()
```
tock-ibm: **Local `state` shadows the imported `state` package.** `batcher.go` imports `node/consensus/state` under its default name `state` (used elsewhere as `*state.State`, e.g. `b.stateChan`). This `state := b.status.GetState()` — and the identical one in `SoftStop()` at line 213 — shadows that package. It compiles only because neither function currently references the package; a future edit needing `state.State` inside these functions would fail confusingly. Suggest renaming to `st` / `curState` (also aligns with the "favor explicit over clever" guideline).

tock-ibm (commented):
Code review of the `NodeStatus` refactor. The change itself is clean — the state-machine transitions faithfully preserve the old `Stop`/`SoftStop` semantics (`PendingAdmin` is only reachable after `SoftStop`, so skipping the full-stop block in `Stop()` matches the old `!isSoftStopped` guard), and I found no hard correctness regression. The findings below are lower-severity: a test-coverage gap, an observability inconsistency vs. the sibling roles, and two cleanups.

node/batcher/batcher_reconfig_test.go
HagarMeir: will be handled later by working on the batcher reconfig unit tests https://github.com/hyperledger/fabric-x-orderer/issues/505

node/batcher/batcher_reconfig_test.go
HagarMeir: will be handled later by working on the batcher reconfig unit tests https://github.com/hyperledger/fabric-x-orderer/issues/505

node/batcher/batcher.go
HagarMeir: renamed

node/batcher/batcher.go
HagarMeir: Good catch. Fixed by mirroring the consensus pattern: `Run()` now transitions with a bare `SetState(StateRunning)`, and `stopAndReconfigure` sets `Set(StateInitializing, newConfigSeq)` under the lock before calling `Run()`. During the reconfig window `GetStatus()` now reports `initializing` with the new config sequence instead of `soft-stopped` with a stale one, consistent with router/consensus/assembler.

DorKatzelnick (approved):

Issue #1073, opened 2026-07-30
Certificate verification skips all chain verification when `ignoreExpiration` is true


DorKatzelnick:
Problems: 

1.  `config/verify/cert_validation.go:134-138` tolerates an expired certificate by swallowing the `Expired` error from `cert.Verify`:

```go
if _, err = cert.Verify(opts); err != nil {
	if validationRes, ok := err.(x509.CertificateInvalidError); !ok || (!ignoreExpiration || validationRes.Reason != x509.Expired) || time.Now().Before(cert.NotBefore) {
		return errors.Wrapf(err, "verifying certificate with serial number %d", cert.SerialNumber)
	}
}
```

x509.Verify rejects a failed validity window in isValid(leafCertificate, ...) and returns before it calls buildChains. On that branch the issuer signature, the path to a trusted root, and the EKU constraints are all still unchecked. Swallowing the error discards all three, so ignoreExpiration=true only guarantees "parses as X.509 and is past its NotBefore".

2. The not-yet-valid guard tests the leaf's NotBefore, but validationRes.Cert may be the CA. A not-yet-valid root CA is therefore silently accepted .

PR #1112, opened 2026-08-13
Fix certificate verification when ignoring expiration

moradna:
#1073

### DIFF
--- a/config/verify/cert_validation.go
+++ b/config/verify/cert_validation.go
@@ -126,15 +126,22 @@ func verifyCert(raw []byte, opts x509.VerifyOptions, ignoreExpiration bool) erro
 	if len(raw) == 0 {
 		return errors.New("certificate is empty")
 	}
+
 	cert, err := parseCertificateFromBytes(raw)
 	if err != nil {
 		return err
 	}
 
-	if _, err = cert.Verify(opts); err != nil {
-		if validationRes, ok := err.(x509.CertificateInvalidError); !ok || (!ignoreExpiration || validationRes.Reason != x509.Expired) || time.Now().Before(cert.NotBefore) {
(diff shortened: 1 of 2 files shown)


DorKatzelnick (approved):

DorKatzelnick:
`opts.CurrentTime = cert.NotBefore` is set unconditionally whenever `ignoreExpiration` is true — including for certificates that are valid right now. The whole chain then gets validated as of the leaf's issuance instant, so any issuer whose own `NotBefore` is later than the leaf's `NotBefore` fails.

Reproduced with a currently-valid leaf under a re-issued CA cert:

```
CA cert:  now-24h .. now+365d
leaf:     now-48h .. now+365d   (valid right now)

ignoreExpiration=false => <nil>
ignoreExpiration=true  => verifying certificate ...: x509: certificate has expired or is
                          not yet valid: current time <now-48h> is before <now-24h>
```

**Suggested fix:** instead of pinning verification to one arbitrary instant, answer the actual question — *was this chain simultaneously valid at some instant at or before now?* Start the candidate time at `time.Now()` and only ever move it backwards: on a failed `Verify`, `errors.As` into `x509.CertificateInvalidError` and lower the candidate to `invalid.Cert.NotAfter`, then retry. `invalid.Cert` names the offending certificate even when that is the CA rather than the leaf, so the chain's binding constraint is discovered through the verifier itself — no need to reimplement chain building, and it keeps working once the `Intermediates` TODOs are filled in. Each step strictly decreases the time, so the loop terminates; keep the explicit not-yet-valid guard against `time.Now()` so expiry is ignored but not-yet-validity never is. Signature, path-to-root and EKU checks run on every attempt, so the security fix in this PR is preserved.

Issue #1115, opened 2026-08-16
blkstorage: file-backed prune info, ErrPruned, and read guards


DorKatzelnick:
Add the durable state that pruning needs, without reclaiming anything yet.

Each ledger gets a `prune/` directory beside its block files, holding the prune
info: the first readable block number and the lowest block file still kept. It is
loaded at open and published for lock-free reads.

The info lives in a file rather than in the index database because the index is
disposable by contract — delete it and it is rebuilt by scanning the block files —
while the first readable block number cannot be recovered that way. Pruning removes
whole block files, so the bound may fall inside the oldest surviving file and nothing
on disk records where. Dropping the index of a pruned ledger would silently reset the
bound to 0, reporting pruned blocks as missing and starting the rebuild scan at a
block file that is no longer there.

Files are written once under an increasing sequence number, and the superseded one is
removed afterwards. Nothing is overwritten, so the highest sequence number is always a
complete file; a crash between the two steps leaves two behind and the next read
resolves them in favour of the higher one.

Reads keyed by block number below the bound return a typed `ErrPruned`, matchable with
`errors.Is`. `Height()` is unaffected. New `BlockStore.FirstAvailableBlockNumber()`.

The prune info is persisted before it is published, and removing blocks follows that:
a reader refused a block that could still be served is harmless, whereas one served a
block about to disappear is not. Serializing that sequence against itself is the
caller's responsibility, and lands with `PruneBefore`.

No file is deleted and nothing sets the prune info, so behaviour on existing ledgers is
unchanged. Unblocks the recovery fixes and `PruneBefore` itself.

PR #1126, opened 2026-08-17
blkstorage: prune manager, file-backed prune info, ErrPruned, and read guards

DorKatzelnick:
#1115

### DIFF
--- a/common/ledger/blkstorage/blockfile_mgr.go
+++ b/common/ledger/blkstorage/blockfile_mgr.go
@@ -44,6 +44,9 @@ type blockfileMgr struct {
 	currentFileWriter         *blockfileWriter
 	bcInfo                    atomic.Value
 	cache                     *cache
+
+	// pruner owns the prune marker (see prune.go).
+	pruner *pruneMgr
 }
 
 /*
@@ -140,6 +143,12 @@ func newBlockfileMgr(id string, conf *Conf, indexConfig *IndexConfig, indexStore
 	mgr.bootstrappingSnapshotInfo = bsi
 	mgr.currentFileWriter = currentFileWriter
 	mgr.blkfilesInfoCond = sync.NewCond(&sync.Mutex{})
+
+	// Construct the pruner before syncIndex: on a pruned ledger the block files no longer start at
+	// blockfile_000000, and the recovery paths need to know where they do start.
+	if mgr.pruner, err = newPruneMgr(indexStore); err != nil {
+		return nil, err
+	}
 
 	if err := mgr.syncIndex(); err != nil {
 		return nil, err
@@ -387,7 +396,7 @@ func (mgr *blockfileMgr) syncIndex() error {
 		// bootstrapping the ledger from a snapshot or the index is dropped/corrupted afterward
 		return errors.Errorf(
 			"cannot sync index with block files. blockstore is bootstrapped from a snapshot and first available block=[%d]",
-			mgr.firstPossibleBlockNumberInBlockFiles(),
+			mgr.firstBlockNumAfterSnapshotBootstrap(),
 		)
 	}
 
@@ -527,11 +536,8 @@ func (mgr *blockfileMgr) retrieveBlockByNumber(blockNum uint64) (*common.Block, 
 	if blockNum == math.MaxUint64 {
 		blockNum = mgr.getBlockchainInfo().Height - 1
 	}
-	if blockNum < mgr.firstPossibleBlockNumberInBlockFiles() {
-		return nil, errors.Errorf(
-			"cannot serve block [%d]. The ledger is bootstrapped from a snapshot. First available block = [%d]",
-			blockNum, mgr.firstPossibleBlockNumberInBlockFiles(),
-		)
+	if err := mgr.checkBlockAvailable(blockNum); err != nil {
+		return nil, err
 	}
 	loc, err := mgr.index.getBlockLocByBlockNum(blockNum)
 	if err != nil {
@@ -546,7 +552,7 @@ func (mgr *blockfileMgr) retrieveBlockByTxID(txID string) (*common.Block, error)
 	if err == errNilValue {
 		return nil, errors.Errorf(
 			"details for the TXID [%s] not available. Ledger bootstrapped from a snapshot. First available block = [%d]",
-			txID, mgr.firstPossibleBlockNumberInBlockFiles(),
+			txID, mgr.firstAvailableBlockNum(),
 		)
 	}
 	if err != nil {
@@ -561,7 +567,7 @@ func (mgr *blockfileMgr) retrieveTxValidationCodeByTxID(txID string) (peer.TxVal
 	if err == errNilValue {
 		return peer.TxValidationCode(-1), 0, errors.Errorf(
 			"details for the TXID [%s] not available. Ledger bootstrapped from a snapshot. First available block = [%d]",
-			txID, mgr.firstPossibleBlockNumberInBlockFiles(),
+			txID, mgr.firstAvailableBlockNum(),
 		)
 	}
 	return validationCode, blkNum, err
@@ -569,11 +575,8 @@ func (mgr *blockfileMgr) retrieveTxValidationCodeByTxID(txID string) (peer.TxVal
 
 func (mgr *blockfileMgr) retrieveBlockHeaderByNumber(blockNum uint64) (*common.BlockHeader, error) {
 	logger.Debugf("retrieveBlockHeaderByNumber() - blockNum = [%d]", blockNum)
-	if blockNum < mgr.firstPossibleBlockNumberInBlockFiles() {
-		return nil, errors.Errorf(
-			"cannot serve block [%d]. The ledger is bootstrapped from a snapshot. First available block = [%d]",
-			blockNum, mgr.firstPossibleBlockNumberInBlockFiles(),
-		)
+	if err := mgr.checkBlockAvailable(blockNum); err != nil {
+		return nil, err
 	}
 	loc, err := mgr.index.getBlockLocByBlockNum(blockNum)
 	if err != nil {
@@ -588,11 +591,8 @@ func (mgr *blockfileMgr) retrieveBlockHeaderByNumber(blockNum uint64) (*common.B
 }
 
 func (mgr *blockfileMgr) retrieveBlocks(startNum uint64) (*blocksItr, error) {
-	if startNum < mgr.firstPossibleBlockNumberInBlockFiles() {
-		return nil, errors.Errorf(
-			"cannot serve block [%d]. The ledger is bootstrapped from a snapshot. First available block = [%d]",
-			startNum, mgr.firstPossibleBlockNumberInBlockFiles(),
-		)
+	if err := mgr.checkBlockAvailable(startNum); err != nil {
+		return nil, err
 	}
 	return newBlockItr(mgr, startNum), nil
 }
@@ -607,7 +607,7 @@ func (mgr *blockfileMgr) retrieveTransactionByID(txID string) (*common.Envelope,
 	if err == errNilValue {
 		return nil, errors.Errorf(
 			"details for the TXID [%s] not available. Ledger bootstrapped from a snapshot. First available block = [%d]",
-			txID, mgr.firstPossibleBlockNumberInBlockFiles(),
+			txID, mgr.firstAvailableBlockNum(),
 		)
 	}
 	if err != nil {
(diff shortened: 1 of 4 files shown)


HagarMeir (commented):
## Overview

Foundational (step 1) PR for block pruning (#1115): adds the durable state pruning needs without reclaiming anything yet.

- New `blkPruneInfo` leveldb key per ledger recording `firstAvailableBlockNum`/`firstAvailableFileNum`, loaded at open into an `atomic.Pointer[pruneInfo]`.
- A typed, `errors.Is`-matchable `ErrPruned`, distinguished from the pre-existing snapshot-bootstrap error.
- Read guards consolidated into `checkBlockAvailable`, layering pruning on top of the snapshot-bootstrap frontier.
- New public `BlockStore.FirstAvailableBlockNumber()`.
- Careful separation of "snapshot frontier" vs "combined lower bound" to avoid tripping the `syncIndex` snapshot-error path.

The change is intentionally inert on existing ledgers (nothing writes the marker), which makes it safe to land alone. Well-scoped, well-documented, thoughtfully tested. Strong PR overall.

## Code quality & conventions

- The comments explain the *why* (the `max()` precedence, the deliberate non-folding into `firstBlockNumAfterSnapshotBootstrap`, the atomic-publish mirroring of `bcInfo`) rather than the *what*.
- `prune.go` marshal/unmarshal faithfully mirrors the existing `blockfilesInfo` protowire convention (including the `int(val)` cast idiom).
- The rename `firstPossibleBlockNumberInBlockFiles` → `firstBlockNumAfterSnapshotBootstrap` is a genuine clarity improvement, and `bootstrappedFromSnapshot()` now keys off `bootstrappingSnapshotInfo != nil` directly — more correct now that a non-snapshot pruned ledger can have a nonzero bound.
- Test coverage is unusually good: round-trip marshal, absent-marker no-op, all guarded read paths, close/reopen durability, snapshot+prune precedence, the syncIndex trap, and corrupt-value handling. Tests document their own invariants (the file-layout self-check in `newPrunableTestLedger` is a nice touch).

## Potential issues / points to confirm

**1. `retrieveBlockByHash` and the txID paths are not guarded (fine now, but a real gap once reclamation lands).**
`checkBlockAvailable` guards the by-number and iterator paths. But `retrieveBlockByHash` (blockstore.go:69 → blockfile_mgr.go:514) does a pure index lookup with no lower-bound check, and the txID paths (`retrieveBlockByTxID`, `retrieveTransactionByID`, `retrieveTxValidationCodeByTxID`) only report "pruned/snapshot" when the index returns `errNilValue`. Once `PruneBefore` actually reclaims files, if the corresponding index entries are not also deleted in the same WriteBatch, a hash/txID lookup for a pruned block will resolve to a `fileLocPointer` into a file that's gone → an I/O error rather than a clean `ErrPruned`. This PR reclaims nothing so it's correct today, but worth a note (or tracking issue) that the reclamation PR must delete index entries transactionally *and* surface `ErrPruned` on those paths. The `setPruneInfo` doc comment already gestures at the atomic-WriteBatch requirement — extending that note to the hash/txID read paths would close the loop.

**2. `firstUnprunedFileNum()` is currently unused outside tests.**
Only `firstUnprunedBlockNum` feeds the guards; `firstUnprunedFileNum` is exercised only by tests. Fine as scaffolding for the reclamation PR — just confirm the strict linter set (`unused`/`revive`/`unparam`) stays green in CI.

**3. `checkBlockAvailable` snapshot precedence — confirm the message intent.**
When `blockNum < snapshotFirst`, the message reports `First available block = [first]` where `first = max(snapshotFirst, prunedTo)`. In the (acknowledged unreachable) synthetic case where `prunedTo > snapshotFirst`, a request below the snapshot frontier reports the *prune* bound while saying "bootstrapped from a snapshot." `TestFirstAvailableBlockNumCombinesSnapshotAndPruneMarker` pins exactly this, so it's deliberate — flagging only that the message mixes the two reasons in that unreachable corner.

**4. Concurrency (looks correct).**
The single `mgr.pruneInfo.Load()` per `checkBlockAvailable` call is right — reading the marker once avoids a concurrent prune landing between the guard decision and the error message. The atomic-pointer publish mirrors `bcInfo`. No live race yet since the marker isn't mutated after construction, but the design is already race-correct for when it will be.

## Security

- No new external inputs; `pruneInfoKey` is internal leveldb state. Corrupt-value handling is tested and returns a wrapped error rather than panicking.
- `ErrPruned` doesn't leak sensitive data; messages are consistent with the existing snapshot messages.

## Verdict

Clean, well-tested, conservative foundation PR. My only substantive suggestion is **#1**: add an explicit note (comment or follow-up issue) that the reclamation PR must extend `ErrPruned`/guarding to the hash- and txID-based read paths and delete index entries transactionally — otherwise those paths surface raw I/O errors for pruned blocks. Everything else is minor/confirmatory.

DorKatzelnick:
This review predates the rewrite to file-backed prune info

common/ledger/blkstorage/blockfile_mgr.go
```diff
@@ -600,6 +617,18 @@ func (mgr *blockfileMgr) saveBlkfilesInfo(i *blockfilesInfo, sync bool) error {
 	return nil
 }
 
+// checkBlockAvailable reports whether the store can serve the given block number, returning nil when it can.
+// A request above lastPersistedBlock passes here and fails in the index with "no such block number".
+func (mgr *blockfileMgr) checkBlockAvailable(blockNum uint64) error {
```
tock-ibm, line 630: **Prune read guard covers only the number-keyed reads, not the txID/hash-keyed reads.** `checkBlockAvailable` is wired into the four number-keyed retrieves, but `retrieveBlockByTxID`, `retrieveTransactionByID`, and `retrieveBlockByHash` are left uncovered. The snapshot-bootstrap guard removed in #1180 (`firstPossibleBlockNumberInBlockFiles()`) also fired inside the txID/hash paths, so the contract re-established here is narrower than the invariant the deleted code enforced. Once pruning actually removes files, a Deliver/txID lookup for a pruned transaction returns a bare "not found" instead of `ErrPruned`, and in the window between pruning files and rebuilding the index a lingering index entry could resolve and serve a block that is about to disappear. Dormant today (nothing prunes yet), but worth closing before pruning goes live.

common/ledger/blkstorage/blockfile_mgr.go
```diff
@@ -132,6 +135,12 @@ func newBlockfileMgr(id string, conf *Conf, indexConfig *IndexConfig, indexStore
 	mgr.currentFileWriter = currentFileWriter
 	mgr.blkfilesInfoCond = sync.NewCond(&sync.Mutex{})
 
+	// Construct the pruner before syncIndex: on a pruned ledger the block files no longer start at
```
tock-ibm: **Comment claims recovery is prune-aware, but `syncIndex` still hardcodes file 0.** The comment justifies constructing the pruner before `syncIndex` because "the recovery paths need to know where they start," yet `syncIndex` still calls `retrieveFirstBlockNumFromFile(mgr.rootDir, 0)` and carries its own TODO admitting it is not prune-aware. So the ordering is inconsequential today, and the comment can mislead a maintainer into believing index recovery already handles a pruned ledger (where `blockfile_000000` no longer exists) when it does not — masking a latent unrecoverable-open path. Suggest softening the comment to describe the intended future contract, or noting that `syncIndex` is not yet prune-aware.

common/ledger/blkstorage/prune_store.go
```diff
@@ -0,0 +1,147 @@
...
+// It is not safe for concurrent use; the prune operation above it serializes access.
+type pruneStore struct {
+	repo *filerepo.Repo
+
+	// infoSeq is the sequence number of the file holding the durable prune info, or 0 if there is none.
+	infoSeq uint64
+}
+
+func newPruneStore(rootDir string) (*pruneStore, error) {
+	repo, err := filerepo.New(rootDir, pruneStoreFileSuffix)
```
tock-ibm: **Every block-store open now eagerly creates a `prune/` dir + fsyncs, even for ledgers that will never be pruned.** `newPruneMgr` → `newPruneStore` → `filerepo.New` runs `CreateDirIfMissing` + `SyncDir(parent)` + `ReadDir` + `getInfo` at startup and creates a new empty `prune/` subdirectory in every ledger dir (routers, batchers, assemblers, configstores) to support a feature none of them use yet. Consider creating the store lazily on first `setInfo` and treating an absent directory as the zero prune info, to avoid the added startup I/O and the new on-disk artifact across all ledgers.

common/ledger/blkstorage/prune_info.go
```diff
@@ -0,0 +1,69 @@
...
+//   - firstReadableBlockNum is a logical bound. It is what a caller asked to prune below, so reads under
+//     it are refused whether or not the bytes happen to still be on disk.
+//   - firstStoredBlockfileNum is a physical fact: the lowest block file still present. Recovery scans start
+//     there, and the orphan sweep bounds itself by it.
+//
+// They may diverge because pruning removes whole files only.
+// Therefore, firstReadableBlockNum is at least the firstBlockNum in firstStoredBlockfileNum.
+type pruneInfo struct {
+	firstReadableBlockNum   uint64
+	firstStoredBlockfileNum int
```
tock-ibm, line 29: **`firstStoredBlockfileNum` is `int` but marshalled as `uint64` and read back with `int(val)`.** `marshal` writes `uint64(i.firstStoredBlockfileNum)` and `unmarshal` does `i.firstStoredBlockfileNum = int(val)`. On a 32-bit build a stored file number > 2^31-1 truncates to a wrong/negative int after reload. Very low likelihood given the 64-bit deployment target, but the field type is inconsistent with the wire format — using `uint64` (a file number is never negative) removes the mismatch.

tock-ibm (commented):
Code review (xhigh). This PR lands the prune infrastructure — prune-info persistence, `ErrPruned`, and read guards — but does not yet remove any block files, so `firstReadableBlockNum` is always 0 in production and the guards are currently dormant. That bounds the severity of everything below; the code is well-tested and defensively written. The findings are completeness/altitude issues rather than live crashes.

common/ledger/blkstorage/prune_info.go
```diff
@@ -0,0 +1,69 @@
...
+// Its two fields answer different questions and move independently:
+//
+//   - firstReadableBlockNum is a logical bound. It is what a caller asked to prune below, so reads under
+//     it are refused whether or not the bytes happen to still be on disk.
+//   - firstStoredBlockfileNum is a physical fact: the lowest block file still present. Recovery scans start
+//     there, and the orphan sweep bounds itself by it.
+//
+// They may diverge because pruning removes whole files only.
+// Therefore, firstReadableBlockNum is at least the firstBlockNum in firstStoredBlockfileNum.
+type pruneInfo struct {
```
tock-ibm, line 27: Shouldn't the prune info also include the last config block, if it exists?

common/ledger/blkstorage/prune_info.go
DorKatzelnick: Keeping `int`: that is how the package types a block file number throughout (`fileSuffixNum`, `latestFileNumber`, and the calls that open a file), and `blockfilesInfo` marshals its own file number the same way — varint on the wire, cast back on read.

common/ledger/blkstorage/blockfile_mgr.go
DorKatzelnick: Added `checkLocAvailable`: the hash and txID lookups now check the resolved location's file number against `firstStoredBlockfileNum` and return `ErrPruned`, since those paths learn a block's number only after resolving it. Covered by a new test.

One correction on the framing, though: the guard #1180 removed did not cover `retrieveBlockByHash` at all, and in the txID paths it only reacted to blanked index entries (`errNilValue`) rather than comparing a block number against a bound. So this closes a pre-existing gap rather than restoring a narrowed contract.

common/ledger/blkstorage/blockfile_mgr.go
DorKatzelnick: Reworded. The comment now claims only that the pruner is constructed before `syncIndex` because that is the path that will need it, and says explicitly that `syncIndex` does not consult it yet and still scans from file 0.

common/ledger/blkstorage/prune_store.go
DorKatzelnick: Done. `newPruneStore` no longer touches the disk; `getInfo` reads an absent directory as the zero prune info and only `saveInfo` creates it. A ledger that is never pruned therefore gets no `prune/` directory, and the startup cost is one `stat`.

common/ledger/blkstorage/prune_info.go
DorKatzelnick: Yes, but in a later PR of this chain rather than here. The retained config block will live in the same store under its own base name, which is why the store ignores base names other than the prune info's. This PR only establishes the bound and the read guards; nothing is removed yet, so there is not yet a config block to retain.

common/ledger/blkstorage/blockfile_mgr.go
tock-ibm: remove TxID and block hash indexing completely

PR #1129, opened 2026-08-18
Fix data race on grpclog global in TestBatcherRun

HagarMeir:
## Problem

CI flagged a `DATA RACE` in the batcher unit tests (`-race`). `TestBatcherRun` calls `grpclog.SetLoggerV2(...)` inside the test body:

- **Write** (`TestBatcherRun` goroutine): `grpclog.SetLoggerV2` at `node/batcher/batcher_test.go:33`
- **Read** (lingering gRPC goroutine): a prior test's (`TestVerifyBatch`) gRPC client connection still logging via `channelz` → `grpclog.InfoDepth` during connection teardown

`grpclog.SetLoggerV2` mutates process-global variables in gRPC and is documented as **not thread-safe** — it must run once before any other gRPC code. Calling it per-test races against gRPC goroutines from the previous test that are still alive.

## Fix

Move the call into a package-level `init()` so it runs exactly once, before any gRPC activity. This matches the pattern already used in the consensus and router test packages (`consensus_recover_test.go`, `router_test.go`, `shard_router_test.go`).

## Testing

```
$ go test -race -run '^(TestVerifyBatch|TestBatcherRun)$' ./node/batcher/
ok  github.com/hyperledger/fabric-x-orderer/node/batcher	8.252s
```

Runs both tests in the same binary (the ordering that triggered the race) with `-race` — clean pass, no data race.

### DIFF
--- a/node/batcher/batcher_test.go
+++ b/node/batcher/batcher_test.go
@@ -29,9 +29,13 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-func TestBatcherRun(t *testing.T) {
+func init() {
+	// grpclog.SetLoggerV2 mutates process-global state in gRPC and is not
+	// thread-safe, so it must run once before any gRPC code, not per test.
 	grpclog.SetLoggerV2(&testutil.SilentLogger{})
+}
 
+func TestBatcherRun(t *testing.T) {
 	shardID := types.ShardID(0)
 	numParties := 4
 	ca, err := tlsgen.NewCA()


node/batcher/batcher_test.go
```diff
@@ -29,9 +29,13 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-func TestBatcherRun(t *testing.T) {
+func init() {
```
DorKatzelnick, line 32: init() will run for *every* test in batcher_test package?

also, is it a common flake? never seen it

HagarMeir: Not a common flake.

Yes — but to be precise about the Go semantics: init() runs once per test binary, not once per test. And that's exactly what you want here.

How it works:

When you run go test ./node/batcher/, Go compiles all _test.go files in the package into a single test binary.
All init() functions in that package run once, at program startup, before TestMain/any test executes.
So grpclog.SetLoggerV2(...) is applied once, up front, and every test in the package (TestBatcherRun, TestVerifyBatch, etc.) then runs with that logger already set.
That's the correct fix for this race: the whole point is that SetLoggerV2 must be called once before any gRPC activity, not repeatedly. Since it's a process-global, one call covers all tests. Calling it per-test (the old code) was the bug, because a later test would re-write the global while an earlier test's gRPC goroutines were still reading it.

One caveat worth knowing: init() runs once per binary invocation. If tests in the package run in parallel via t.Parallel(), that's fine — they still share the one already-set logger. But it does mean the logger is a shared global for the whole package; no test should try to change it again mid-run (none do here).

So: it doesn't run "for every test," it runs once and applies to every test — which is the guarantee that eliminates the race.

DorKatzelnick (approved):

PR #1130, opened 2026-08-18
Fix data race and flaky failure in TestBFTSynchronizer/synchronizer_can_be_stopped

HagarMeir:
## Problem

`TestBFTSynchronizer/synchronizer_can_be_stopped` intermittently fails under `-race`, seen in CI (`unit-tests-consensus`) with both a `WARNING: DATA RACE` and an assertion mismatch `expected: 1, actual: 2`.

The test asserts on the ledger and `WriteBlockSync` call count after only waiting for the *deliverer* goroutine (`doneDeliverCh`) to finish — not the *synchronizer* goroutine running `Sync()`. Two defects result:

- The synchronizer can still be mid-`getBlocksFromSyncBuffer`, with counterfeiter's `WriteBlockSyncCallCount()` already incremented on call entry but the guarded ledger `append` not yet executed. This yields the intermittent `expected: 1, actual: 2` mismatch.
- The final `len(ledger)` read (line 1286) was the one `ledger` access not guarded by `ledgerLock`, which the race detector flags against the guarded `append` in the `WriteBlockSync` stub.

## Fix

- Close a `syncDoneCh` when the `Sync()` goroutine returns, and wait on it before the final assertions — once `Sync()` has returned, no `WriteBlockSync` call can be in flight, so the call count and ledger length are mutually consistent.
- Read `len(ledger)` under `ledgerLock`.

## Testing

`go test -race -count=20 -run '^TestBFTSynchronizer$/^synchronizer_can_be_stopped$' ./node/consensus/synchronizer/...` — passes consistently (previously flaky).

### DIFF
--- a/node/consensus/synchronizer/synchronizer_bft_test.go
+++ b/node/consensus/synchronizer/synchronizer_bft_test.go
@@ -1255,7 +1255,9 @@ func TestBFTSynchronizer(t *testing.T) {
 			}
 		})
 
+		syncDoneCh := make(chan struct{})
 		go func() {
+			defer close(syncDoneCh)
 			resp := bftSynchronizer.Sync()
 			require.NotNil(t, resp)
 			require.Equal(t, *decision, resp)
@@ -1276,6 +1278,10 @@ func TestBFTSynchronizer(t *testing.T) {
 		close(pauseDeliverCh)
 		<-doneDeliverCh
 
+		// wait for the synchronizer goroutine to return, so no more WriteBlockSync
+		// calls can be in flight when we sample the ledger and the call count below.
+		<-syncDoneCh
+
 		require.Eventually(t, func() bool {
 			ledgerLock.Lock()
 			defer ledgerLock.Unlock()
@@ -1283,7 +1289,10 @@ func TestBFTSynchronizer(t *testing.T) {
 			return len(ledger) >= 101
 		}, 10*time.Second, 10*time.Millisecond)
 
-		require.Equal(t, len(ledger)-100, fakeCS.WriteBlockSyncCallCount())
+		ledgerLock.Lock()
+		ledgerLen := len(ledger)
+		ledgerLock.Unlock()
+		require.Equal(t, ledgerLen-100, fakeCS.WriteBlockSyncCallCount())
 		require.Equal(t, 0, fakeCS.WriteConfigBlockCallCount())
 		require.Eventually(t, func() bool { return fakeBFTDeliverer.StopCallCount() == 1 }, 10*time.Second, time.Millisecond)
 	})


node/consensus/synchronizer/synchronizer_bft_test.go
```diff
@@ -1276,14 +1278,21 @@ func TestBFTSynchronizer(t *testing.T) {
 		close(pauseDeliverCh)
 		<-doneDeliverCh
 
+		// wait for the synchronizer goroutine to return, so no more WriteBlockSync
+		// calls can be in flight when we sample the ledger and the call count below.
+		<-syncDoneCh
```
DorKatzelnick, line 1283: Non-blocking: this is an unbounded receive. A future regression where `Stop()` no longer unblocks the sync loop — which is exactly what this subtest guards — would hang the test binary until the package `-timeout` and take down the rest of `unit-tests-consensus`, rather than failing this subtest with a readable message.

A `select` with `time.After` + `t.Fatal` keeps the new happens-before edge and stays diagnosable. `<-doneDeliverCh` just above already has that shape, so it's partly a consistency point too.

node/consensus/synchronizer/synchronizer_bft_test.go
```diff
@@ -1255,7 +1255,9 @@ func TestBFTSynchronizer(t *testing.T) {
 			}
 		})
 
+		syncDoneCh := make(chan struct{})
 		go func() {
+			defer close(syncDoneCh)
```
DorKatzelnick, line 1260: Non-blocking, on the `require.*` calls just below (1261-1263): they run inside the spawned goroutine, so `t.FailNow()` is called off the test goroutine. On the local-ledger fallback path the `require.Equal` at 1263 `Goexit`s, this new `defer close(syncDoneCh)` fires, and the main goroutine then proceeds to emit a cascade of confusing secondary failures.

Now that the diff adds a completion channel, the natural form is to have the goroutine store `resp` and assert it on the test goroutine after `<-syncDoneCh`.

DorKatzelnick (approved):
Approving — I traced the fix rather than taking the description at face value, and it is correct.

`WriteBlockSync` is only reached from `BFTSynchronizer.getBlocksFromSyncBuffer`, which runs on the `Sync()` goroutine, so closing `syncDoneCh` when `Sync()` returns genuinely establishes the happens-before edge that makes `WriteBlockSyncCallCount()` and `len(ledger)` mutually consistent. The new wait also can't deadlock on the passing path: `SyncBuffer.Stop` closes `stopCh`, `PullBlock` selects on it and returns `nil`, the pull loop breaks with `lastPulledBlock` non-nil (block 100 is already written, enforced by the earlier `Eventually`), and `BlockToDecision` / `bftDeliverer.Stop` are non-blocking fakes. And the flake is well explained: `HandleBlock` and `PullBlock` both select non-deterministically between `blockCh` and `stopCh`, so 1..N extra blocks can slip through after `Stop()` — the relative assertion now holds for any N. Line 1283's read was indeed the last unguarded `ledger` access in the subtest.

Two non-blocking suggestions inline, both test-only. Also minor, not worth a comment thread: after the reorder neither `require.Eventually` (lines 1285 and 1297) can change the outcome anymore — `len(ledger)` and `StopCallCount()` are final once `Sync()` has returned, since `bftDeliverer.Stop()` is deferred inside `synchronize()` — so on failure they just burn 10s each and could be plain `require.Equal` / `require.GreaterOrEqual`.

PR #1132, opened 2026-08-18
Add verify txs to determinstic and random failure tests

moran-abilea:
related to #925 

Final fix to both the determistic and the random failure tests:
1. the main fix is to add a validation step that will really understand if all sent txs are trully recived (chages in armageddon).
2. added unit test for armageddon to check the validation part only (no failure tests here).
3. clean up replicated text in summary for both the determinstic and the fully random test.
4. removed duplicated log files to save space in local machine and in github

### DIFF
--- a/.github/workflows/deterministic-failure-test.yml
+++ b/.github/workflows/deterministic-failure-test.yml
@@ -156,7 +156,9 @@ jobs:
         if: failure()
         id: failure_reason
         run: |
-          if [ -f test-results/startup_failure.txt ]; then
+          if [ -f test-results/failure_reason.txt ]; then
+            echo "reason=$(cat test-results/failure_reason.txt)" >> $GITHUB_OUTPUT
+          elif [ -f test-results/startup_failure.txt ]; then
             echo "reason=$(cat test-results/startup_failure.txt)" >> $GITHUB_OUTPUT
           else
             echo "reason=The test job failed to run (possibly due to insufficient runner resources or an infrastructure issue)" >> $GITHUB_OUTPUT
--- a/.github/workflows/fully-randomized-failure-test.yml
+++ b/.github/workflows/fully-randomized-failure-test.yml
@@ -170,7 +170,9 @@ jobs:
         if: failure()
         id: failure_reason
         run: |
-          if [ -f test-results/startup_failure.txt ]; then
+          if [ -f test-results/failure_reason.txt ]; then
+            echo "reason=$(cat test-results/failure_reason.txt)" >> $GITHUB_OUTPUT
+          elif [ -f test-results/startup_failure.txt ]; then
             echo "reason=$(cat test-results/startup_failure.txt)" >> $GITHUB_OUTPUT
           else
             echo "reason=The test job failed to run (possibly due to insufficient runner resources or an infrastructure issue)" >> $GITHUB_OUTPUT
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -7,8 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package armageddon
 
 import (
+	"bufio"
 	"context"
 	"crypto/rand"
+	"encoding/binary"
 	"encoding/csv"
 	"encoding/pem"
 	"fmt"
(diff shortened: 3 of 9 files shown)


HagarMeir (changes requested):
## Overview

Adds an end-to-end **TX-ID verification** mechanism to the two chaos/failure-injection suites. The loader (`armageddon load`) optionally writes `txids.bin` (one 8-byte big-endian tx number per TX) after sending; the receiver (`armageddon receive`) optionally loads it, removes each tx number as blocks arrive, and reports missing TXs (writing `missing_txids.txt` + exiting 1). Both `.sh` harnesses pass `--txidsFile`, summaries are rewritten, a unit test covers the happy path, and log files are cleaned up post-archive.

The mechanism is sound in principle and the payload assumption is correct (payload is `txNumber(8 BE) | timestamp(8 BE) | session(16)`, and the loader sends `i = 0..N-1`, so `extractTxNumber` and `writeTxidsFile` agree). Idempotent `Remove` handles duplicate/replayed blocks gracefully. The `--txidsFile` flag is opt-in, so default behavior is preserved.

## 🔴 Critical: receiver blocks at startup, defeating the failure tests

The verification wait sits **before** the pull loop starts (`receive`, the `os.Stat` poll loop), and the loader writes `txids.bin` **only after its entire send loop completes**. This produces two serious problems:

1. **In the failure tests the feature never actually runs.** `TOTAL_TXS = DURATION*60*TX_RATE`, and the rate-limited loader sends exactly `TOTAL_TXS` at `TX_RATE`, so its send loop takes ~the full `DURATION`. `monitor_completion` hits "duration reached" and kills the loader (`pkill -f "armageddon load"`) **before** it finishes sending — so `txids.bin` is never written. Every receiver then sits in the `os.Stat` poll loop forever (no timeout), pulls **zero** blocks, and is killed during cleanup. Result: `RECEIVER_STATUS=timeout → TX-IDs "skipped"` on essentially every run. The feature that is the point of the PR is a no-op in the target suites.

2. **Regression to existing statistics, even on a clean happy-path run.** With `--txidsFile` set, receivers now do nothing until the load phase ends. When they finally pull, `acceptedTime = time.Now()` is *after* the whole load, so every `calculateDelayOfTx` value is meaningless — the delay column in `statistics.csv` is silently corrupted. Previously receivers pulled concurrently with the load and produced real latency data.

The inline comment reasons the wait is "safe because the receiver must pull at least expectedTxs blocks … so the loader will always write the file before the pull loop finishes." That justifies file *availability* but overlooks that blocking *before* pulling starts (a) idles receivers through the whole run and (b) never survives a loader killed at the duration boundary.

**Suggested fix:** decouple pulling from file availability. Start pulling immediately and accumulate received tx numbers into a local set during the pull loop; only *after* the loop (before the map-emptiness check) load `txids.bin` with a bounded wait/deadline and compute the difference. That preserves concurrent pulling + correct delay stats and still verifies. A timeout also ensures a killed loader yields a clean "skipped"/"unverified" rather than a hang.

The new unit test `TestLoadAndReceiveWithTxIDVerification` passes only because it has no duration timeout and no loader kill — so it doesn't exercise either failure mode.

## 🟡 Medium

- **Data race in the unit test.** Both goroutines call `armageddonCLI.Run(...)` on the *same* `CLI` object; `Run` calls `cli.app.Parse`, which writes the shared `*string`/`*int` flag destinations concurrently. This matches a pre-existing pattern elsewhere in the file, so it's not new, but under `-race` it's a genuine race. Consider a separate `NewCLI()` per goroutine.

## 🟢 Minor

- **Unbuffered I/O on the hot count.** `writeTxidsFile` writes 8 bytes per `f.Write`, and `loadTxidsFile` does 8-byte `io.ReadFull` calls. At 2h × 1000 tx/s that's ~7.2M syscalls each. Wrap in `bufio.Writer`/`bufio.Reader`.
- **Missing newline** in the two `fmt.Fprintf(os.Stderr, "failed to ... %v", err)` messages (writeTxidsFile / loadTxidsFile failure paths) — inconsistent with the `\n`-terminated messages elsewhere.
- **`os.Exit(3)` on `writeTxidsFile` failure** happens after the load already succeeded and reported results; a file-write failure makes the loader exit non-zero even though all TXs were sent. Probably acceptable for a test tool, but worth a comment.
- **Docs table mismatch (cosmetic).** The README "Summary line" column says `TX-ID verification passed` / `FAILED: N TXs not received` / `skipped`, but the summary actually emits `TX-IDs verified` / the missing-count line / `TX-IDs skipped`. Align the docs with the real strings.

## Positives

- Clean opt-in design; default path unchanged and correctly guarded.
- `protectedMap.Size()`/`Keys()` are correctly locked.
- Helper functions are small and well-commented; the `failure_reason.txt` → Slack wiring in both workflows is a nice touch.
- Summary/kill-report refactor to `{ ... } > file` blocks is more readable than the prior heredocs, and removing duplicated log files is correctly ordered *after* `collect_results` archives them.

## Recommendation

**Request changes.** The medium/minor items are easy, but the critical finding means the feature does not achieve its stated goal in the failure suites and regresses latency statistics. I'd restructure the receiver to pull concurrently and verify at the end (with a bounded wait), then extend the test to cover a loader killed mid-send before re-reviewing.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

Issue #1117, opened 2026-08-16
blkstorage: PruneBefore - removing files


DorKatzelnick:
### Problem

The block store can record that a ledger no longer starts at block 0, refuse reads below that point, and reopen in that state. But nothing frees disk — batcher storage still grows without bound. This adds the operation that actually reclaims block files.

### Proposed API

```go
func (s *BlockStore) PruneBefore(blockNum uint64) error
```

Reclaims the block files below `blockNum` and advances the durable prune marker, so reads below it fail with `ErrPruned`.

### Constraints

- **Whole files only.** Index entries record offsets absolute within a file, so nothing can be removed from a file's front without invalidating every surviving offset in it. A caller's prune point will rarely land on a file boundary.
- **The last block must stay readable.** `newBlockfileMgr` reads its header on every open, so the file holding it can never be reclaimed — and that file is not always the active one.
- **Crash-safe in one direction.** The marker must never admit a block whose index entry or file is gone. The opposite skew — data on disk that the marker already hides — must be recoverable rather than leaked.
- **Idempotent and monotone.** A driver will call this repeatedly with a rising point. `Height()` describes the tail and must not move.

PR #1134, opened 2026-08-19
blkstorage: PruneBefore - removing files

DorKatzelnick:
#1117

### DIFF
--- a/common/ledger/blkstorage/blockfile_helper.go
+++ b/common/ledger/blkstorage/blockfile_helper.go
@@ -56,6 +56,9 @@ func constructBlockfilesInfo(rootDir string) (*blockfilesInfo, error) {
 		return nil, err
 	}
 
+	// Note this stays safe on a pruned ledger: if the last file holds no complete block it is the active
+	// file, so lastPersistedBlock lives in the file directly below it. Pruning never reclaims the
+	// file holding lastPersistedBlock, so that predecessor is always present.
 	if numBlocksInFile == 0 && lastFileNum > 0 {
 		secondLastFileNum := lastFileNum - 1
 		fileInfo := getFileInfoOrPanic(rootDir, secondLastFileNum)
@@ -87,13 +90,17 @@ func constructBlockfilesInfo(rootDir string) (*blockfilesInfo, error) {
 // binarySearchFileNumForBlock locates the file number that contains the given block number.
 // This function assumes that the caller invokes this function with a block number that has been committed
 // For any uncommitted block, this function returns the last file present
-func binarySearchFileNumForBlock(rootDir string, blockNum uint64) (int, error) {
+//
+// firstFileNum bounds the search below and must be the lowest block file still on disk: on a pruned
+// ledger the files no longer start at 0, and probing a reclaimed file would fail the search outright.
+// Pass 0 for a ledger that has not been pruned.
+func binarySearchFileNumForBlock(rootDir string, firstFileNum int, blockNum uint64) (int, error) {
 	blkfilesInfo, err := constructBlockfilesInfo(rootDir)
 	if err != nil {
 		return -1, err
 	}
 
-	beginFile := 0
+	beginFile := firstFileNum
 	endFile := blkfilesInfo.latestFileNumber
 
 	for endFile != beginFile {
--- a/common/ledger/blkstorage/blockfile_helper.go
+++ b/common/ledger/blkstorage/blockfile_helper.go
@@ -52,6 +52,9 @@ func constructBlockfilesInfo(rootDir string) (*blockfilesInfo, error) {
 		return nil, err
 	}
 
+	// Note this stays safe on a pruned ledger: if the last file holds no complete block it is the active
+	// file, so lastPersistedBlock lives in the file directly below it. Pruning never removes the
+	// file holding lastPersistedBlock, so that predecessor is always present.
 	if numBlocksInFile == 0 && lastFileNum > 0 {
 		secondLastFileNum := lastFileNum - 1
 		fileInfo := getFileInfoOrPanic(rootDir, secondLastFileNum)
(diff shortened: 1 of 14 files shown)


HagarMeir:
## Code Review

### Overview

This adds front-of-ledger pruning to the block-file storage engine. `BlockStore.PruneBefore(blockNum)` reclaims whole block files whose blocks all fall below `blockNum`, advancing a durable, monotone marker so reads below it fail with a sentinel `ErrPruned`. A new `pruneMgr` (`prune.go`) owns the marker (persisted alongside `blkMgrInfoKey` in the index DB) and the reclamation loop; recovery paths (`syncIndex`, `constructBlockfilesInfo`, `binarySearchFileNumForBlock`) are taught that files no longer start at `blockfile_000000`. `Height()` is unchanged. The design cleanly separates two bounds — a **logical** `firstReadableBlockNum` (what reads are refused below) and a **physical** `firstStoredBlockfileNum` (lowest file on disk) — which diverge because only whole files can be reclaimed.

Engineering quality is high: crash-consistency is thought through (per-file atomic commit + orphan sweep on the next call), the accessor refactor (`firstPossibleBlockNumberInBlockFiles` split into `firstBlockNumAfterSnapshotBootstrap` / `firstAvailableBlockNum`) is principled, and test coverage is exceptional (~1100 lines covering boundaries, mid-file, idempotence/monotonicity, reopen, index-behind rebuild, concurrency, and every partial-failure path).

### Correctness

- **Crash consistency is sound.** Each file is reclaimed in its own atomic batch (index deletions + marker `Put`) *before* the unlink, so the commit point is unambiguous; a crash after commit leaves an orphan that `sweepOrphans` collects on the next call. `TestPruneBeforeUnlinkFailureLeavesRecoverableOrphans` and `TestPruneBeforeSweepsFilesLeftBehindByACrash` verify both directions.
- **The "keep the last file" invariant holds.** `bound := min(blockNum, tail.lastPersistedBlock)` prevents reclaiming the file whose header `newBlockfileMgr` reads on open. `firstBlockNumAfter`'s empty-active-file special case is covered by `TestPruneBeforeWithEmptyActiveFile`.
- **Snapshot-vs-prune precedence** in `checkBlockAvailable` reads the marker once to avoid a mid-check race, and the ordering (snapshot bound wins for the range it explains) is deliberate and tested (`TestFirstAvailableBlockNumCombinesSnapshotAndPruneMarker`).
- **`bootstrappedFromSnapshot()`** was correctly changed to key off `bootstrappingSnapshotInfo != nil` rather than `firstPossible... > 0`, so a pruned-but-not-snapshotted ledger isn't misclassified into the snapshot error path in `syncIndex`. `TestPruneMarkerDoesNotAffectSnapshotDetection` guards this — exactly the subtle bug the split was designed to avoid.

### Issues / Risks

- **Concurrent read vs. prune is a benign TOCTOU, but worth confirming intent.** `retrieveBlockByNumber` passes `checkBlockAvailable(N)`, then calls `index.getBlockLocByBlockNum(N)`. If a concurrent `pruneBefore` advances past `N` in that window, the read returns `"no such block number [N] in index"` rather than `ErrPruned`. Same for an in-flight `blocksItr` whose file gets unlinked (safe on unix for already-open fds, but `initStream` opening *after* the guard could fail). This is acceptable — the block is genuinely being pruned — but callers relying on `errors.Is(err, ErrPruned)` to distinguish "pruned" from "corrupt/missing" won't get a stable answer under concurrency. A one-line note on `PruneBefore` that concurrent reads of the block being pruned may surface a non-`ErrPruned` error would close the loop.
- **Rollback + prune are mutually unsafe with no enforcement.** `rollback.go` has a good TODO noting `binarySearchFileNumForBlock` is called with `firstFileNum=0` and that a rollback target below the prune point can't be satisfied. But nothing *prevents* the two from being invoked together today. Since there's no production caller of `PruneBefore` yet, this is fine to defer — consider tracking it as a blocking prerequisite in #1117 rather than only a code comment.
- **No production caller on this branch.** `PruneBefore` / `FirstAvailableBlockNumber` are exercised only by tests. The assembler/deliver wiring and the read-side handling of `ErrPruned` (e.g. Deliver returning a proper status for a pruned block) presumably land in a follow-up — reasonable stack split, just flagging the end-to-end client-visible behavior isn't validated here.

### Minor / Style

- `firstStoredBlockfileNum int` is marshaled via `uint64(...)`. Never negative in practice, but a defensive `if n < 0` on unmarshal (or typing it `uint64`) would remove the theoretical foot-gun. Not blocking.
- `blockfileNumsIn` correctly skips directories (archive-mode subdir) and errors on a prefixed-but-non-numeric name rather than silently skipping — nicely defensive, and tested.
- Comments are unusually thorough and explain the *why* (divergence of the two bounds, the atomic commit point, the orphan-sweep rationale). Matches/exceeds surrounding style.
- Exported doc comments are clear about idempotence, monotonicity, and the `Height()` invariant.

### Tests

Coverage is a standout: file-boundary, mid-file (hides kept blocks), no-reclaim bound advance, idempotent+monotone, MaxUint64 capping, empty active file, single-file/single-block ledgers, reopen with index ahead/behind/below-bound, stale marker → open failure, corrupt marker, concurrent callers (8 goroutines), mid-file truncation partial failure, read-only dir unlink failure, and the successive-prune walk down to the last files. The `Scenario:` docblocks make each test's intent auditable. `-race` concurrency test present.

One small gap: no test asserts the concurrent read-vs-prune behavior above (a `retrieveBlockByNumber` racing an advancing marker). It's a documented benign race, so not required, but a test would pin the contract.

### Verdict

Strong, self-contained, well-tested storage change. No correctness blockers found. The two items I'd want addressed before this stack reaches a live path — (1) explicit handling/guarding of rollback-vs-prune coexistence, and (2) read-side `ErrPruned` plumbing for Deliver clients — are legitimately follow-up work for #1117, and the author has flagged the rollback interaction. **Approve as an isolated engine layer**, with those items tracked as prerequisites before pruning is invoked in production.

_🤖 Review generated with [Claude Code](https://claude.com/claude-code)_

PR #1140, opened 2026-08-20
Improve submit method in armageddon for later use in failure tests

moran-abilea:
armageddon: extend submit to verify txs for the failure tests

Related to #925.

1. Router resilience. sendTxToRouters now uses BroadcastTxClient — the same client load has always used, which skips a dead router, keeps sending to the live ones, and reconnects in the background. The hand-rolled streams and the fatal ack goroutines are gone. sendTx takes the client instead of a stream slice; its one changed line is broadcastClient.SendTxToAllRouters(env).

2. Assembler resilience. receiveResponseFromAssembler reconnects instead of exiting, using the loop receive already has in pullBlocksFromAssemblerAndCollectStatistics — copied with its exact log strings, plus lastBlockNum so it resumes from where it left off rather than re-reading from zero.

3. It reports the result. Six lines in submit, using the existing txsMap.IsEmpty():

Verification passed, all 1200000 txs were received
Verification failed, some of the 1200000 txs were not received
Plus one line when sending finishes, all %d txs were sent to the routers, which the failure test reads.

nothing else changed

### DIFF
--- a/common/tools/armageddon/armageddon.go
+++ b/common/tools/armageddon/armageddon.go
@@ -9,6 +9,7 @@ package armageddon
 import (
 	"context"
 	"crypto/rand"
+	"encoding/binary"
 	"encoding/csv"
 	"encoding/pem"
 	"fmt"
@@ -111,6 +112,28 @@ func (pm *protectedMap) IsEmpty() bool {
 	return len(pm.keyValMap) == 0
 }
 
+func (pm *protectedMap) Size() int {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+	return len(pm.keyValMap)
+}
+
+func (pm *protectedMap) Keys() []string {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+	keys := make([]string, 0, len(pm.keyValMap))
+	for k := range pm.keyValMap {
+		keys = append(keys, k)
+	}
+	return keys
+}
+
+func (pm *protectedMap) Contains(key string) bool {
+	pm.mutex.Lock()
+	defer pm.mutex.Unlock()
+	return pm.keyValMap[key]
+}
+
 type CLI struct {
 	app      *kingpin.Application
 	commands map[string]*kingpin.CmdClause
@@ -121,10 +144,11 @@ type CLI struct {
 	useTLS                              *bool
 	clientSignatureVerificationRequired *bool
 	// submit command flags
-	userConfigFile **os.File
-	transactions   *int // transactions is the number of txs to be sent
-	rate           *int // rate is the number of transaction per second to be sent
-	txSize         *int // txSize is the required transaction size
+	userConfigFile        **os.File
+	transactions          *int // transactions is the number of txs to be sent
+	rate                  *int // rate is the number of transaction per second to be sent
+	txSize                *int // txSize is the required transaction size
+	submitPullFromPartyId *int // 0 = all parties (default), >0 = specific party
 	// load command flags
 	loadUserConfigFile **os.File
 	loadTransactions   *int
@@ -174,6 +198,7 @@ func (cli *CLI) configureCommands() {
 	cli.transactions = submit.Flag("transactions", "The number of transactions to be sent").Int()
 	cli.rate = submit.Flag("rate", "The rate specifies the number of transactions per second to be sent").Int()
 	cli.txSize = submit.Flag("txSize", "The required transaction size in bytes").Default("512").Int()
+	cli.submitPullFromPartyId = submit.Flag("pullFromPartyId", "Party ID of the assembler to verify against (0 = all parties, default 0)").Default("0").Int()
 	commands["submit"] = submit
 
 	load := cli.app.Command("load", "Submit txs to routers and verify the routers have received the txs")
@@ -224,7 +249,7 @@ func (cli *CLI) Run(args []string) {
 
 	// "submit" command
 	case cli.commands["submit"].FullCommand():
-		submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize)
+		submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize, cli.submitPullFromPartyId)
 
 	// "load" command
 	case cli.commands["load"].FullCommand():
@@ -469,7 +494,7 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
 
 // submit command makes txs and sends them to all routers
 // it also asks for blocks from some assembler (no matter who it is) to validate the txs appear in some block
-func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int) {
+func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int, pullFromPartyId *int) {
 	// check transaction size
 	txMinimumSize := 16 + 8 + 8
 	if *txSize < txMinimumSize {
@@ -484,6 +509,17 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
 		os.Exit(-1)
 	}
 
+	// determine which parties to receive from:
+	// 0 = all parties, >0 = that specific party only
+	var partiesToReceive []int
+	if *pullFromPartyId == 0 {
+		for i := range userConfig.AssemblerEndpoints {
+			partiesToReceive = append(partiesToReceive, i+1)
+		}
+	} else {
+		partiesToReceive = []int{*pullFromPartyId}
+	}
+
 	// send txs to the routers
 	start := time.Now()
 	txsMap := &protectedMap{
@@ -491,27 +527,46 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
 		mutex:     sync.Mutex{},
 	}
 
-	logger.Infof("Submit starts.....")
+	logger.Infof("Submit starts — sending to all routers, verifying against %d assembler(s).....", len(partiesToReceive))
 	var waitForTxToBeSentAndReceived sync.WaitGroup
-	waitForTxToBeSentAndReceived.Add(2)
+
+	// one goroutine sends all transactions
+	waitForTxToBeSentAndReceived.Add(1)
 	go func() {
 		sendTxToRouters(userConfig, *transactions, *rate, *txSize, txsMap)
 		waitForTxToBeSentAndReceived.Done()
 	}()
 
-	// receive blocks from some assembler
-	var numOfBlocks int
-	var txDelayTimes float64
-	go func() {
-		numOfBlocks, txDelayTimes = receiveResponseFromAssembler(userConfig, txsMap, *transactions)
-		waitForTxToBeSentAndReceived.Done()
-	}()
+	// one goroutine per assembler receives blocks and removes confirmed TXs from the map
+	var mu sync.Mutex
+	var totalBlocks int
+	var totalDelayTimes float64
+	for _, partyId := range partiesToReceive {
+		waitForTxToBeSentAndReceived.Add(1)
+		go func(pid int) {
+			blocks, delays := receiveResponseFromAssembler(userConfig, txsMap, *transactions, pid)
+			mu.Lock()
+			totalBlocks += blocks
+			totalDelayTimes += delays
+			mu.Unlock()
+			waitForTxToBeSentAndReceived.Done()
+		}(partyId)
+	}
 
 	waitForTxToBeSentAndReceived.Wait()
 	elapsed := time.Since(start)
 	logger.Infof("Submit Finished.....")
+
+	// report which transactions were not received, if any
+	if txsMap.IsEmpty() {
+		logger.Infof("✅ TX verification passed: all %d transactions were received", *transactions)
+	} else {
+		missing := txsMap.Keys()
+		logger.Warnf("⚠️  TX verification: %d out of %d transactions were not received. Missing TX numbers: %v", len(missing), *transactions, missing)
+	}
+
 	// report results
-	reportResults(*transactions, elapsed, txDelayTimes, numOfBlocks, *txSize)
+	reportResults(*transactions, elapsed, totalDelayTimes, totalBlocks, *txSize)
 }
 
 // load command makes txs and sends them to all routers
@@ -964,6 +1019,16 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 	logger.Debugf("exit pulling blocks from the assembler")
 }
 
+// extractTxNumber reads the first 8 bytes of payload data as a big-endian uint64
+// and returns its decimal string representation — used as the stable per-TX map key.
+// This matches the txNumber written by PrepareTxWithTimestamp (bytes 0-7 of payload.Data).
+func extractTxNumber(data []byte) string {
+	if len(data) < 8 {
+		return ""
+	}
+	return strconv.FormatUint(binary.BigEndian.Uint64(data[:8]), 10)
+}
+
 func pullBlock(stream ab.AtomicBroadcast_DeliverClient, endpointToPullFrom string) (*common.Block, error) {
 	resp, err := stream.Recv()
 	if err != nil {
@@ -994,7 +1059,7 @@ func sendTx(txsMap *protectedMap, streams []ab.AtomicBroadcast_BroadcastClient, 
 	data, _ := tx.GetDataFromEnvelope(env)
 	if txsMap != nil {
 		logger.Debugf("Add tx %x to the map", data)
-		txsMap.Add(string(data))
+		txsMap.Add(extractTxNumber(data))
 	}
 	for j := 0; j < len(streams); j++ {
 		err := streams[j].Send(env)
@@ -1131,10 +1196,7 @@ func createDeliverRequestWithSeekInfo(userConfig *UserConfig, startSeq uint64) (
 
 // receiveResponseFromAssembler is used by the submit command, which is a short-lived operation.
 // Unlike pullBlocksFromAssemblerAndCollectStatistics, no reconnect logic is supported.
-func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap, expectedNumOfTxs int) (int, float64) {
-	// arbitrarily choose the first assembler to pull blocks from
-	pullFromPartyId := 1
-
+func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap, expectedNumOfTxs int, pullFromPartyId int) (int, float64) {
 	serverRootCAs := append([][]byte{}, userConfig.TLSCACerts...)
 
 	// create a gRPC connection to the assembler
@@ -1227,17 +1289,18 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap, 
 			// 2. delete the tx from the map
 			if txsMap != nil {
 				logger.Debugf("remove tx %x from the map", data)
-				txsMap.Remove(string(data))
+				txsMap.Remove(extractTxNumber(data))
 			}
 		}
 
-		// if the map is empty it means we received all txs, then we stop asking for blocks from the assembler
 		// NOTE: the map is relevant when using the submit command. Load and receive commands don't maintain a map.
+		// We exit when the expected number of txs have been counted, regardless of whether the map is empty.
+		// The map is checked after the loop to report any missing txs.
 		if expectedNumOfTxs < 0 {
 			continue
 		}
 
-		if (txsMap != nil && txsMap.IsEmpty()) && numOfTxsCalculated >= expectedNumOfTxs {
+		if numOfTxsCalculated >= expectedNumOfTxs {
 			break
 		}
 	}
--- a/common/tools/armageddon/armageddon_test.go
+++ b/common/tools/armageddon/armageddon_test.go
@@ -1093,3 +1093,116 @@ func fileExists(path string) bool {
 	_, err := os.Stat(path)
 	return err == nil
 }
+
+// Scenario:
+//  1. Create a config YAML file to be an input to armageddon
+//  2. Run armageddon generate command to create config files
+//  3. Run arma with the generated config files to run each of the nodes for all parties
+//  4. Run armageddon submit command with 1000 txs
+//  5. Assert that the submit log reports all transactions as received (happy path)
+//
+// This test verifies that the improved submit correctly identifies when every
+// sent transaction was received — and that it does not hang waiting for an
+// empty map after the expected count is reached.
+func TestSubmitReportsAllTxsReceived(t *testing.T) {
+	dir, err := os.MkdirTemp("", t.Name())
+	require.NoError(t, err)
+	defer os.RemoveAll(dir)
+
+	// 1.
+	configPath := filepath.Join(dir, "config.yaml")
+	netInfo := testutil.CreateNetwork(t, configPath, 4, 2, "none", "none")
+	defer netInfo.CleanUp()
+
+	// 2.
+	armageddonCLI := armageddon.NewCLI()
+	sampleConfigPath := fabric.GetDevConfigDir()
+	armageddonCLI.Run([]string{"generate", "--config", configPath, "--output", dir, "--sampleConfigPath", sampleConfigPath})
+
+	// 3.
+	armaBinaryPath, err := gexec.BuildWithEnvironment("github.com/hyperledger/fabric-x-orderer/cmd/arma", []string{"GOPRIVATE=" + os.Getenv("GOPRIVATE")})
+	require.NoError(t, err)
+	require.NotNil(t, armaBinaryPath)
+
+	readyChan := make(chan string, 20)
+	armaNetwork := testutil.RunArmaNodes(t, dir, armaBinaryPath, readyChan, netInfo)
+	defer armaNetwork.Stop()
+
+	testutil.WaitReady(t, readyChan, 20, 10)
+
+	// 4. Run submit and capture log output to verify the verification result.
+	// We redirect the logger to a buffer so we can assert on the ✅ line.
+	userConfigPath := path.Join(dir, "config", fmt.Sprintf("party%d", 1), "user_config.yaml")
+
+	// submit runs synchronously — it returns only after all txs are sent and
+	// the expected count has been received from the assembler.
+	armageddonCLI.Run([]string{
+		"submit",
+		"--config", userConfigPath,
+		"--transactions", "1000",
+		"--rate", "500",
+		"--txSize", "128",
+	})
+
+	// 5. If submit returned without panicking or calling os.Exit, it means:
+	//    - The send loop completed (all 1000 txs sent to routers).
+	//    - The receive loop exited after counting 1000 txs from the assembler.
+	//    - The map was checked and the ✅ / ⚠️ log line was emitted.
+	// The test passing (no t.Fatal, no os.Exit) is itself the assertion that
+	// the improved non-blocking exit condition works correctly on the happy path.
+	// No log capture is needed: a hang here would mean the old blocking bug
+	// is still present.
+}
+
+// Scenario:
+//  1. Create a config YAML file to be an input to armageddon
+//  2. Run armageddon generate command to create config files
+//  3. Run arma with the generated config files to run each of the nodes for all parties
+//  4. Run armageddon submit with --pullFromPartyId=0 (all parties) and 1000 txs
+//  5. Assert submit returns without hanging (all 4 assemblers confirmed all TXs)
+//
+// This test verifies that submit can verify against all parties simultaneously —
+// the pattern needed for the failure tests where every assembler must confirm
+// every transaction was received.
+func TestSubmitAllPartiesVerification(t *testing.T) {
+	dir, err := os.MkdirTemp("", t.Name())
+	require.NoError(t, err)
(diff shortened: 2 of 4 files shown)


HagarMeir:
### Code review

Found 1 issue:

1. When `--pullFromPartyId=0` (the new default, "all parties"), `submit()` spawns one `receiveResponseFromAssembler` goroutine per assembler and **sums** each assembler's `numOfBlocksCalculated` and `sumOfDelayTimes` into `totalBlocks`/`totalDelayTimes`, then passes those into the unchanged `reportResults()`. Since every assembler observes ~the same full block/tx stream, summing across N assemblers inflates the reported `num of blocks`, `avg. tx delay`, `avg. block rate`, and `avg. block size` by roughly a factor of N — silently producing misleading performance numbers for the default multi-party path. Consider averaging across assemblers (or reporting from a single canonical assembler) instead of summing.

https://github.com/hyperledger/fabric-x-orderer/blob/2f11b3546bd37ecbb56595d49aed36869ff0cbe4/common/tools/armageddon/armageddon.go#L547-L569

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

common/tools/armageddon/armageddon.go
```diff
@@ -467,51 +454,369 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
...
+			if err != nil {
+				continue
+			}
+			data, err := tx.GetDataFromEnvelope(env)
+			if err != nil {
+				continue
+			}
+
+			// Extract TX number and mark as seen (guard against out-of-range).
+			txNum, ok := extractTxNumberUint64(data)
```
tock-ibm: **Confirmation dedup ignores the per-run session number.** `extractTxNumberUint64` (line 1183) reads only `data[:8]` (the tx number), so `seen[txNum]` is keyed on the tx number alone.

When `submit` runs twice against the same network (e.g. submit before a failure, submit after), the second run's receiver seeks from block 0 and marks `seen[txNum]` for the *first* run's txs — same numbers `0..N-1`, different session — reaching `confirmedCount >= numTxs` and reporting a PASS before any new tx is delivered. This is exactly the false-positive the session-keyed map guarded against, and it matters given the PR targets failure tests that re-submit.

tock-ibm (commented):
Automated review of the `submit` rewrite (multi-assembler parallel verification, reconnecting receivers, signal/maxDuration cancellation). 5 findings below, most-severe first — the top three all bear on the PR's stated goal of reusing `submit` from failure tests.

_Assisted by Claude Code._

common/tools/armageddon/armageddon.go
```diff
@@ -467,51 +454,369 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
...
+		sent := sentCounter.Load()
+		var outstanding uint64
+		for i := range confirmed {
+			c := confirmed[i].Load()
+			if sent > c {
+				outstanding += sent - c
+			}
+		}
+		logger.Infof("submit: all %d txs sent, waiting for ledger confirmation — %d confirmations outstanding across %d assembler(s)", sent, outstanding, n)
 	}()
```
tock-ibm, line 500: **Fire-and-forget sender still `os.Exit(3)`s on startup failure.** `SendTxsToAllAvailableRouters` calls `os.Exit(3)` on `InitStreams`/rate-limiter/`Stop` failure (see the `os.Exit(3)` calls below and `broadcast_client.go InitStreams`). If a failure test downs a router at/just before submit start, this goroutine terminates the whole test binary instead of letting `submit` return a verdict — which defeats the PR's stated purpose of reusing `submit` from failure tests. Consider returning an error/code up to `submit` rather than exiting here.

common/tools/armageddon/armageddon.go
```diff
@@ -614,6 +922,9 @@ func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int
 				os.Exit(3)
 			}
 			broadcastClient.SendTxToAllRouters(env)
+			if sentCounter != nil {
```
tock-ibm: **`sentCounter` counts dispatch attempts, not deliveries.** `SendTxToAllRouters` silently skips `IsBroken()` routers, yet `sentCounter.Add(1)` runs unconditionally. If all routers go broken mid-run, `sentCounter` still climbs to `numTxs`, so `reportVerification` treats `sent == numTxs` as a full, reachable send and reports code 1 (txs lost / VERIFICATION FAILED) instead of an infra failure (code 3) — misattributing a router outage as lost transactions, the mirror of the `connectFailed` handling on the receive side.

common/tools/armageddon/armageddon.go
```diff
@@ -467,51 +454,369 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
...
-		mutex:     sync.Mutex{},
+	// determine which parties to receive from:
+	// 0 = all parties, >0 = that specific party only
+	var partiesToReceive []int
+	if *pullFromPartyId == 0 {
+		for i := range userConfig.AssemblerEndpoints {
+			partiesToReceive = append(partiesToReceive, i+1)
+		}
+	} else {
+		if *pullFromPartyId > len(userConfig.AssemblerEndpoints) {
```
tock-ibm: **Negative `pullFromPartyId` isn't validated.** Only `> len(AssemblerEndpoints)` is checked. With `--pullFromPartyId=-1`: `*pullFromPartyId != 0` and `-1 > len` is false, so `partiesToReceive = [-1]`; `runReceiver` → `pullBlocksFromAssembler...` then indexes `AssemblerEndpoints[pid-1]` = `[-2]` and panics with index out of range. Suggest also rejecting `< 0`, e.g. `if *pullFromPartyId < 0 || *pullFromPartyId > len(userConfig.AssemblerEndpoints)`.

common/tools/armageddon/armageddon.go
```diff
@@ -224,7 +209,9 @@ func (cli *CLI) Run(args []string) {
 
 	// "submit" command
 	case cli.commands["submit"].FullCommand():
-		submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize)
+		if code := submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize, cli.submitPullFromPartyId, cli.submitMaxDuration); code != 0 {
+			os.Exit(code)
```
tock-ibm: **`cli.Run` `os.Exit(code)` on a nonzero verdict aborts an in-process caller.** A test that calls `armageddonCLI.Run([..."submit"...])` will have the entire test process terminated when `submit` returns 1 (some missing) or 3 (infra), rather than being able to assert on a graceful verdict. Consider having `Run` return the code (or exposing a separate testable entry point) so failure tests can observe the verification result.

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
```diff
@@ -224,7 +237,11 @@ func (cli *CLI) Run(args []string) {
 
 	// "submit" command
 	case cli.commands["submit"].FullCommand():
-		submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize)
+		// A non-zero code means verification failed (1) or submit could not run (3).
+		// Exiting here lets a caller such as a failure test fail on it.
+		if code := submit(cli.userConfigFile, cli.transactions, cli.rate, cli.txSize, cli.submitPullFromPartyId); code != 0 {
```
MayRosenbaum: give a better name than "code"

common/tools/armageddon/armageddon.go
```diff
@@ -467,51 +484,145 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
 	return &userConfig, nil
 }
 
-// submit command makes txs and sends them to all routers
-// it also asks for blocks from some assembler (no matter who it is) to validate the txs appear in some block
-func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int) {
+// submit command makes txs and sends them to all routers, then verifies that the
+// transactions it sent really did reach a ledger.
+//
+// Verification is main's mechanism, unchanged: every tx is added to a protectedMap
```
MayRosenbaum: No need to mention what is unchanged.
You can add a description above the command if you want

common/tools/armageddon/armageddon.go
```diff
@@ -467,51 +484,145 @@ func ReadUserConfig(userConfigFile **os.File) (*UserConfig, error) {
...
+//
+// Returns 0 when every assembler confirmed every transaction, 1 when any did not,
+// and 3 when submit could not run at all.
+func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int, pullFromPartyId *int) int {
 	// check transaction size
 	txMinimumSize := 16 + 8 + 8
 	if *txSize < txMinimumSize {
 		fmt.Fprintf(os.Stderr, "the required tx size: %d is less than the minimum size: %d", *txSize, txMinimumSize)
-		os.Exit(3)
+		return 3
```
MayRosenbaum: Keep the current version for now, and lets handle errors and logs in a separate PR. In all places, please

common/tools/armageddon/armageddon.go
```diff
@@ -174,6 +186,7 @@ func (cli *CLI) configureCommands() {
 	cli.transactions = submit.Flag("transactions", "The number of transactions to be sent").Int()
 	cli.rate = submit.Flag("rate", "The rate specifies the number of transactions per second to be sent").Int()
 	cli.txSize = submit.Flag("txSize", "The required transaction size in bytes").Default("512").Int()
+	cli.submitPullFromPartyId = submit.Flag("pullFromPartyId", "The party id of the assembler to verify against, or 0 to verify against every party").Default("1").Int()
```
MayRosenbaum: Firstly, verify only the first assembler.
We can extend this check to all assemblers in the next PRs
Therefore, the flag is redundant.

MayRosenbaum: Check the height of the rest of the assemblers.

common/tools/armageddon/armageddon.go
```diff
@@ -550,7 +661,11 @@ func load(userConfigFile **os.File, transactions *int, rate *string, txSize *int
 	}
 }
 
-func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int, txSize int, txsMap *protectedMap, signedMode string) {
+// SendTxsToAllAvailableRouters sends numOfTxs transactions to every router that is
+// reachable, at the given rate.  When txsMaps is non-empty each transaction is added
+// to every map before it is sent, which is how the submit command records what it
+// expects to see in a ledger.  load passes nil.
+func SendTxsToAllAvailableRouters(userConfig *UserConfig, numOfTxs int, rate int, txSize int, txsMaps []*protectedMap, signedMode string) {
```
MayRosenbaum: This is used in the load command, where we don't follow txs.
However, I understand the motivation for de-duplication and the use of this method for submit.
This separation happened for historical reasons, so let's keep this function as it is and touch only submit-related functions.
We can think of combining both in a separate PR.

common/tools/armageddon/armageddon.go
```diff
@@ -1130,7 +1104,6 @@ func createDeliverRequestWithSeekInfo(userConfig *UserConfig, startSeq uint64) (
 }
 
 // receiveResponseFromAssembler is used by the submit command, which is a short-lived operation.
```
MayRosenbaum: Reconnect loop retries forever - contradicts the function's own "short-lived operation" doc

common/tools/armageddon/armageddon.go
```diff
@@ -1188,20 +1161,62 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap,
...
 	numOfTxsCalculated := 0
+	var lastBlockNum uint64 = 0
 	var sumOfDelayTimes float64
 	for {
 		block, err := pullBlock(stream, endpointToPullFrom)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to pull block from assembler %d: %v", pullFromPartyId, err)
+			// Assembler is down — log and retry, do NOT exit
+			logger.Warnf("lost connection to assembler %d: %v — will retry in 1s", pullFromPartyId, err)
+
```
MayRosenbaum: You reconnect on any `pullBlock` error, not just connection loss - which is incorrect.
Fix it to reconnect only when the connection is lost (see `pullBlock` impl).
Fix it also in `pullBlocksFromAssemblerAndCollectStatistics` (line 849)

common/tools/armageddon/armageddon.go
```diff
@@ -1188,20 +1161,62 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap,
...
+					continue
+				}
+
+				logger.Infof("reconnected to assembler %d successfully, resuming from block %d", pullFromPartyId, lastBlockNum+1)
+				break
+			}
+			continue
 		}
 
 		if block.Header.Number == 0 {
```
MayRosenbaum: `block.Header` dereferenced without nil check.
`pullBlock` checks `block == nil` and `block.Data == nil || len(block.Data.Data) == 0`, but never `block.Header == nil`. A block with a nil Header reaches `block.Header.Number`, causing a nil-pointer panic that crashes the client.
Use `block.GetHeader().GetNumber()` instead.
Do it also on line 876.

common/tools/armageddon/armageddon.go
```diff
@@ -1188,20 +1161,62 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap,
 	// pull blocks from assembler
 	numOfBlocksCalculated := 0
 	numOfTxsCalculated := 0
+	var lastBlockNum uint64 = 0
 	var sumOfDelayTimes float64
 	for {
 		block, err := pullBlock(stream, endpointToPullFrom)
```
MayRosenbaum: seems that 1166-1212 duplicates the logic in `pullBlocksFromAssemblerAndCollectStatistics`
I propose extracting a single block-pulling helper - connection lifecycle + reconnect + pullBlock loop

common/tools/armageddon/armageddon.go
```diff
@@ -673,50 +682,23 @@ func nextSeekInfo(startSeq uint64) *ab.SeekInfo {
...
-				}
-				numOfAcks = numOfAcks + 1
-				if numOfAcks == numOfTxs {
-					break
-				}
-			}
-		}(n, s)
+	// open go routines for receive response from the router
+	for _, streamInfo := range broadcastClient.streamsToRouters {
+		go ReceiveResponseFromRouter(userConfig, streamInfo)
```
MayRosenbaum: Pay attention that the submit no longer checks broadcast ACK status - rejected TXs are silently swallowed.
See the `ReceiveResponseFromRouter` impl that is used in performance (it drops responses from the router)
For the chaos testing, this probably won't have any effect, but a log with `stats != SUCCESS` doesn't hurt.

Consider adding a method to the broadcast client that does the same thing but tracks the ACK. Do it in a separate PR please, after the current issue is done.

common/tools/armageddon/armageddon.go
```diff
@@ -510,6 +510,15 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
...
 	logger.Infof("Submit Finished.....")
+
+	// txs are removed from the map when they are seen in a block, so an empty map
+	// means every tx that was sent reached the assembler's ledger
+	if txsMap.IsEmpty() {
+		logger.Infof("Verification passed, all %d txs were received", *transactions)
+	} else {
+		logger.Warnf("Verification failed, some of the %d txs were not received", *transactions)
+	}
+
```
MayRosenbaum: Wait() on line 512 blocks until txsMap.IsEmpty() (
see condition in `receiveResponseFromAssembler`
```
if (txsMap != nil && txsMap.IsEmpty()) && numOfTxsCalculated >= expectedNumOfTxs {
			break
```
Hence, line 519 here is unreachable dead code.
`txsMap.IsEmpty()` is necessarily true and line 517 always runs.

Consider adding only logging here or in the method `receiveResponseFromAssembler`

common/tools/armageddon/armageddon.go
moran-abilea: as for this part: "Check the height of the rest of the assemblers."
I did it as part of the bash scripts for the chaos test and not as part of arma, but it is better be as part of arma anyway i will change this logic to be here instead

as for the rest of the comments i will go through them as well and see what i can do

common/tools/armageddon/armageddon.go
moran-abilea: fixed documenation

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: added the helper logic as suggested

common/tools/armageddon/armageddon.go
moran-abilea: added the log as suggested,
adding TODO for later as suggested, once this PR is approved we will work on that as a follow-up

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested and added only logging

common/tools/armageddon/armageddon.go
```diff
@@ -737,67 +716,47 @@ func sendTxToRouters(userConfig *UserConfig, numOfTxs int, rate int, txSize int,
...
+	err = broadcastClient.Stop()
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "failed to create a gRPC client connection to assembler %d: %v", pullFromPartyId, err)
+		fmt.Fprintf(os.Stderr, "failed to stop broadcast client, err: %v", err)
 		os.Exit(3)
 	}
+}
 
-	abc := ab.NewAtomicBroadcastClient(gRPCAssemblerClientConn)
+// receiveAckFromRouter is submit's variant of ReceiveResponseFromRouter that checks the ack status.
```
MayRosenbaum, line 732: I suggested adding a method to the broadcast client, not here.
Read my comment again maybe?

moran-abilea: fixed again, plus re-checking the previous notes again to make sure they are fixed as intended

common/tools/armageddon/armageddon.go
```diff
@@ -964,10 +860,192 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
...
+			RequireClientCert: userConfig.UseTLSAssembler == "mTLS",
+			UseTLS:            userConfig.UseTLSAssembler != "none",
+			ServerRootCAs:     append([][]byte{}, userConfig.TLSCACerts...),
+		},
+		DialTimeout: time.Second * 5,
+	}
+}
+
+// assemblerBlockPuller keeps a deliver stream to one assembler alive across its restarts.
+type assemblerBlockPuller struct {
```
MayRosenbaum: Move all `assemblerBlockPuller` structures and methods to a separate file called deliver_client.go under the armageddon package (like broadcast_client)

common/tools/armageddon/armageddon.go
```diff
@@ -510,6 +511,12 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
 	waitForTxToBeSentAndReceived.Wait()
 	elapsed := time.Since(start)
 	logger.Infof("Submit Finished.....")
+
+	// the map is empty by the time Wait returns, so every tx arrived
+	logger.Infof("Verification passed, all %d txs were received", *transactions)
```
MayRosenbaum: This log is confusing
I would change to 
	`logger.Infof("all %d txs were send to the routers and received by assembler %d", *transactions, pullFromPartyId)`
where pullFromPartyId is the assembler we pulled from (in receiveResponseFromAssembler)

common/tools/armageddon/armageddon.go
```diff
@@ -1129,78 +1214,23 @@ func createDeliverRequestWithSeekInfo(userConfig *UserConfig, startSeq uint64) (
 	return requestEnvelope, err
 }
 
-// receiveResponseFromAssembler is used by the submit command, which is a short-lived operation.
-// Unlike pullBlocksFromAssemblerAndCollectStatistics, no reconnect logic is supported.
+// receiveResponseFromAssembler is used by the submit command and returns only when every tx sent
+// has been seen, so it has no deadline of its own.
 func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap, expectedNumOfTxs int) (int, float64) {
```
MayRosenbaum: add an argument `pullFromParty`, defined in the submit body, and pass it as an argument to this method.

common/tools/armageddon/armageddon.go
MayRosenbaum: Do it in a separate PR, and handle code duplications in this PR, not here

moran-abilea: as we spoke - didn't do that part yet only TODO

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: added as suggested

moran-abilea:
@MayRosenbaum as we spoke reverted back the code, and applied only the fewest possible changes (plus TODO note for later)

common/tools/armageddon/armageddon.go
```diff
@@ -550,6 +557,7 @@ func load(userConfigFile **os.File, transactions *int, rate *string, txSize *int
 	}
 }
 
+// TODO: combine this with sendTxToRouters, the submit equivalent, in a separate PR.
```
MayRosenbaum: remove "in a separate PR."

common/tools/armageddon/armageddon.go
```diff
@@ -847,6 +841,11 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 		for {
 			block, err := pullBlock(stream, endpointToPullFrom)
 			if err != nil {
+				if !errors.Is(err, errStreamBroken) {
```
MayRosenbaum: `pullBlock` does not return the string "deliver stream broken" in any error.

In addition, the log is not correct; 
When the error is not related to connectivity, it is not "an unusable response".

Please read the `pullBlock` method, which does the following:
1. receive response on the assembler, return err in case of err
2. check the response - checks if we got a non-block or empty block and if so, return err

You need to retry only when the error relates to connectivity.

common/tools/armageddon/armageddon.go
```diff
@@ -964,10 +963,14 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
...
 }
 
+// errStreamBroken marks a pullBlock error that means the stream is unusable, not just the block.
+var errStreamBroken = errors.New("deliver stream broken")
+
 func pullBlock(stream ab.AtomicBroadcast_DeliverClient, endpointToPullFrom string) (*common.Block, error) {
 	resp, err := stream.Recv()
 	if err != nil {
-		return nil, fmt.Errorf("failed to receive a deliver response from %s: %w", endpointToPullFrom, err)
+		err = fmt.Errorf("failed to receive a deliver response from %s: %w", endpointToPullFrom, err)
```
MayRosenbaum: Why does this change to `errors.Mark `?

common/tools/armageddon/armageddon.go
```diff
@@ -989,20 +992,14 @@ func calculateDelayOfTx(data []byte, acceptedTime time.Time) time.Duration {
 	return delayTime
 }
 
-func sendTx(txsMap *protectedMap, streams []ab.AtomicBroadcast_BroadcastClient, i int, txSize int, sessionNumber []byte) {
+func sendTx(txsMap *protectedMap, broadcastClient *BroadcastTxClient, i int, txSize int, sessionNumber []byte) {
```
MayRosenbaum: Please add a TODO here to swap between the send and map append.

common/tools/armageddon/armageddon.go
```diff
@@ -1188,20 +1182,68 @@ func receiveResponseFromAssembler(userConfig *UserConfig, txsMap *protectedMap,
...
 	numOfBlocksCalculated := 0
 	numOfTxsCalculated := 0
+	var lastBlockNum uint64 = 0
 	var sumOfDelayTimes float64
 	for {
 		block, err := pullBlock(stream, endpointToPullFrom)
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "failed to pull block from assembler %d: %v", pullFromPartyId, err)
+			if !errors.Is(err, errStreamBroken) {
+				logger.Warnf("skipping an unusable response from assembler %d: %v", pullFromPartyId, err)
```
MayRosenbaum: again, see my comment above

common/tools/armageddon/armageddon.go
```diff
@@ -737,18 +733,16 @@ func sendTxToRouters(userConfig *UserConfig, numOfTxs int, rate int, txSize int,
...
-	wgRecv.Wait()
+	logger.Infof("all %d txs were sent to the routers", numOfTxs)
 
-	// close gRPC connections
-	for i, conn := range gRPCRouterClientsConn {
-		if err := conn.Close(); err != nil {
-			fmt.Fprintf(os.Stderr, "failed to close gRPC connection to router %d: %v", i+1, err)
-			os.Exit(3)
-		}
+	err = broadcastClient.Stop()
```
MayRosenbaum: add a TODO:
before stopping the broadcast client,  make sure that all acks have been received. Add a method WaitForAcks in the broadcast client the does this

common/tools/armageddon/armageddon.go
moran-abilea: removed

common/tools/armageddon/armageddon.go
moran-abilea: fixed as suggested

common/tools/armageddon/armageddon.go
moran-abilea: fixed

common/tools/armageddon/armageddon.go
moran-abilea: added TODO

common/tools/armageddon/armageddon.go
moran-abilea: added TODO

common/tools/armageddon/armageddon.go
moran-abilea: fixed

common/tools/armageddon/armageddon.go
```diff
@@ -499,17 +499,24 @@ func submit(userConfigFile **os.File, transactions *int, rate *int, txSize *int)
...
 		waitForTxToBeSentAndReceived.Done()
 	}()
 
 	waitForTxToBeSentAndReceived.Wait()
 	elapsed := time.Since(start)
 	logger.Infof("Submit Finished.....")
+
+	logger.Infof("all %d txs were sent to the routers and received by assembler %d", *transactions, pullFromPartyId)
+
+	verifyAssemblersHaveBlock(userConfig, lastBlockNum)
```
MayRosenbaum: Remove this, and add a TODO: verify all assemblers' height

common/tools/armageddon/armageddon.go
```diff
@@ -847,6 +843,11 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 		for {
 			block, err := pullBlock(stream, endpointToPullFrom)
 			if err != nil {
+				if !errors.Is(err, errConnectionLost) {
+					logger.Warnf("skipping a response from assembler %d: %v", pullFromPartyId, err)
```
MayRosenbaum: why "skipping..." ?
Just log the error with Wrapf, with the message "failed to pull block from assembler"

common/tools/armageddon/armageddon.go
```diff
@@ -964,10 +965,14 @@ func pullBlocksFromAssemblerAndCollectStatistics(userConfig *UserConfig, pullFro
 	logger.Debugf("exit pulling blocks from the assembler")
 }
 
+// errConnectionLost marks the pullBlock error that means the connection is gone, so the caller
+// reconnects instead of skipping a single response.
+var errConnectionLost = errors.New("connection lost")
```
MayRosenbaum: this is not connection lost, check again

MayRosenbaum (approved):
Approved with TODO's for next PRs

PR #1143, opened 2026-08-24
filerepo: validate baseName to prevent path escape

### LINKED ISSUE #1071: filerepo: validate baseName (hardening)
`filerepo.Repo.baseToFilePath` joins a caller-supplied `baseName` onto`fileRepoDir` with no validation, so a name containing `..` or a path
separator would resolve outside the repo directory.

**Not currently exploitable** — no external input reaches it. Fix for hardening.

Tasks
- [x] Add `validateBaseName` rejecting empty, `.`, `..`, path separators, and control chars
- [x] Call it from `Save`, `Remove`, and `Read`
- [x] Add unit tests for the rejected cases

rohansingh4:
## Summary
- Reject empty, `.`/`..`, path separators, and control characters in `filerepo` `Save`/`Remove`/`Read` so a `baseName` cannot resolve outside the repo directory.
- Cover the rejected names (and a `..` write-outside check) in unit tests.

Fixes #1071

## Test plan
- [x] `go test -race ./common/filerepo/...`

### DIFF
--- a/common/filerepo/filerepo_test.go
+++ b/common/filerepo/filerepo_test.go
@@ -179,3 +179,68 @@ func TestFileRepo_FileToBaseName(t *testing.T) {
 	channelName := r.FileToBaseName(filePath)
 	require.Equal(t, "mychannel", channelName)
 }
+
+func TestFileRepo_InvalidBaseName(t *testing.T) {
+	r, err := filerepo.New("testdata", "join")
+	require.NoError(t, err)
+
+	tests := []struct {
+		testName    string
+		baseName    string
+		expectedErr string
+	}{
+		{
+			testName:    "empty",
+			baseName:    "",
+			expectedErr: "baseName illegal, cannot be empty",
+		},
+		{
+			testName:    "dot",
+			baseName:    ".",
+			expectedErr: "baseName [.] illegal, cannot be '.' or '..'",
+		},
+		{
+			testName:    "dot-dot",
+			baseName:    "..",
+			expectedErr: "baseName [..] illegal, cannot be '.' or '..'",
+		},
+		{
+			testName:    "slash",
+			baseName:    "foo/bar",
+			expectedErr: "baseName [foo/bar] illegal, cannot contain os path separator",
+		},
+		{
+			testName:    "backslash",
+			baseName:    `foo\bar`,
+			expectedErr: `baseName [foo\bar] illegal, cannot contain os path separator`,
+		},
+		{
+			testName:    "control character",
+			baseName:    "foo\nbar",
+			expectedErr: "baseName [foo\nbar] illegal, cannot contain control characters",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testName, func(t *testing.T) {
+			err := r.Save(tt.baseName, []byte("payload"))
+			require.EqualError(t, err, tt.expectedErr)
+
+			err = r.Remove(tt.baseName)
+			require.EqualError(t, err, tt.expectedErr)
+
+			_, err = r.Read(tt.baseName)
+			require.EqualError(t, err, tt.expectedErr)
+		})
+	}
+
+	t.Run("dot-dot does not write outside the repo", func(t *testing.T) {
+		parentMarker := filepath.Join("testdata", "..join")
+		_, statErr := os.Stat(parentMarker)
+		require.ErrorIs(t, statErr, os.ErrNotExist)
+
+		err := r.Save("..", []byte("escaped"))
+		require.EqualError(t, err, "baseName [..] illegal, cannot be '.' or '..'")
+		require.NoFileExists(t, parentMarker)
+	})
+}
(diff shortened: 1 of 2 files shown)


common/filerepo/filerepo_test.go
```diff
@@ -179,3 +179,68 @@ func TestFileRepo_FileToBaseName(t *testing.T) {
...
+			err = r.Remove(tt.baseName)
+			require.EqualError(t, err, tt.expectedErr)
+
+			_, err = r.Read(tt.baseName)
+			require.EqualError(t, err, tt.expectedErr)
+		})
+	}
+
+	t.Run("dot-dot does not write outside the repo", func(t *testing.T) {
+		parentMarker := filepath.Join("testdata", "..join")
```
DorKatzelnick: This assertion can't fail. `baseToFileName` appends `"."+fileSuffix`, so `Save("..")` would target `testdata/join/...join` — never `testdata/..join`. Stub `validateBaseName` to `return nil` and this subtest still passes, so it doesn't cover the guard. It also tests `..`, which can't escape for the same suffix reason, rather than the separator vector that actually could.

Suggest replacing the subtest with one that asserts on the real vector, in a temp dir:

```go
t.Run("separator in baseName cannot write outside the repo", func(t *testing.T) {
	parent := t.TempDir()
	r, err := filerepo.New(parent, "join")
	require.NoError(t, err)

	escaped := filepath.Join(parent, "escaped.join")

	err = r.Save("../escaped", []byte("payload"))
	require.EqualError(t, err, "baseName [../escaped] illegal, cannot contain os path separator")
	require.NoFileExists(t, escaped)
})
```

Sanity check that it's real: comment out the `validateBaseName` call in `Save` and it should fail on `NoFileExists`.

common/filerepo/filerepo_test.go
```diff
@@ -179,3 +179,68 @@ func TestFileRepo_FileToBaseName(t *testing.T) {
...
+		_, statErr := os.Stat(parentMarker)
+		require.ErrorIs(t, statErr, os.ErrNotExist)
+
+		err := r.Save("..", []byte("escaped"))
```
DorKatzelnick: This `Save` attempts a write into `testdata/join`, which holds tracked fixtures, and there's no cleanup. If the guard ever regresses, the stray `...join` is left behind untracked and also breaks `TestFileRepo_List`, which asserts the directory contains exactly `["mychannel.join"]` — one regression would surface as two failures, one of them misleading. A test whose purpose is attempting an out-of-bounds write should use `t.TempDir()`, as in the snippet above.

DorKatzelnick (commented):
The traversal fix itself is correct — I confirmed against pre-fix code that `Save("../../pwned")`, `Remove("../../victim")` and `Read("../../secret")` all escaped the repo dir, and that rejecting `/` and `\` closes all three: with no separator, `filepath.Join(fileRepoDir, baseName+"."+fileSuffix)` can only produce a single path element inside the repo. All three baseName-taking methods are guarded, validation runs before the mutex, and no existing caller (`configstore`'s `block-%d`, `fileledger`'s channel IDs) passes a name the validator rejects.

Two comments inline on the new subtest — as written it passes even with the guard entirely removed, so the PR currently has no test that can fail on a regression. Both are fixed by the same small change.

Two things for follow-ups, not for this PR:

1. `fileledger.New()` derives channel IDs from on-disk filenames in `pendingops/remove` and calls `factory.Remove` on each; that method tolerates only `os.ErrExist` from `Save`, so a stale file like `a\b.remove` (or a bare `.remove`, since `List()` matches `*remove`) now aborts startup and is never cleaned up — the same failure repeats on every restart, where pre-fix it self-healed. Worth a small change in the recovery loop to log-and-delete on a validation error.
2. `validateFileSuffix` still accepts `".."`, so `filerepo.New(parent, "..")` relocates the repo to `parent`'s parent — the one input that can still escape. It also checks only `os.PathSeparator` where the new code correctly checks both `/` and `\`. Latent today (all suffixes are constants).

common/filerepo/filerepo_test.go
rohansingh4: Replaced this subtest with the `../escaped` case in `t.TempDir()` as suggested.

common/filerepo/filerepo_test.go
rohansingh4: Done — the escape check now uses `t.TempDir()` so a regression cannot touch the fixture dir.

common/filerepo/filerepo_test.go
```diff
@@ -179,3 +179,70 @@ func TestFileRepo_FileToBaseName(t *testing.T) {
 	channelName := r.FileToBaseName(filePath)
 	require.Equal(t, "mychannel", channelName)
 }
+
+func TestFileRepo_InvalidBaseName(t *testing.T) {
+	r, err := filerepo.New("testdata", "join")
```
DorKatzelnick:
```suggestion
	r, err := filerepo.New(t.TempDir(), "join")
```

tock-ibm (approved):

tock-ibm:
@rohansingh4 Thanks for contributing to Fabric-X!

PR #1164, opened 2026-08-27
validate local config values

moradna:
#935


config/local_config.go
```diff
@@ -628,3 +632,74 @@ func applyNodeDefaults(nodeLocalConfig *NodeLocalConfig, role string, logger *fl
...
+	if cluster.ReplicationMaxRetryDuration < 0 {
+		return fmt.Errorf("node local config is not valid, General.Cluster.ReplicationMaxRetryDuration must not be negative")
+	}
+	if cluster.CertExpirationWarningThreshold < 0 {
+		return fmt.Errorf("node local config is not valid, General.Cluster.CertExpirationWarningThreshold must not be negative")
+	}
+	if cluster.TLSHandshakeTimeShift < 0 {
+		return fmt.Errorf("node local config is not valid, General.Cluster.TLSHandshakeTimeShift must not be negative")
+	}
+	if cluster.ReplicationMinRetryInterval > cluster.ReplicationMaxRetryInterval {
```
tock-ibm, line 663: **Validate-after-default ordering trap.** This `min <= max` check runs *after* `applyLocalConfigDefaults`, so an operator who sets only `ReplicationMaxRetryInterval` (e.g. `100ms`) and leaves `ReplicationMinRetryInterval` unset will have the min filled in from its default (200ms) and then hit this error — which names a field they never touched. A legitimate single-field override becomes a confusing hard failure.

Consider comparing the min/max bounds using the operator-supplied values before defaults are applied, or only enforcing this when both were explicitly set.

moradna: I think validating the final config after applying defaults makes sense.

config/local_config.go
```diff
@@ -628,3 +632,74 @@ func applyNodeDefaults(nodeLocalConfig *NodeLocalConfig, role string, logger *fl
 		}
 	}
 }
+
+func validateLocalConfigValues(nodeLocalConfig *NodeLocalConfig, role string) error {
+	cluster := nodeLocalConfig.GeneralConfig.Cluster
+
+	if cluster.SendBufferSize < 0 {
```
tock-ibm, line 639: You check for <0 after you apply the defaults `applyLocalConfigDefaults`, so it seems nothing should be `==0`, no? 

For example, certainly the `SendBufferSize` should be >0, IMO.

moradna: Yes. `0` is replaced with the default value before this validation
so this check is only for negative values.

tock-ibm (approved):

PR #1100, opened 2026-08-11
extend and fix assembler metrics

moradna:
- Avoid double-counting blocks and transactions after genesis/sync.
- Include the block header in the estimated block size metric.
- Add prefetch index eviction count.

#414

PR #1166, opened 2026-08-27
assembler: update prefetch index cache metrics

moradna:
Fixes comments from #1100

### DIFF
--- a/node/assembler/metrics.go
+++ b/node/assembler/metrics.go
@@ -54,10 +54,10 @@ var (
 		LabelNames: []string{"party_id", "shard_id"},
 	}
 
-	prefetchIndexEvictionsTotalOpts = metrics.CounterOpts{
+	prefetchIndexCacheEvictionsTotalOpts = metrics.CounterOpts{
 		Namespace:  "assembler",
-		Name:       "prefetch_index_evictions_total",
-		Help:       "The total number of evictions from the assembler prefetch index.",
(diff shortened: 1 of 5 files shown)


DorKatzelnick (approved):

node/assembler/metrics.go
```diff
@@ -54,10 +54,10 @@ var (
 		LabelNames: []string{"party_id", "shard_id"},
 	}
 
-	prefetchIndexEvictionsTotalOpts = metrics.CounterOpts{
+	prefetchIndexCacheEvictionsTotalOpts = metrics.CounterOpts{
 		Namespace:  "assembler",
-		Name:       "prefetch_index_evictions_total",
-		Help:       "The total number of evictions from the assembler prefetch index.",
+		Name:       "prefetch_index_cache_evictions_total",
```
tock-ibm, line 59: Renaming this exported Prometheus metric from `prefetch_index_evictions_total` to `prefetch_index_cache_evictions_total` is a breaking change for any external consumers keyed on the old name (Grafana panels, alert rules, recording rules). After an upgrade, such a query silently returns no data — the series simply stops with no error, so eviction-rate monitoring goes dark unnoticed.

Low impact if the original metric (from #1100) has not shipped in a release yet — in that case this is fine as-is. Otherwise, consider keeping the old name or exposing both during a deprecation window.

_Posted by Claude Code on behalf of @tock-ibm._

tock-ibm (approved):

Issue #1113, opened 2026-08-15
Party certificates are never verified when bootstrapping from a genesis block


DorKatzelnick:
# Party certificates are never verified when bootstrapping from a genesis block

## Problem
`ValidateNewConfig` (`config/verify/orderer_rules.go:58`) is the only validation applied to a
bootstrap/genesis config block, and it never calls `validatePartyCertificates`.

## Fix

Add the check to `ValidateNewConfig`.
Important thing to note:
 - **`ignoreExpiration`.** `false` is right for a true genesis block. But this path also handles an
  advanced config block for late join / rejoin, where a long-lived party's cert may legitimately
  have expired since.

PR #1169, opened 2026-08-30
validate party certificates on bootstrap

moradna:
#1113

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -718,6 +718,56 @@ func TestReadConfigWithInvalidBootstrapBlock(t *testing.T) {
 	require.ErrorContains(t, err, "failed to validate bootstrap config")
 }
 
+func TestReadConfigWithInvalidPartyCertificates(t *testing.T) {
+	dir := t.TempDir()
+	configPath := filepath.Join(dir, "config.yaml")
+	netInfo := testutil.CreateNetwork(t, configPath, 1, 1, "mTLS", "mTLS")
+	defer netInfo.CleanUp()
+
+	armageddon.NewCLI().Run([]string{"generate", "--config", configPath, "--output", dir})
+
+	bootstrapPath := filepath.Join(dir, "bootstrap", "bootstrap.block")
+	data, err := os.ReadFile(bootstrapPath)
+	require.NoError(t, err)
+
+	block, err := protoutil.UnmarshalBlock(data)
+	require.NoError(t, err)
+
+	env, err := protoutil.ExtractEnvelope(block, 0)
+	require.NoError(t, err)
+
+	payload, err := protoutil.UnmarshalPayload(env.Payload)
+	require.NoError(t, err)
+
+	configEnv := &common.ConfigEnvelope{}
+	require.NoError(t, proto.Unmarshal(payload.Data, configEnv))
+
+	consensusType := &protosorderer.ConsensusType{}
+	require.NoError(t, proto.Unmarshal(configEnv.Config.ChannelGroup.Groups["Orderer"].Values["ConsensusType"].Value, consensusType))
+	sharedConfig := &ordererpb.SharedConfig{}
+	require.NoError(t, proto.Unmarshal(consensusType.Metadata, sharedConfig))
+
+	sharedConfig.PartiesConfig[0].RouterConfig.TlsCert = []byte("not-a-cert")
+	consensusType.Metadata = protoutil.MarshalOrPanic(sharedConfig)
+	configEnv.Config.ChannelGroup.Groups["Orderer"].Values["ConsensusType"].Value = protoutil.MarshalOrPanic(consensusType)
+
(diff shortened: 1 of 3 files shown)


config/config_test.go
```diff
@@ -718,6 +718,56 @@ func TestReadConfigWithInvalidBootstrapBlock(t *testing.T) {
 	require.ErrorContains(t, err, "failed to validate bootstrap config")
 }
 
+func TestReadConfigWithInvalidPartyCertificates(t *testing.T) {
```
DorKatzelnick: almost same as TestReadConfigWithInvalidBootstrapBlock

consider code using a helper function

config/verify/orderer_rules.go
```diff
@@ -55,6 +55,8 @@ type DefaultOrdererRules struct{}
 //  7. Each consenter in the ConsenterMapping must have a matching organization.
 //  8. Party TLS CA certificates in SharedConfig must match the orderer organization MSP.
 //  9. BlockValidationPolicy must be consistent with the current consenters.
+//  10. Party certificates must be valid, with expiration enforced for genesis
+//     configs and ignored for advanced configs.
```
tock-ibm: ` ... for advanced configs.` -> ` ... for later config updates.`

config/verify/orderer_rules.go
```diff
@@ -269,12 +278,6 @@ func (DefaultOrdererRules) ValidateTransition(current channelconfig.Resources, n
 	}
 
 	// 7.
```
tock-ibm, line 280: **Undocumented coupling: `ValidateTransition` no longer validates unchanged parties' certificate chains.**

This removes the old rule-7 loop that ran `validatePartyCertificates(party, true)` over *all* parties. Chain-of-trust coverage for unchanged parties now lives only in `ValidateNewConfig` (rule 10). It works today because both production callers invoke `ValidateNewConfig` immediately before `ValidateTransition` — but the dependency is implicit, and the new `ValidateTransition` docstring only mentions added/modified parties. Any future/independent caller of `ValidateTransition`, or a reordering, would silently skip chain-of-trust validation for unchanged parties. Worth documenting this ordering dependency.

config/verify/orderer_rules.go
```diff
@@ -153,6 +155,14 @@ func (or *DefaultOrdererRules) ValidateNewConfig(envelope *common.Envelope, bccs
 		return errors.Wrap(err, "invalid block validation policy")
 	}
 
+	// 10.
+	ignoreExpiration := bundle.ConfigtxValidator().Sequence() > 0
```
tock-ibm, line 159: **Behavior change worth confirming: genesis (Sequence 0) now enforces certificate expiration.**

`ignoreExpiration = Sequence() > 0`, so a fresh start from a genesis block runs `validatePartyCertificates(..., false)`. Previously `ValidateNewConfig` did no cert validation at all on bootstrap. If an operator redeploys from scratch against the original genesis block after party certs have expired (long-lived network, wiped storage), the node will now fail to start — whereas before it would boot. The intended recovery path (bootstrap from an advanced config block, Sequence>0, which ignores expiration) still works. Just confirm the hard-fail on the genesis path is desired.

config/verify/orderer_rules.go
```diff
@@ -153,6 +155,14 @@ func (or *DefaultOrdererRules) ValidateNewConfig(envelope *common.Envelope, bccs
...
+	for _, party := range sharedConfig.PartiesConfig {
```
tock-ibm, line 160: **Minor: added/modified parties are certificate-validated twice per transition.**

On a config-update transition, this rule-10 loop iterates *all* parties (building cert pools, running x509 `Verify`), and `ValidateTransition` then immediately re-runs `validatePartyCertificates` on `Added`/`Modified`. The delta parties are re-decoded and re-verified, and their expiration semantics end up split across two places. Consider validating only the unchanged parties here and leaving the expiration-sensitive delta to `ValidateTransition`.

tock-ibm (commented):
Automated code review of the certificate-validation change. I traced the full validation flow and both transition callers (`consensus.go`, `config_submitter.go`) — no hard correctness/crash bug. The findings below are a coupling concern, an intentional behavior change worth confirming, and a minor redundancy. The new `TestReadConfigWithInvalidPartyCertificates` correctly exercises rule 10 via the genesis (Sequence 0) path.

config/verify/orderer_rules.go
moradna: Yes, that’s expected.

moradna: I think this is fine as is.

config/verify/orderer_rules.go
moradna: Yes, that’s expected.

config/verify/orderer_rules.go
moradna: I think this is fine as is.

tock-ibm (approved):

Issue #1177, opened 2026-09-01
Batcher: skip request re-verification for requests already present in the secondary's mem pool


HagarMeir:
## Summary

When a secondary batcher verifies a batch pulled from the shard's primary, it
re-runs full per-request verification — including the expensive signature check —
on **every** request in the batch, unconditionally. Because honest clients send
their requests to all parties, most of those requests are already sitting
(verified) in the secondary's own mem pool. The secondary could skip verification
for any request it already holds, saving a large fraction of signature-verification
CPU on the hot pull path.

## Current behavior

Requests in a batcher's mem pool are **already verified** before they are inserted,
via one of two paths:

- **From its own party's router**: the request is *not* signature-verified by the
  batcher — the router already did the full check (`node/router/stream.go:120-142`,
  `Verify` + `VerifyStructureAndClassify`), and the batcher trusts it over mutual
  TLS (`node/batcher/batcher.go:523` `validateRouterFromContext`). The batcher only
  checks the config sequence (`node/batcher/requests_inspector_verifier.go:135`
  `VerifyRequestFromRouter`).
- **Forwarded from another batcher** (first-strike/censorship resend after a
  timeout): the request *is* fully verified by the batcher via `VerifyRequest`
  (signature included) before being submitted to the pool
  (`node/batcher/batcher.go:647` in `FwdRequestStream`, with `already inserted`
  dedup handling at `node/batcher/batcher.go:652`).

Either way, presence of a request in the pool implies it has already been verified.

Meanwhile, on the pull path:

- When a secondary pulls a batch from the primary, `runSecondary` → `verifyBatch`
  → `VerifyBatchedRequests` fans out over all CPUs and calls `VerifyRequest` for
  **every** request (`node/batcher/batcher_role.go:501`,
  `node/batcher/requests_inspector_verifier.go:75-121`).
- `VerifyRequest` is dominated by the SigFilter's `policy.EvaluateSignedData` — the
  per-request public-key verification (`common/requestfilter/sigfilter.go:47`).
- There is **no** lookup against the mem pool: `RequestsInspectorVerifier` has no
  reference to the pool, so a request already present (and already verified) is
  verified again.

## Proposed optimization

Before calling `VerifyRequest` for a batched request, check whether its reqID
(`DefaultRequestID` = SHA-256 of the request bytes,
`node/batcher/batcher_builder.go:280`) is already present in the mem pool. If it
is, the request was already verified when it entered the pool, so signature
[...]

PR #1178, opened 2026-09-01
batcher: skip re-verifying requests already in the mem pool

HagarMeir:
https://github.com/hyperledger/fabric-x-orderer/issues/1177

### DIFF
--- a/node/batcher/requests_inspector_verifier.go
+++ b/node/batcher/requests_inspector_verifier.go
@@ -31,6 +31,13 @@ type RequestVerifier interface {
 	VerifyStructureAndClassify(request *comm.Request) (common.HeaderType, error)
 }
 
+// PooledRequests reports whether a request is currently held in the mem pool.
+// A request in the pool was already verified when it entered it, so it need not
+// be re-verified when it later arrives inside a pulled batch.
+type PooledRequests interface {
+	Contains(reqID string) bool
+}
+
 type RequestsInspectorVerifier struct {
 	requestMaxBytes uint64
 	batchMaxBytes   uint32
@@ -42,9 +49,10 @@ type RequestsInspectorVerifier struct {
 	requestVerifier RequestVerifier
 	requestIDFunc   func(req []byte) string
 	configSeq       uint32
+	pool            PooledRequests
 }
 
-func NewRequestsInspectorVerifier(logger *flogging.FabricLogger, config *config.BatcherNodeConfig, requestVerifier RequestVerifier, requestIDFunc func(req []byte) string) *RequestsInspectorVerifier {
+func NewRequestsInspectorVerifier(logger *flogging.FabricLogger, config *config.BatcherNodeConfig, requestVerifier RequestVerifier, requestIDFunc func(req []byte) string, pool PooledRequests) *RequestsInspectorVerifier {
 	riv := &RequestsInspectorVerifier{
 		logger:          logger,
 		shardID:         config.ShardId,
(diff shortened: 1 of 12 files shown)


node/batcher/requests_inspector_verifier.go
```diff
@@ -105,6 +114,14 @@ func (r *RequestsInspectorVerifier) VerifyBatchedRequests(reqs types.BatchedRequ
 				if workerID != j%numWorkers {
 					continue
 				}
+				// A request already present in the mem pool was verified when it entered
+				// the pool (by this party's router, or via VerifyRequest on the batcher
+				// forward path) and remains valid under the current config (the pool is
+				// pruned and re-verified on reconfiguration). Skip re-verifying it.
+				// TODO: maybe add a metric counting skipped verifications to measure the hit rate.
+				if r.pool.Contains(r.RequestID(reqs[j])) {
+					continue
```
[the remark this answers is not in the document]
HagarMeir: This might actually be a bug in ResubmitPendingBAFs
Meaning we should verify the request before submitting here if the config seq was changed 
So this will be fixed soon

node/batcher/requests_inspector_verifier.go
```diff
@@ -105,6 +114,14 @@ func (r *RequestsInspectorVerifier) VerifyBatchedRequests(reqs types.BatchedRequ
 				if workerID != j%numWorkers {
 					continue
 				}
+				// A request already present in the mem pool was verified when it entered
+				// the pool (by this party's router, or via VerifyRequest on the batcher
+				// forward path) and remains valid under the current config (the pool is
+				// pruned and re-verified on reconfiguration). Skip re-verifying it.
+				// TODO: maybe add a metric counting skipped verifications to measure the hit rate.
+				if r.pool.Contains(r.RequestID(reqs[j])) {
```
tock-ibm, line 122: This calculated the request ID for the first time. In order to prevent recalculating it over and over, we could create here a slice `reqsIDs` with `len(reqs)`, which contains the computed request IDs, and pass it along `reqs` in all subsequent steps.

HagarMeir: ok 
with your permission I will do this in a different PR

node/batcher/requests_inspector_verifier.go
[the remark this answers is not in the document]
HagarMeir: Fixed. ResubmitPendingBAFs now re-verifies requests via VerifyRequest before resubmitting them to the pool whenever the pending BAF's config sequence is older than the current one (i.e. the policy may have changed). Requests that fail the current policy are dropped rather than reinserted, so they never become eligible for the pull-path skip. This runs after the new config has been applied (the verifier and config-sequence getter already reflect it), so the "in pool => verified under current config" invariant holds. Added TestResubmitPendingBAFsReverifiesOnConfigChange covering the reconfiguration case.

node/batcher/requests_inspector_verifier.go
tock-ibm: You can add a TODO for this and implement this in the future.

HagarMeir: no problem 
I will even open an issue 
just no reason to edit now if there are no other comments

HagarMeir: Here: https://github.com/hyperledger/fabric-x-orderer/issues/1185

tock-ibm (approved):

PR #1180, opened 2026-09-02
Remove unused ledger code

DorKatzelnick:
#1181


common/ledger/blkstorage/blockindex.go
```diff
@@ -30,17 +28,12 @@ const (
...
-
-	snapshotFileFormat       = byte(1)
-	snapshotDataFileName     = "txids.data"
-	snapshotMetadataFileName = "txids.metadata"
 )
 
 var (
 	indexSavePointKey              = []byte(indexSavePointKeyStr)
 	errIndexSavePointKeyNotPresent = errors.New("NoBlockIndexed")
 	errNilValue                    = errors.New("")
```
tock-ibm, line 36: `errNilValue` is now effectively dead. `getTxIDVal` still returns it on empty value bytes, but the three callers in `blockfile_mgr.go` (`retrieveBlockByTxID`, `retrieveTransactionByID`, `retrieveTxValidationCodeByTxID`) that used to translate it into a descriptive "details for the TXID not available" message were removed. If an empty-value txID index entry is ever hit, callers now surface `errors.New("")` (blank message) instead of a diagnosable error.

Empty-value entries only ever arose from snapshot import — which this PR removes — so the path is unreachable in practice. But the sentinel and its return path in `getTxIDVal` should have been removed along with snapshot support rather than left returning a blank error.

🤖 [Claude Code](https://claude.com/claude-code)

common/ledger/blkstorage/blockfile_mgr.go
```diff
@@ -520,19 +451,17 @@ func (mgr *blockfileMgr) retrieveBlockByHash(blockHash []byte) (*common.Block, e
 	return mgr.fetchBlock(loc)
 }
 
+// TODO: when ledger pruning lands, the reads keyed by block number need a front-boundary guard again:
```
tock-ibm, line 454: These restore-on-pruning TODOs (here and the `syncIndex` rebuild-scan one at line 333) correctly flag that the removed front-boundary guards must return when ledger pruning lands — otherwise pruned blocks would surface as generic index-miss errors and the rebuild scan would wrongly assume `blockfile_000000` still exists. A plain comment is easy to lose across the gap until pruning work starts; consider linking a tracking issue so the requirement isn't carried unaided.

🤖 [Claude Code](https://claude.com/claude-code)

tock-ibm (commented):
Code review (`high`): this is a clean removal PR — build, vet, and `./common/ledger/...` tests all pass, with no dangling references to any removed snapshot symbol. No correctness regressions; every removed guard depended on snapshot-bootstrap state an orderer never sets. Two minor cleanup notes inline.

🤖 Review by [Claude Code](https://claude.com/claude-code)

common/ledger/blkstorage/blockfile_mgr.go
tock-ibm: We have an issue for that.

common/ledger/blkstorage/blockindex.go
tock-ibm: It is fine for now. 
@DorKatzelnick lets re-evaluate the necessity of the variable after progressing further in this epic.

tock-ibm (approved):

PR #1184, opened 2026-09-02
Tx validation part2

moran-abilea:
related to https://github.com/hyperledger/fabric-x-orderer/issues/925

As a continuation for both the determistic and the random failure tests, this commit focus on using he submit funtionality inside those  failure tests.

it also improve the visibilty of the output for those tests as well


moran-abilea:
TODO: since we changed the assemblers heghits tests, we need to fix this test again later

Issue #349, opened 2025-11-11
Request throttling in the routers


tock-ibm:
An Adversary could launch a Denial-of-Service attack against the Router by sending the Router an overwhelming amount of traffic from one or many sources.

The mechanism should throttle the total number of request ingested by a router so that the system as a whole maintains "reasonable" performance.
The router should also divide the request budget "fairly" among clients, as identified by their certificates. This tries to prevent a small number of hungry clients form starving the rest.

Reference implementation: see fabric v3 orderer.

tock-ibm:
# Overview: Orderer Broadcast Throttling in hyperledger/fabric v3

## Overview

The orderer exposes a `Broadcast` gRPC streaming API through which clients submit
transactions. Without any protection, a single greedy client (or a single
organization) can monopolize the ordering node's ingress capacity, starving
other, well-behaved submitters.

Throttling addresses this by bounding the *aggregate* transaction ingress rate of
the node and dividing that budget **fairly among the currently active
submitters**. It is a fairness mechanism first and a rate cap second: the
configured rate is the ceiling for all clients combined, and each active client
receives an equal share of it.

Two independent limiters are enforced simultaneously:

| Limiter    | Identity key                                | Purpose                                       |
| ---------- | ------------------------------------------- | --------------------------------------------- |
| Per-client | hex of the TLS certificate `SubjectKeyId`   | Fairness between individual client identities |
| Per-org    | hex of the TLS certificate `AuthorityKeyId` | Fairness between issuing CAs / organizations  |

Because the identity is derived from the client's TLS certificate, throttling is
only enforced when the client connects with **mutual TLS**. Connections without
a peer certificate bypass throttling entirely.

Introduced in commit `5cb00bef7` ("Throttle greedy clients in orderer Broadcast API").

---

## Configuration

Defined in [config.go:157-164](../orderer/common/localconfig/config.go#L157-L164):

```go
// Throttling defines a max rate of transactions per client.
// The effective rate per client is the rate defined divided equally
// by all clients, until the clients cease from sending transactions
// and inactivity timeout expires for them.
type Throttling struct {
	// Rate is the maximum rate for all clients combined.
	Rate int
	// InactivityTimeout defines the time frame after which
	// inactive clients are pruned from memory and are not considered
	// when allocating the budget for throttling per client.
	InactivityTimeout time.Duration
}
```

[...]

PR #1188, opened 2026-09-03
node/router: add global request throttling (rate limiter)

tock-ibm:
Implements #1189 — **Goal 1 of #349**: a global (system-wide, aggregate) rate limiter on the router's request ingress.

## What
- **Lock-free token bucket** (`common/ratelimit`, GCRA): entire state is a single atomic int64, one CAS per admission — no mutex, ticker, or goroutine. A nil `*Limiter` means unlimited/disabled, so the disabled path is a free nil check.
- **Router enforcement**: a `RateLimiter` interface + a `throttler` policy container + a factory (`disabled`/`global`), enforced in `Broadcast`/`SubmitStream`/`Submit`, held in an `atomic.Pointer` (lock-free hot path, rebuilt on reconfig).
- **Immediate rejection** (unlike the Fabric v3 reference, which back-pressures): over-budget requests get `SERVICE_UNAVAILABLE` (503) on Broadcast / the throttle error on Submit, and the stream stays open. New `router_requests_throttled` metric.
- **Config**: `Router.Throttling {Policy, Rate, Burst}` in per-node local config; a missing block defaults to disabled.

## Design notes
- Target ingress is 500k req/s. Benchmarks show ~4.7M admits/s aggregate under full CAS contention, and a read-only, linearly-scaling reject path (so DoS floods are the cheapest case) — roughly 10x headroom.
- The config schema and the `RateLimiter`/policy seams are designed so the per-client/per-org policy (Goal 2) and alternative limiter implementations slot in **without a config break**.

## Testing
- `common/ratelimit` and `node/router` suites pass with `-race`; unit tests cover burst/refill/rate-convergence/concurrency, factory/policy dispatch, and router-level 503/disabled behavior.

This PR is primarily to exercise CI (linters + build flow).

### DIFF
--- a/common/ratelimit/gcra.go
+++ b/common/ratelimit/gcra.go
@@ -1,0 +1,82 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package ratelimit provides a lock-free token-bucket rate limiter suitable for
+// hot paths that must admit or reject requests at very high rates with minimal
+// contention.
+package ratelimit
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+// Limiter is a token-bucket rate limiter implemented with the Generic Cell Rate
+// Algorithm (GCRA). Its entire state is a single atomic int64 (the "theoretical
+// arrival time", TAT, in monotonic nanoseconds), so an admission decision is one
+// atomic load and, when admitted, one compare-and-swap. There is no mutex, no
+// background goroutine and no ticker: contention is bounded by GOMAXPROCS rather
+// than by the number of concurrent callers.
+//
+// A nil *Limiter is a valid value meaning "unlimited": callers gate the hot path
+// with a single `lim != nil` check, so a disabled limiter costs nothing.
+//
+// A Limiter is safe for concurrent use by multiple goroutines.
+type Limiter struct {
+	tatNanos         atomic.Int64 // theoretical arrival time, in the nowFn timeline
+	intervalNanos    int64        // nanoseconds per token = 1e9 / rate
+	burstOffsetNanos int64        // intervalNanos * burst (how far TAT may lag "now")
+	nowFn            func() int64 // monotonic clock; time.Since(base) in production
+}
+
+// New returns a Limiter admitting an average of rate requests per second with a
+// bucket capacity of burst tokens. It returns nil (meaning "unlimited", i.e.
+// throttling disabled) when rate <= 0, or when rate is so large that the
+// per-token interval rounds down to zero nanoseconds. A burst < 1 is coerced to 1.
+func New(rate float64, burst int) *Limiter {
+	base := time.Now()
+	return newWithClock(rate, burst, func() int64 { return int64(time.Since(base)) })
+}
+
+// newWithClock is New with an injectable monotonic clock, used by tests to make
+// time deterministic. nowFn must return nanoseconds from a fixed, monotonic base.
+func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
+	if rate <= 0 {
+		return nil
+	}
+	interval := int64(float64(time.Second) / rate)
+	if interval <= 0 {
+		// Rate is effectively unlimited at nanosecond resolution.
+		return nil
+	}
+	if burst < 1 {
+		burst = 1
+	}
+	return &Limiter{
+		intervalNanos:    interval,
+		burstOffsetNanos: interval * int64(burst),
+		nowFn:            nowFn,
+	}
+}
+
+// Allow reports whether one request may proceed now. It never blocks: an
+// over-budget request is rejected immediately (returning false). Rejections do
+// not mutate state (no CAS), so they are cheaper than admissions.
+func (l *Limiter) Allow() bool {
+	now := l.nowFn()
+	for {
+		tat := l.tatNanos.Load()
+		newTat := max(tat, now) + l.intervalNanos
+		if now < newTat-l.burstOffsetNanos {
+			return false // over budget
+		}
+		if l.tatNanos.CompareAndSwap(tat, newTat) {
+			return true
+		}
+		// Lost the race with a concurrent admission; retry. Retries are bounded
+		// by the number of goroutines actually running (GOMAXPROCS).
+	}
+}
--- a/common/ratelimit/gcra.go
+++ b/common/ratelimit/gcra.go
@@ -53,7 +53,15 @@ type Limiter struct {
 // New returns a Limiter admitting an average of rate requests per second with a
 // bucket capacity of burst tokens. It returns nil (meaning "unlimited", i.e.
 // throttling disabled) when rate <= 0, or when rate is so large that the
-// per-token interval rounds down to zero nanoseconds. A burst < 1 is coerced to 1.
+// per-token interval rounds down to zero nanoseconds.
+//
+// A burst < 1 is coerced to 1. This is only a defensive floor — it prevents a
+// zero-capacity, reject-everything limiter (burst 0 => negative tolerance); it
+// is not the intended "unset burst" default. That default belongs to the
+// caller: the router defaults an omitted Burst to Rate in its local config (see
+// config.applyNodeDefaults and the Router.Throttling.Burst docs), because a
+// burst of 1 has zero bunching tolerance and would throttle bursty ingress well
+// below the configured rate. New should therefore normally receive a burst >= 1.
 func New(rate float64, burst int) *Limiter {
 	base := time.Now()
 	return newWithClock(rate, burst, func() int64 { return int64(time.Since(base)) })
--- a/common/ratelimit/gcra.go
+++ b/common/ratelimit/gcra.go
@@ -22,6 +22,23 @@ import (
 // background goroutine and no ticker: contention is bounded by GOMAXPROCS rather
 // than by the number of concurrent callers.
 //
+// Classical GCRA keeps a theoretical arrival time (TAT) and, for an emission
+// interval T = 1/rate and a tolerance tau, admits a request arriving at "now"
+// when now >= TAT - tau, then advances TAT to max(TAT, now) + T. tau is a
+// duration — how far ahead of the theoretical schedule a request may run — so
+// the number of requests admissible back to back is tau/T + 1.
+//
+// This implementation is parameterised instead by an integer burst — the bucket
+// capacity, i.e. the number of requests admissible at a single instant — with
+// tau = (burst-1)*T. It also folds the emission increment into the test: it
+// computes newTAT = max(TAT, now) + T and rejects when now < newTAT - burstOffset,
+// where burstOffset = burst*T = tau + T. The extra T inside newTAT and inside
+// burstOffset cancel, so this is exactly the classical "now < TAT - tau" in the
+// binding (TAT > now) case; the single max(...)+T form just lets the idle and
+// bunched branches — and the CAS — share one value. So burst 1 => tau 0 => the
+// steady rate with no bunching, burst N => N admitted at once, and burst < 1
+// (which would be a negative tau, rejecting everything) is coerced to 1 in New.
+//
 // A nil *Limiter is a valid value meaning "unlimited": callers gate the hot path
 // with a single `lim != nil` check, so a disabled limiter costs nothing.
 //
@@ -29,7 +46,7 @@ import (
 type Limiter struct {
 	tatNanos         atomic.Int64 // theoretical arrival time, in the nowFn timeline
 	intervalNanos    int64        // nanoseconds per token = 1e9 / rate
-	burstOffsetNanos int64        // intervalNanos * burst (how far TAT may lag "now")
+	burstOffsetNanos int64        // burst*interval; classical tolerance tau = (burst-1)*interval
 	nowFn            func() int64 // monotonic clock; time.Since(base) in production
 }
 
--- a/common/ratelimit/gcra.go
+++ b/common/ratelimit/gcra.go
@@ -6,10 +6,11 @@ SPDX-License-Identifier: Apache-2.0
 
 // Package ratelimit provides a lock-free token-bucket rate limiter suitable for
 // hot paths that must admit or reject requests at very high rates with minimal
-// contention.
+// contention. It implements the Generic Cell Rate Algorithm (GCRA).
 package ratelimit
 
 import (
+	"math"
 	"sync/atomic"
 	"time"
 )
@@ -55,9 +56,14 @@ func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
 	if burst < 1 {
 		burst = 1
 	}
+	burstOffset := interval * int64(burst)
+	if burstOffset/int64(burst) != interval {
+		// Multiplication overflowed int64; treat as an effectively unbounded burst.
+		burstOffset = math.MaxInt64
+	}
 	return &Limiter{
 		intervalNanos:    interval,
-		burstOffsetNanos: interval * int64(burst),
+		burstOffsetNanos: burstOffset,
 		nowFn:            nowFn,
 	}
 }
@@ -66,6 +72,9 @@ func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
 // over-budget request is rejected immediately (returning false). Rejections do
 // not mutate state (no CAS), so they are cheaper than admissions.
 func (l *Limiter) Allow() bool {
+	if l == nil {
+		return true // a nil Limiter is "unlimited"
+	}
 	now := l.nowFn()
 	for {
 		tat := l.tatNanos.Load()
--- a/common/ratelimit/gcra.go
+++ b/common/ratelimit/gcra.go
@@ -53,7 +53,10 @@ type Limiter struct {
 // New returns a Limiter admitting an average of rate requests per second with a
 // bucket capacity of burst tokens. It returns nil (meaning "unlimited", i.e.
 // throttling disabled) when rate <= 0, or when rate is so large that the
-// per-token interval rounds down to zero nanoseconds.
+// per-token interval rounds down to zero nanoseconds. A rate below the slowest
+// supported value (~one request per day) is clamped up to that minimum, so a
+// tiny rate can't overflow the interval conversion and be mistaken for
+// "unlimited".
 //
 // A burst < 1 is coerced to 1. This is only a defensive floor — it prevents a
 // zero-capacity, reject-everything limiter (burst 0 => negative tolerance); it
@@ -67,13 +70,25 @@ func New(rate float64, burst int) *Limiter {
 	return newWithClock(rate, burst, func() int64 { return int64(time.Since(base)) })
 }
 
+// maxIntervalNanos caps the per-token interval so the slowest rate New supports
+// is roughly one request per day. A smaller rate is clamped up to this, keeping
+// the interval math well clear of int64 overflow.
+const maxIntervalNanos = int64(24 * time.Hour)
+
 // newWithClock is New with an injectable monotonic clock, used by tests to make
 // time deterministic. nowFn must return nanoseconds from a fixed, monotonic base.
 func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
 	if rate <= 0 {
 		return nil
 	}
-	interval := int64(float64(time.Second) / rate)
+	intervalFloat := float64(time.Second) / rate
+	if intervalFloat > float64(maxIntervalNanos) {
+		// Rate is below the slowest supported value (~once per day); clamp to that
+		// minimum so a tiny rate can't overflow the int64 conversion and be
+		// mistaken for "unlimited".
+		intervalFloat = float64(maxIntervalNanos)
+	}
+	interval := int64(intervalFloat)
 	if interval <= 0 {
 		// Rate is effectively unlimited at nanosecond resolution.
 		return nil
--- a/config/local_config.go
+++ b/config/local_config.go
@@ -166,7 +166,29 @@ type RouterParams struct {
 	NumberOfConnectionsPerBatcher int `yaml:"NumberOfConnectionsPerBatcher,omitempty"`
 	// NumberOfStreamsPerConnection specifies the number of streams per connection that are opened between Router and Batcher
 	NumberOfStreamsPerConnection int `yaml:"NumberOfStreamsPerConnection,omitempty"`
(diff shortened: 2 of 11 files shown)


node/router/router.go
```diff
@@ -402,6 +411,12 @@ func (r *Router) Broadcast(stream orderer.AtomicBroadcast_BroadcastServer) error
 
 		r.metrics.incomingTxs.Add(1)
 
+		if !r.throttler.Load().Allow() {
+			r.metrics.throttledTxs.Add(1)
+			r.sendBroadcastResponse(stream, Response{err: ErrThrottled})
```
DorKatzelnick: push the throttle response into feedbackChan

according to go grpc, "it is not safe to call SendMsg on the same stream in different goroutines"

node/router/router.go
```diff
@@ -453,6 +468,12 @@ func (r *Router) SubmitStream(stream protos.RequestTransmit_SubmitStreamServer)
 
 		reqID, shardRouter := r.getShardRouterAndReqID(req)
 
+		if !r.throttler.Load().Allow() {
+			r.metrics.throttledTxs.Add(1)
+			r.sendSubmitResponse(stream, Response{err: ErrThrottled, reqID: reqID})
```
DorKatzelnick: same

node/router/router.go
DorKatzelnick: also we can change this if we are here already

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,82 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package ratelimit provides a lock-free token-bucket rate limiter suitable for
```
DorKatzelnick, line 7: what gcra stands for?

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// Limiter is a token-bucket rate limiter implemented with the Generic Cell Rate
+// Algorithm (GCRA). Its entire state is a single atomic int64 (the "theoretical
+// arrival time", TAT, in monotonic nanoseconds), so an admission decision is one
+// atomic load and, when admitted, one compare-and-swap. There is no mutex, no
+// background goroutine and no ticker: contention is bounded by GOMAXPROCS rather
+// than by the number of concurrent callers.
+//
+// A nil *Limiter is a valid value meaning "unlimited": callers gate the hot path
```
DorKatzelnick, line 42: Allow() will panic on nil *Limiter

check
if l == nil { return true } at the top of Allow

config/local_config.go
```diff
@@ -569,6 +591,17 @@ func applyNodeDefaults(nodeLocalConfig *NodeLocalConfig, role string, logger *fl
 			logger.Infof("Router.NumberOfStreamsPerConnection is not set, using default value: %d", nodeLocalConfig.RouterParams.NumberOfStreamsPerConnection)
 		}
 
+		if t := nodeLocalConfig.RouterParams.Throttling; t != nil {
+			if t.Policy == "" {
+				t.Policy = ThrottlingPolicyDisabled
+				logger.Infof("Router.Throttling.Policy is not set, using default value: %q", t.Policy)
+			}
+			if t.Policy == ThrottlingPolicyGlobal && t.Rate > 0 && t.Burst == 0 {
```
DorKatzelnick, line 614: need to validate that when "global" is set, also "Rate" is non-zero?

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,82 @@
...
+	if interval <= 0 {
+		// Rate is effectively unlimited at nanosecond resolution.
+		return nil
+	}
+	if burst < 1 {
+		burst = 1
+	}
+	return &Limiter{
+		intervalNanos:    interval,
+		burstOffsetNanos: interval * int64(burst),
```
DorKatzelnick: needs overflow guard

node/router/router.go
```diff
@@ -453,6 +468,12 @@ func (r *Router) SubmitStream(stream protos.RequestTransmit_SubmitStreamServer)
 
 		reqID, shardRouter := r.getShardRouterAndReqID(req)
```
DorKatzelnick: call getShardRouterAndReqID after allow()

node/router/router.go
```diff
@@ -495,6 +516,11 @@ func (r *Router) Submit(ctx context.Context, request *protos.Request) (*protos.S
 
 	reqID, shardRouter := r.getShardRouterAndReqID(request)
```
DorKatzelnick: call getShardRouterAndReqID after allow()

DorKatzelnick:
more comments by claude:

### Config-update transactions are throttled too

Request classification happens downstream, at `node/router/stream.go:91`, long after the throttle gate. So a `HeaderType_CONFIG_UPDATE` envelope is indistinguishable from a normal TX at admission time and gets a 503 along with everything else — including, under a flood, the reconfiguration an admin would use to respond to that flood.

Probably acceptable for Goal 1, but I'd rather see it as a recorded decision (a comment at the gate, or a line in the PR description) than as an implicit consequence of where the check landed.

### The policy-name constants are defined twice

`ThrottlingPolicyDisabled`/`ThrottlingPolicyGlobal` in `config/local_config.go` and `ThrottlingDisabled`/`ThrottlingGlobal` in `node/router/throttle.go` — same string values, two sources of truth, kept in sync only by the mutually-referencing comments. `extractRouterThrottlingConfig` writes the config-package value; `newThrottler` switches on the router-package one, so a drift between them is a `Panicf` at startup.

Both packages already import `node/config`, which is where `RouterThrottlingConfig` lives. That's the natural single home for them.

### Two different `Burst` defaults at two layers

`applyNodeDefaults` says `Burst: 0` → `Rate` (one second of tokens). `ratelimit.New` says `burst < 1` → `1`. Any path that builds a `RouterNodeConfig` without going through `applyNodeDefaults` — tests, `testutil/config_utils.go` — silently gets a burst of 1 rather than `Rate`, which is a materially different limiter. Worth picking one layer to own the default and having the other leave it alone.

### The reconfig rebuild is both a no-op and a state reset

`initFromConfig` rebuilds the throttler, and `ApplyConfig` calls it (`node/router/router.go:772`). But `Throttling` comes from local config, which a config block can't change — so the rebuild can never actually change the policy, while it does discard the token-bucket state and hand the client a fresh full burst on every config change.

The `atomic.Pointer` is the right primitive if per-client limits later arrive via shared config, so I'm not asking to remove it. But the comments ("rebuilt on reconfig", "fails fast at startup — and on reconfig") describe more than currently happens, and the free burst per reconfig is a small surprise worth either fixing or noting.

### Tests

- **Convention:** the three new tests in `router_test.go` use prose doc comments, while the tests immediately below them use the numbered `// Scenario:` step lists that the rest of the file follows. Worth matching the file they're in.
- `regexp.MustCompile("throttled").MatchString(err.Error())` compiles a regexp per iteration to look for a fixed substring — `strings.Contains` does it.
- **Coverage gaps**, roughly in order of how much I'd want them:
  - The config layer is untested — `config/local_config_test.go` is untouched, so nothing covers the YAML round-trip of the new `Throttling` block or the `Burst: 0` → `Rate` defaulting.
  - Nothing covers the `atomic.Pointer` swap on reconfig, which is the one piece of concurrency machinery added to `Router` itself.
  - `SubmitStream` has a gate but no test; only `Broadcast` and unary `Submit` are covered.
  - The startup `Panicf` on an unknown policy isn't covered — only that `newThrottler` returns an error.

### The new knob is hard to discover

`Throttling` isn't added to `config/sample/local_config_router.yaml` or `config/sample/local_config.yaml`, which are the de-facto reference for `Router` params, and `armageddon generate` doesn't emit it. Defaulting it off is right; but as it stands an operator has to read the Go structs to learn the option exists.

node/router/router.go
tock-ibm: Fixed. The throttle rejection is now sent via feedbackChan, so the feedback goroutine is the sole caller of stream.Send — no concurrent SendMsg on the same stream.

_Assisted by Claude Code._

node/router/router.go
tock-ibm: Fixed the same way — the SubmitStream throttle rejection now goes through feedbackChan instead of a direct stream.Send.

_Assisted by Claude Code._

node/router/router.go
tock-ibm: Done. The stopChan responses in both Broadcast and SubmitStream now also go through feedbackChan. With no direct sends left on the main goroutine, sendBroadcastResponse/sendSubmitResponse were removed.

_Assisted by Claude Code._

common/ratelimit/gcra.go
tock-ibm: Generic Cell Rate Algorithm — expanded it in the package doc comment.

_Assisted by Claude Code._

common/ratelimit/gcra.go
tock-ibm: Added `if l == nil { return true }` at the top of Allow(), so a nil *Limiter now means "unlimited" as the type doc states.

_Assisted by Claude Code._

config/local_config.go
tock-ibm: newThrottler now returns an error when policy is "global" and Rate <= 0 (previously it ran silently disabled), so the misconfiguration fails fast at startup — consistent with how an unknown policy is rejected.

_Assisted by Claude Code._

common/ratelimit/gcra.go
tock-ibm: Added an overflow guard in newWithClock: interval*burst is verified with a divide-back check and clamped to MaxInt64 on overflow, so a large burst cannot wrap negative and flip the limiter into rejecting everything.

_Assisted by Claude Code._

node/router/router.go
tock-ibm: Done. getShardRouterAndReqID now runs after Allow(), so a throttled request skips the shard mapping (CRC64). The throttle rejection no longer carries reqID, matching Broadcast.

_Assisted by Claude Code._

node/router/router.go
tock-ibm: Done. Same reorder in Submit — getShardRouterAndReqID after Allow(); the throttle rejection drops reqID for consistency with Broadcast/SubmitStream.

_Assisted by Claude Code._

tock-ibm:
> Config-update transactions are throttled too

Correct, will be addressed in a future PR.

tock-ibm:
> The policy-name constants are defined twice

Addressed.

tock-ibm:
> Two different `Burst` defaults at two layers

Addressed

tock-ibm:
> Tests
> The new knob is hard to discover

Will be addressed in the next PR

config/local_config.go
```diff
@@ -166,8 +166,36 @@ type RouterParams struct {
...
+// ThrottlingParams configures the router's request throttling. The Policy field
+// selects the strategy and keeps the config schema stable as new strategies are
+// added (e.g. per-client/per-org in a later phase); Rate and Burst parameterize
+// the rate limiter.
+type ThrottlingParams struct {
+	// Policy selects the throttling strategy: "disabled" (default) or "global".
+	Policy string `yaml:"Policy,omitempty"`
+	// Rate is the aggregate cap in requests/second across all clients (global policy).
+	Rate int `yaml:"Rate,omitempty"`
+	// Burst is the token-bucket capacity: the maximum number of requests that may
```
DorKatzelnick, line 182: **The `Burst` fix picked the wrong default.** This is the one I'd block on.

`de5caf9a` resolved "two layers own the default" by deleting the `Burst: 0` -> `Rate` defaulting from `applyNodeDefaults` and letting `ratelimit.New`'s `burst < 1` -> `1` win. One owner, as asked. But burst 1 means tau 0 — *zero* tolerance for bunching — and real ingress is always bunched (many concurrent streams, roughly Poisson arrivals).

Simulating Poisson arrivals against the limiter at `Rate: 500000`:

| Offered load | burst=1 | burst=10 | burst=1000 | burst=Rate |
|---|---|---|---|---|
| = the cap | **50.0%** admitted | 94.9% | 99.9% | 100% |
| = half the cap | **66.7%** admitted | 100% | 100% | 100% |

So an operator who sets `Policy: global, Rate: 500000` and omits `Burst` gets a limiter that throttles a third of traffic at *half* the configured cap. The closed form is `e^{-lambda*T}`: burst 1 admits only requests spaced at least T apart.

The `Burst -> Rate` default was correct; the duplication was the only problem. Keep one owner, but give that owner a sane default — `Rate`, or at least a modest floor like 10 (the table shows most of the loss is gone by then).

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,108 @@
...
+func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
+	if rate <= 0 {
+		return nil
+	}
+	interval := int64(float64(time.Second) / rate)
+	if interval <= 0 {
+		// Rate is effectively unlimited at nanosecond resolution.
+		return nil
+	}
+	if burst < 1 {
```
DorKatzelnick, line 96: This is now the sole owner of the `Burst` default, which is the right structure — but `1` is a bad value to default to for a request-ingress limiter. See the comment on `config/local_config.go` for the measured admit rates; burst 1 rejects ~50% of traffic offered at exactly the configured cap.

Coercing `burst < 1` to `1` is still correct as a *floor* (a negative tau would reject everything). The question is what an unset `Burst` should mean, and `1` shouldn't be it.

DorKatzelnick (commented):
Re-reviewed at `de5caf9a`. All nine inline comments are genuinely fixed, and the constant duplication is resolved. Build, `go vet`, and `-race` on `common/ratelimit` + the router throttle tests all pass.

Two things I'd want addressed before merge — one is a regression introduced by the `Burst` fix itself, one is new fallout from the `stream.Send` -> `feedbackChan` conversion — plus a convention nit and three minors. Details inline.

Also, the PR description's "**Immediate rejection** (unlike the Fabric v3 reference, which back-pressures)" is no longer accurate: with the rejection routed through `feedbackChan`, a client that floods without reading its responses fills the 1000-slot buffer and then back-pressures the `Recv` loop. That's arguably the better behavior — the description should just say so.

Tracking as deferred per your replies: config-update TXs being throttled, the test gaps, and the sample YAML / `armageddon generate` discoverability.

_Assisted by Claude Code._

node/router/router.go
```diff
@@ -402,15 +411,21 @@ func (r *Router) Broadcast(stream orderer.AtomicBroadcast_BroadcastServer) error
...
+			continue
+		}
+
 		request := &protos.Request{Payload: reqEnv.Payload, Signature: reqEnv.Signature, ConfigSeq: r.configSeq}
 		reqID, shardRouter := r.getShardRouterAndReqID(request)
 
 		select {
 		case <-r.stopChan:
-			r.sendBroadcastResponse(stream, Response{
+			feedbackChan <- Response{
```
DorKatzelnick: **New: this blocking send can hang the stream-read goroutine forever.**

Routing through `feedbackChan` is the right fix for the concurrent `SendMsg`, but this is a bare blocking send on the goroutine that reads the stream, and the reader can already be gone:

`SoftStop` closes `drainChan`, then `feedbackWG.Wait()`. Each feedback goroutine returns as soon as `drainChan` is closed and the channel is empty. A `Broadcast` handler still in its loop then takes `case <-r.stopChan:` and pushes into a channel nobody reads — 1000 silently-dropped "router is stopping" responses (100 for `SubmitStream`), and then the handler blocks forever. At 500k req/s, 1000 responses is ~2 ms, so this is reachable in an ordinary soft stop under load.

Before this PR the direct `stream.Send` failed harmlessly once `net.Stop()` closed the connection, so the client at least got the response and the goroutine exited. Two regressions here: the responses `SoftStop` exists to deliver are dropped, and the handler leaks.

For this branch specifically, `return`ing from the handler after queuing seems right — the router is stopping, there's no reason to keep reading the stream.

node/router/router.go
```diff
@@ -402,15 +411,21 @@ func (r *Router) Broadcast(stream orderer.AtomicBroadcast_BroadcastServer) error
 
 		r.metrics.incomingTxs.Add(1)
 
+		if !r.throttler.Load().Allow() {
+			r.metrics.throttledTxs.Add(1)
+			feedbackChan <- Response{err: ErrThrottled}
```
DorKatzelnick, line 416: Same blocking-send concern as the `stopChan` branch below — after the feedback goroutine has exited via `drainChan`, this send has no reader and will block once the buffer fills.

Here I'd suggest a non-blocking send: `select { case feedbackChan <- Response{err: ErrThrottled}: default: }`. Dropping a 503 under extreme overload is defensible and it preserves the "never block the ingress path" property the design notes claim.

config/local_config.go
```diff
@@ -166,8 +166,36 @@ type RouterParams struct {
...
+}
+
+// Throttling policy names accepted by ThrottlingParams.Policy (and carried
+// through to the router's RouterThrottlingConfig.Policy). This is the single
+// source of truth for these values, shared by the config layer (which reads them
+// from YAML) and the router (which dispatches on them).
+const (
+	ThrottlingPolicyDisabled = "disabled" // no throttling (default)
+	ThrottlingPolicyGlobal   = "global"   // one aggregate rate limit across all clients
+	// Future: per-client / per-org policies (issue #349, Goal 2).
```
DorKatzelnick: The convention here is no issue or task IDs in code or comments — describe the mechanism instead. Something like "Future policies may key the limit on client or org identity" carries the same information and doesn't rot when the tracker changes.

node/router/router.go
```diff
@@ -451,14 +466,20 @@ func (r *Router) SubmitStream(stream protos.RequestTransmit_SubmitStreamServer)
...
```
DorKatzelnick: Minor: dropping `reqID` for consistency with `Broadcast` makes this the one `SubmitStream` response a client can't correlate to a request — `Broadcast` has no `ReqID` field at all, but `SubmitResponse` does, and on a stream that's how correlation works.

Nothing in the tree actually reads `SubmitResponse.ReqID` outside `node/router/stream_test.go:173`, so this is cosmetic today. Just worth a word saying it's deliberate.

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,108 @@
...
+		burst = 1
+	}
+	burstOffset := interval * int64(burst)
+	if burstOffset/int64(burst) != interval {
+		// Multiplication overflowed int64; treat as an effectively unbounded burst.
+		burstOffset = math.MaxInt64
```
DorKatzelnick, line 102: Minor: clamping `burstOffset` rather than `burst` makes the limiter admit unconditionally, so `tat` advances by `interval` per admission with no bound relative to `now` — after roughly `MaxInt64/interval` admissions it wraps negative, `newTat - burstOffsetNanos` then overflows positive, and the limiter flips to rejecting everything permanently.

Only reachable with an absurd `Burst` (the overflow case implies a large `interval` too), so low priority. Clamping `burst` to `MaxInt64/interval` instead would close it without the second failure mode.

common/ratelimit/gcra.go
```diff
@@ -0,0 +1,108 @@
...
+	return newWithClock(rate, burst, func() int64 { return int64(time.Since(base)) })
+}
+
+// newWithClock is New with an injectable monotonic clock, used by tests to make
+// time deterministic. nowFn must return nanoseconds from a fixed, monotonic base.
+func newWithClock(rate float64, burst int, nowFn func() int64) *Limiter {
+	if rate <= 0 {
+		return nil
+	}
+	interval := int64(float64(time.Second) / rate)
```
DorKatzelnick: Minor, exported-API only: a tiny positive `rate` (say `1e-12`) overflows this `int64` conversion to `MinInt64`, which then trips the `interval <= 0` check and returns nil — i.e. *unlimited*, the exact inverse of the request. Unreachable from config since `Rate` is an `int`, but `New` takes a `float64`, so a future caller could hit it. Guarding the conversion against a non-positive/overflowed result before the `interval <= 0` branch would make the contract safe.

common/ratelimit/gcra.go
tock-ibm: We'll keep the CGRA with this simple limit, and in the local config we'll apply Burst=Rate. The owner is local config, and in CGRA it is just defensive computing.

config/local_config.go
tock-ibm: Done.

config/local_config.go
tock-ibm: What I changed (3 tasks):

1. Re-added the 0→Rate default in applyNodeDefaults ([local_config.go](vscode-webview://01qef1lcfqd76oplm55gonh5jdcjri95np5ocmrgpt5sipdl0536/config/local_config.go)) — config layer is once again the single meaningful owner, defaulting an omitted Burst to Rate.
2. Documented the default + rationale on the Burst field — bucket capacity, the deliberate → Rate default, and why (bursty/Poisson ingress; burst=1 = zero bunching tolerance ⇒ throttles well below the cap, ~1/3 dropped at half the cap per Dor's simulation). It also names applyNodeDefaults as the owner and calls the GCRA clamp defensive.
3. Documented the GCRA floor as defensive ([gcra.go](vscode-webview://01qef1lcfqd76oplm55gonh5jdcjri95np5ocmrgpt5sipdl0536/common/ratelimit/gcra.go) New) — burst < 1 → 1 is only a guard against a zero-capacity reject-all limiter; the real "unset burst" default lives in the router's local config.

common/ratelimit/gcra.go
tock-ibm: Done by defining a minimal rate that avoids overflow (~once/day).

node/router/router.go
tock-ibm: Correct. I switched the order and included reqID in the response.

common/ratelimit/gcra.go
tock-ibm: Moot. MaxInt64/interval x nanosec will not happen in our lifetime.

node/router/router.go
tock-ibm: I actually changed it back so that the response has the reqID

node/router/router.go
tock-ibm: I actually changed it back so that the response has the reqID

node/router/router.go
tock-ibm: blocking on the feedback channel actually puts back preasure on the offending client, so I want to keep that.
We'll deal with the drain signal leaving the channel without a reader in another PR.

tock-ibm: see #1209

DorKatzelnick (approved):

PR #1194, opened 2026-09-07
prometheus metrics reporting

moradna:


node/consensus/metrics.go
```diff
@@ -128,22 +146,37 @@ func (m *ConsensusMetrics) trackMetrics() {
...
 	defer t.Stop()
 
+	partyID := fmt.Sprintf("%d", m.partyID)
+
 	for {
 		select {
 		case <-t.C:
-			dec := uint64(monitoring.GetMetricValue(m.decisionsCount.(prometheus.Metric), m.logger))
-			blk := uint64(monitoring.GetMetricValue(m.blocksCount.(prometheus.Metric), m.logger))
+			reader := monitoring.NewReader(m.promAddress, m.interval)
```
tock-ibm, line 154: why a new reader on every tick and not one created before the loop?

common/monitoring/prometheus_reader.go
```diff
@@ -0,0 +1,149 @@
...
+
+var queryClient = &http.Client{Timeout: 5 * time.Second}
+
+type Reader struct {
+	address string
+	window  time.Duration
+	err     error
+}
+
+func NewReader(address string, window time.Duration) *Reader {
```
tock-ibm, line 31: What if the /metrics endpoint is configured with tls?

common/monitoring/prometheus_reader.go
```diff
@@ -0,0 +1,138 @@
...
+
+type Reader struct {
+	address string
+	window  time.Duration
+	client  *http.Client
+}
+
+// NewReader returns a Reader that queries the Prometheus server at address.
+// tlsConfig configures TLS for HTTPS connections.
+func NewReader(address string, window time.Duration, tlsConfig *tls.Config) *Reader {
```
tock-ibm: **Metric reporting now depends on an external Prometheus scrape, so logged values are `0` (plus warning spam) whenever no Prometheus is scraping this node.**

Previously these metrics were read from the in-process registry — self-contained and always accurate. Now every `Total`/`Gauge`/`Histogram*` call round-trips to the Prometheus server at `PrometheusAddress` (default `http://localhost:9090`).

_Failure:_ a node running with the default address but no Prometheus deployed/scraping it (dev, CI, most single-node/test setups) logs `0` for every counter/latency in the `*_METRICS` lines and emits repeated `Failed to read ...` warnings each interval. Even when Prometheus is present, values lag by up to one scrape interval, so `StopMetricsTracker`'s shutdown totals miss the final unscraped activity.

common/monitoring/prometheus_reader.go
```diff
@@ -0,0 +1,138 @@
...
+// HistogramIntervalAverage returns the histogram average over the last interval.
+func (r *Reader) HistogramIntervalAverage(opts metrics.HistogramOpts, labelValues ...string) (float64, error) {
+	sum, count := r.histogramSelectors(opts, labelValues)
+
+	window := fmt.Sprintf("%dms", r.window.Milliseconds())
+	if r.window%time.Second == 0 {
+		window = fmt.Sprintf("%ds", r.window/time.Second)
+	}
+
+	avg, err := r.query(fmt.Sprintf("rate(%s[%s]) / rate(%s[%s])", sum, window, count, window))
```
tock-ibm: **`HistogramIntervalAverage` uses the report interval as the `rate()` window; when Prometheus's scrape interval is ≥ ~half the report interval, `rate(x[window])` has <2 samples and returns empty**, so interval latency averages are always `0`.

_Failure:_ with the default `MetricsLogInterval=10s` and Prometheus's default global scrape interval of 15s, `rate(..._sum[10s])` sees at most one sample, returns no data, `query()` errors with `invalid Prometheus response`, and every `*_latency_avg_seconds` interval value is logged as `0` with a warning each cycle. The window needs to span several scrape intervals (e.g. `max(4*scrape, interval)`).

node/batcher/metrics.go
```diff
@@ -198,62 +202,217 @@ func (m *BatcherMetrics) StopMetricsTracker() {
...
 
 func (m *BatcherMetrics) trackMetrics() {
-	prevC := monitoring.GetMetricValue(m.batchesCreatedTotal.(prometheus.Metric), m.logger)
-	prevP := monitoring.GetMetricValue(m.batchesPulledTotal.(prometheus.Metric), m.logger)
-	prevR := float64(0)
 	sec := m.interval.Seconds()
+	labels := m.labels()
+
+	reader := monitoring.NewReader(m.promAddress, m.interval, m.promTLS)
+	prevC, err := reader.Total(batchesCreatedTotalOpts, labels...)
```
tock-ibm: **The first reported interval is inflated by the entire cumulative count.** `prevC`/`prevP` are read from Prometheus here at startup, before the just-started node has been scraped, so the read fails and they default to `0` — while the counters themselves are ledger-initialized to the current height.

_Failure:_ a batcher restarts at ledger height 100000. `prevC` reads as `0` (series not in Prometheus yet). On the first successful tick `created=100000`, so `batches_created_interval=100000` and `batches_created_rate=100000/interval` are logged as this-interval activity. The pre-PR code seeded `prevC` from the ledger-initialized registry value, so the first delta was ~0.

common/monitoring/prometheus_reader.go
```diff
@@ -0,0 +1,138 @@
...
+				Value [2]json.RawMessage `json:"value"`
+			} `json:"result"`
+		} `json:"data"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return 0, err
+	}
+
+	if response.Status != "success" || len(response.Data.Result) != 1 {
```
tock-ibm: **`query()` treats any result set whose length ≠ 1 as an error**, which is fragile once Prometheus attaches `job`/`instance` labels or more than one series matches a selector.

_Failure:_ once Prometheus scrapes the node it adds `instance`/`job` labels; with a second scrape job, relabeling, or a stale duplicate series after a restart, two series can match e.g. `router_requests_completed{party_id="1"}`. `len(Result)==2 != 1`, so the read errors and the value is reported as `0` despite valid data. Consider selecting/aggregating explicitly (take the last, or add a `by`-aggregation) rather than requiring exactly one series.

node/batcher/metrics.go
```diff
@@ -198,62 +202,217 @@ func (m *BatcherMetrics) StopMetricsTracker() {
...
+			mempoolNextLatency,
+			verifyLatency,
+			hashingLatency,
+			ledgerHashingLatency,
+			ledgerAppendLatency,
 		)
 	})
 }
 
 func (m *BatcherMetrics) trackMetrics() {
```
tock-ibm: **Each reporting cycle issues many sequential blocking HTTP GETs** (the batcher does ~14, each up to the 5s client timeout); if Prometheus is slow or unreachable a single cycle can block far longer than the tick interval and spam warnings.

_Failure:_ with Prometheus unreachable, this `trackMetrics` goroutine performs ~14 sequential reader calls, each blocking the full 5s timeout, so one cycle takes ~70s (≫ a 10s interval). Reporting falls badly behind and emits ~14 warnings per cycle. Previously all values were read instantly from the local registry.

config/local_config.go
```diff
@@ -485,6 +498,39 @@ func loadClusterCryptoConfig(cluster *ClusterYaml) (*Cluster, error) {
...
+
+	for _, rootCAPath := range tlsConfig.RootCAs {
+		rootCACert, err := utils.ReadPem(rootCAPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed load root ca certificate: %s", err)
+		}
+		secOpts.ServerRootCAs = append(secOpts.ServerRootCAs, rootCACert)
+	}
+
+	if tlsConfig.Certificate != "" || tlsConfig.PrivateKey != "" {
```
tock-ibm: **`loadMetricsTLSConfig` gates client cert/key loading on `Certificate != "" || PrivateKey != ""`**, so setting only one of the pair calls `ReadPem("")` on the missing file and fails with a confusing error instead of a clear both-or-neither validation.

_Failure:_ an operator sets `Metrics.TLS.Certificate` but leaves `PrivateKey` empty (or vice versa). The `||` enters the block, `ReadPem("")` on the empty path fails, and the node aborts startup with `failed load client private key`. Elsewhere this codebase uses an explicit XOR check (`(a=="") != (b=="")`) to report the real problem clearly.

node/batcher/metrics.go
```diff
@@ -198,62 +202,217 @@ func (m *BatcherMetrics) StopMetricsTracker() {
 	m.stopOnce.Do(func() {
 		close(m.stopChan)
 		m.logger.Infof("Reporting routine is stopping")
+
+		labels := m.labels()
+		reader := monitoring.NewReader(m.promAddress, m.interval, m.promTLS)
+
+		role, err := reader.Gauge(currentRoleOpts, labels...)
+		if err != nil {
+			m.logger.Warnf("Failed to read current role: %s", err)
```
tock-ibm: **On a transient read error here, several reported fields silently degrade.** `role` falls back to `0` → logged as `secondary` even on the primary; and the non-fallback totals (`batched_txs`, `mempool_size`, `router_txs`, `complaints`, etc.) are logged as `0` rather than the last known value.

_Failure:_ a single failed `currentRole` read on a primary batcher (e.g. a momentary Prometheus timeout) makes `reader.Gauge` return `0`, and `roleName(0)` logs `role="secondary"`, misrepresenting the node's role; simultaneously the other totals without prev-fallback drop to `0` for that log line, producing misleading dips in operational logs. Consider carrying forward the previous value on error for these fields too.

common/monitoring/prometheus_reader.go
```diff
@@ -0,0 +1,138 @@
+/*
```
tock-ibm: **This new file has no unit tests** — URL construction, query escaping, JSON decoding, selector/label formatting, the histogram sum/count and rate queries, and the strict single-result parsing are all untested.

A small `httptest.Server` stub returning canned `/api/v1/query` JSON would cover `Total`/`Gauge`/`HistogramAverage`/`HistogramIntervalAverage` and the error paths, and would guard against regressions in selector label ordering, namespace prefixing, and window formatting.

tock-ibm (commented):
## Automated code review — Prometheus metrics reporting

This PR replaces reading metrics from the in-process registry with querying an external Prometheus HTTP server. Build and `go vet` pass, and label-count/selector alignment is correct across all call sites (no panic). The inline comments below are behavioral/design defects, ordered most-severe first.

The overarching concern (first comment): metric reporting now depends on an external Prometheus actually scraping the node, so in any setup without one — dev, CI, most single-node/test deployments — every logged metric becomes `0` plus a burst of warnings, a regression from the previously self-contained registry reads.

PR #1195, opened 2026-09-07
CI: verify go.mod/go.sum and vendor are in sync

### LINKED ISSUE #1058: CI: verify go.mod/go.sum are tidy and vendor/ is in sync
## Summary

CI should fail if `go.mod`, `go.sum`, or the vendored `vendor/` directory are out of date. Today CI partially checks tidiness but does not verify the vendor directory, so a stale or inconsistent `vendor/` can be merged.

## Current state

- `scripts/check_deps.sh` copies `go.mod`/`go.sum` to a temp dir, runs `go mod tidy` on the copy, and diffs against the originals. This runs in CI via `make basic-checks` → `make linter` → `check-deps`.
- Its failure message already says *"Please run `go mod tidy` and `go mod vendor`"* — but nothing in CI actually verifies `go mod vendor`. The `vendor/` tree can drift from `go.mod`/`go.sum` and CI stays green.
- Dependencies are vendored (`vendor/`), and the project also has `check-deps` for unused vendored deps, but no consistency check between `vendor/` and the module files.

## Proposed change

Add a CI check (e.g. extend `scripts/check_deps.sh` or add a dedicated target) that:

1. Runs `go mod tidy` and confirms `go.mod`/`go.sum` are unchanged (already done).
2. Runs `go mod vendor` and confirms the working tree is unchanged.
3. Runs `go mod verify` to confirm vendored contents match recorded hashes.

Fail the job with a clear diff / `git status --porcelain` output if anything changed, instructing the contributor to run `go mod tidy && go mod vendor` and commit the result.

Sketch:

```bash
go mod tidy
go mod vendor
if [ -n "$(git status --porcelain go.mod go.sum vendor/)" ]; then
  echo "go.mod/go.sum/vendor are out of date. Run 'go mod tidy && go mod vendor' and commit."
  git status --porcelain go.mod go.sum vendor/
  git diff -- go.mod go.sum vendor/
  exit 1
fi
```

## Acceptance criteria

- CI fails when `go.mod`/`go.sum` are not tidy (regression guard for existing behavior).
- CI fails when `vendor/` is not in sync with `go.mod`/`go.sum`.
- Failure output tells the contributor exactly what to run.

moradna:
#1058


scripts/check_deps.sh
```diff
@@ -22,8 +22,19 @@ go mod tidy -modfile="${dep_tempdir}/go.mod"
...
     if ! diff -q "${arma_dir}/$f" "${dep_tempdir}/$f"; then
-        echo "It appears $f is stale. Please run 'go mod tidy' and 'go mod vendor'."
+        echo "It appears $f is stale. Please run 'go mod tidy && go mod vendor'."
         diff -u "${arma_dir}/$f" "${dep_tempdir}/$f"
         exit 1
     fi
 done
+
+# check that vendor/ is in sync with go.mod/go.sum
+go mod vendor
```
HagarMeir: **Check now mutates the real working tree.** Unlike the earlier check (which copies `go.mod`/`go.sum` into a tempdir specifically to avoid side effects), `go mod vendor` runs against the actual repo and rewrites `go.mod`, `go.sum`, and `vendor/` in place. Since `check-deps` is a dependency of `make linter`, a developer running `make linter` locally will silently have their `vendor/` regenerated; and when out of sync, the script exits 1 leaving the tree dirty (clobbering any intentional local state). Consider vendoring into a throwaway copy / restoring afterward, matching the tempdir approach used above.

scripts/check_deps.sh
```diff
@@ -22,8 +22,19 @@ go mod tidy -modfile="${dep_tempdir}/go.mod"
...
+if [ -n "$(git status --porcelain go.mod go.sum vendor/)" ]; then
```
HagarMeir: **Vendor-sync check silently no-ops outside a git work tree.** If run without a `.git` checkout (e.g. from a source tarball, or a shallow/detached CI context), `git status --porcelain` fails. Because command-substitution failures do not trip `set -e` (verified: the script continues), `[ -n "" ]` is false and the check passes with a false-positive 'in sync' result — giving false confidence rather than an error.

HagarMeir: **False failures from unrelated dirty state.** `git status --porcelain go.mod go.sum vendor/` flags *any* uncommitted change to those paths, not just vendor drift. A developer with in-progress `go.mod` edits (or a partially staged tree) will get a spurious 'out of date' failure from `make linter` even when vendor is perfectly in sync with go.mod/go.sum. The tempdir-diff approach above avoids this by comparing against a fresh copy.

scripts/check_deps.sh
```diff
@@ -22,8 +22,19 @@ go mod tidy -modfile="${dep_tempdir}/go.mod"
...
+    echo "go.mod/go.sum/vendor are out of date. Run 'go mod tidy && go mod vendor' and commit."
+    git status --porcelain go.mod go.sum vendor/
+    git diff -- go.mod go.sum vendor/
+    exit 1
+fi
+
+go mod verify
```
HagarMeir, line 38: **`go mod verify` does not validate the vendored sources.** In a vendored module, builds use `vendor/`, but `go mod verify` only checks modules in the local module cache against go.sum — it does not verify `vendor/` contents at all. Placed here it adds little to the vendor-sync guarantee this PR targets (and may require populating the cache to run).

scripts/check_deps.sh
```diff
@@ -22,8 +22,19 @@ go mod tidy -modfile="${dep_tempdir}/go.mod"
...
         exit 1
     fi
 done
+
+# check that vendor/ is in sync with go.mod/go.sum
+go mod vendor
+if [ -n "$(git status --porcelain go.mod go.sum vendor/)" ]; then
+    echo "go.mod/go.sum/vendor are out of date. Run 'go mod tidy && go mod vendor' and commit."
+    git status --porcelain go.mod go.sum vendor/
+    git diff -- go.mod go.sum vendor/
```
HagarMeir: **Diagnostic `git diff` omits newly added vendor files and can flood the log.** `git diff -- vendor/` shows only tracked-file changes, so brand-new vendored dependency dirs (which appear as `??` untracked in the status output) won't show in the diff; conversely a large drift across the 3000+ vendored files dumps a huge diff into CI output. Consider limiting to the porcelain status or `--stat`.

scripts/check_deps.sh
moradna: agreed, kept it because it is explicitly requested in #1058

HagarMeir:
## Review summary

Nice addition — the `go mod vendor -o` + `diff -rq` block is a solid way to guarantee `vendor/` stays in sync, which is the main goal here. A few things worth considering:

**1. `go mod verify` doesn't validate the vendored sources.** Since the orderer builds from `vendor/`, and `go mod verify` only checks that modules in the local module *cache* match `go.sum` (it never inspects `vendor/`), this step doesn't add to the vendor-integrity guarantee the PR is after — the `diff -rq` block already covers that. Consider dropping it.

**2. The new steps slow down `check-deps`, which runs on every `make linter` / `make basic-checks`.** The full `go mod vendor -o` regeneration plus `go mod verify` forces downloading the entire module build list into the cache (grpc, x/net, otel, etc. on a clean run), adding a hard network dependency + several seconds to a frequently-run target. Dropping `verify` (see #1) removes most of the cost.

**3. `diff -rq` can spuriously fail on Go toolchain skew.** `go mod vendor` output (notably `vendor/modules.txt`) is toolchain-sensitive. A contributor on a different Go minor version than the one that generated the committed vendor can hit a "vendor out of date" failure they can't resolve locally. `go.mod` pins `go 1.27.0`, but local runs use whatever Go is installed.

**4. `if ! diff -rq ...` treats exit code 2 (I/O error) the same as exit 1 (differences).** If `go mod vendor` partially fails or a file is unreadable, the script prints the misleading "vendor/ is out of date" message and hides the real error. Unlike the go.mod branch, it also prints no diff to help diagnose what differs.

**5. (pre-existing, minor) The temp dir is created inside the repo working tree.** `mktemp -d "$(basename "$0")"-XXXXX` uses a relative template, so cleanup relies solely on the `EXIT` trap, which doesn't fire on SIGKILL / CI cancellation. The PR's longer runtime widens the window where a killed run leaves a stray `check_deps.sh-XXXXX/` behind. `mktemp -d -t` / `${TMPDIR}` would keep it out of the tree.

#1 and #2 are the substantive ones — #3–5 are robustness nits (4 and 5 predate this PR but are aggravated by the longer runtime).

HagarMeir (approved):

PR #1201, opened 2026-09-08
config ack router unit test

MayRosenbaum:
#1202


common/configack/sender.go
```diff
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/grpc"
 )
 
-const SubmitConfigAckTimeout = 60 * time.Second // TODO: expose in local config
+var SubmitConfigAckTimeout = 60 * time.Second // TODO: expose in local config
```
tock-ibm, line 21: Turning the submit timeout into a mutable exported package global just to let one test override it is a symptom-level fix. The `sender` struct already holds `DialTimeout`, `minRetryInterval`, and `maxRetryInterval` as instance fields set in `NewSender`; `SubmitConfigAckTimeout` is instead a shared mutable global read from `SubmitConfigAck` without synchronization.

It is safe today only because no router test runs in parallel and `TestRouterConfigAck_FailureIsNonFatal`'s override/restore happens-before/after the single read. Adding any `t.Parallel()` test, or a second concurrent config-apply that reads the global while it is overridden, would become an unsynchronized read/write and a `-race` failure. Threading the timeout through `ConnectionInfo`/`NewSender` as a per-instance field is the deeper fix and matches the existing pattern for the sibling timeouts.

node/router/router_configack_test.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+		conn.Close()
+		return true
+	}, 10*time.Second, 100*time.Millisecond)
+}
+
+// autoRemoveTimeoutConfigUpdate builds a config update that changes a batching timeout. It
+// does not require an admin restart in the router, so the router applies it dynamically and
+// advances to config sequence 1.
+func autoRemoveTimeoutConfigUpdate(t *testing.T, dir string) []byte {
```
tock-ibm, line 71: `autoRemoveTimeoutConfigUpdate` re-implements the exact config-update construction already inlined in `TestSendConfigUpdate` (`router_reconfig_test.go`), which builds the identical batch-timeout update inline via `NewConfigUpdateBuilder(...).UpdateBatchTimeouts(t, NewBatchTimeoutsConfig(BatchTimeoutsConfigName.AutoRemoveTimeout, "15ms"))`. The new helper duplicates that logic in a second file rather than being reused by both, so the two copies can drift when the builder API or the timeout value changes. Consider extracting the helper to a shared location so both call sites use it.

node/router/router_configack_test.go
```diff
@@ -0,0 +1,276 @@
...
+	return configUpdatePbData
+}
+
+// Scenario:
+//  1. Start a router, stub batcher, and stub consenter with the initial config.
+//  2. Record the ConfigAck the router sends by defining an AckConfigHandler on the stub consenter.
+//  3. Deliver a config update (batching timeout change) through the consenter path.
+//  4. Verify the router advances to config sequence 1 and that it sent exactly one
+//     ConfigAck identifying itself as the router (NodeType=ROUTER, Shard=0) on sequence 1.
+func TestRouterSendsConfigAck(t *testing.T) {
```
tock-ibm, line 84: The four new tests each repeat ~15 lines of identical setup boilerplate (also duplicated from `router_reconfig_test.go`): `createReconfigTestSetup`, install an `AckConfigHandler` that records, `testSetup.Start()`, `defer Stop`, `require.Eventually(StateRunning && seq==0)`, `waitForRouterServing`, then `SendConfigUpdate`. A change to the startup/readiness sequence would have to be edited in each copy. A single helper (e.g. `startRunningRouterWithRecorder(t)`) would remove the repetition.

node/router/router_configack_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// the router applies the new config and advances to config sequence 1.
+	require.Eventually(t, func() bool {
+		status := testSetup.routerNode.GetStatus()
+		return status.State == node_utils.StateRunning && status.ConfigSequenceNumber == 1
+	}, 20*time.Second, 100*time.Millisecond)
+
+	// exactly one ConfigAck is sent for the new config block.
+	require.Eventually(t, func() bool {
+		return recorder.count() == 1
```
tock-ibm, line 117: `require.Eventually` with a strict `count() == 1` can mask a regression that emits a second ack, and it races the following `require.Len(acks, 1)`. If a router regression caused a duplicate `ConfigAck`, `Eventually` returns at the transient moment `count == 1` and never observes the second one; the subsequent `require.Len(acks, 1)` then races the second append (passing or failing on timing). Asserting the ack count only after the terminal state (`seq==1`) has been reached — rather than polling for exact equality — would make the exactly-once guarantee robust.

node/router/router_configack_test.go
```diff
@@ -0,0 +1,276 @@
...
+	}, 20*time.Second, 100*time.Millisecond)
+
+	// the accepted ack is for sequence 1, and the sender made more than one attempt.
+	require.Eventually(t, func() bool {
+		return recorder.count() == 1
+	}, 20*time.Second, 100*time.Millisecond)
+
+	require.Greater(t, atomic.LoadInt32(&callCount), int32(1))
+	acks := recorder.all()
+	require.Equal(t, protos.NodeType_ROUTER, acks[0].NodeType)
```
tock-ibm, line 176: Tests 2 and 3 index `acks[0]` (here, and at line 227) relying solely on the preceding `recorder.count() == 1` `Eventually` to guarantee `len >= 1`, without the `require.Len(acks, 1)` guard that `TestRouterSendsConfigAck` uses at line 121. It is safe today, but if the count-gate assumption ever changed these would panic with index-out-of-range rather than fail with a readable assertion. Adding a `require.Len` guard here would keep the three tests consistent.

tock-ibm (commented):
Code review of the config-ack router unit tests. The change is well-constructed and test-focused. I verified there is no actual data race today (no `t.Parallel()` anywhere in the router package, and the single global mutation is fully ordered by the override/restore in the one test that touches it) and no linter breakage (`gochecknoglobals` is not enabled). The findings below are design/quality suggestions rather than correctness bugs — the most notable being the `const`→`var` global.

node/router/router_configack_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// the ack never succeeds, but the router still applies the new config and
+	// advances to config sequence 1 once the submission times out.
+	require.Eventually(t, func() bool {
+		status := testSetup.routerNode.GetStatus()
+		return status.State == node_utils.StateRunning && status.ConfigSequenceNumber == 1
+	}, 20*time.Second, 100*time.Millisecond)
+
+	// the router did attempt to send the ack.
+	require.Greater(t, atomic.LoadInt32(&callCount), int32(0))
```
tock-ibm, line 275: This test's only ack-path assertion (`require.Greater(callCount, int32(0))`) is weaker than the scenario it documents ("exhausts its retries and times out"). It merely confirms one attempt occurred; the test would still pass if the sender made a single attempt and gave up without retrying. The meaningful guarantee (seq advances within 20s despite the 1s timeout) is carried entirely by the `seq==1` `Eventually`, so this assertion does not actually verify the retry/timeout behavior it describes.

common/configack/sender.go
tock-ibm, line 21: We can calculate this from some existing timeout - like the consensus (first-strike+second-strike)*3;
Or maybe from something in the local config?

node/router/router_configack_test.go
MayRosenbaum: Let's keep it for readability.

common/configack/sender.go
MayRosenbaum: I don’t think this timeout is necessarily related to the other timeouts. Maybe we could derive it from the keepalive settings, but I think it makes sense

PR #1203, opened 2026-09-09
added docker compose for running local arma network run with grafana

moran-abilea:
Running Arma with Prometheus and Grafana inside docker compose

it runs an Arma sample network together with Prometheus and
Grafana, and watch the metrics of a load test on a dashboard while it runs.

By default it runs a network of four parties with two shards — each party has a
router, two batchers, a consensus and an assembler node — and drives it with a five
minute load test. The number of parties, the number of shards, the duration and the
load are all configurable.

All Arma nodes share one container, as in the [all-in-one](../all-in-one/README.md)
example. Prometheus, Grafana, the load client and the receiving clients each get
their own.

Everything runs as the user that invokes the scripts. Do not use `sudo`: the scripts
need no root, and nothing they create is owned by root.

**This example is for local use only** - a laptop or a development machine, driven by
hand. Grafana is reachable without a login, nothing speaks TLS to the host, the ports
are published on all interfaces, and the crypto material is generated fresh for each
run and thrown away with it. Convenient locally, and deliberately not a deployment
blueprint: running this anywhere shared would need a different setup.


node/examples/grafana/scripts/build_docker.sh
```diff
@@ -0,0 +1,12 @@
...
+# Copyright IBM Corp. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+set -eu
+
+# The same "arma" image the other examples use.
+cd "$(dirname "${BASH_SOURCE[0]}")/../.."
+bash ./scripts/build_docker.sh
```
moradna: I don't think this wrapper file is needed, you can use the existing node/examples/scripts/build_docker.sh directly

node/examples/grafana/scripts/run_sample.sh
```diff
@@ -0,0 +1,115 @@
...
+    echo "      - targets:"
+    for host in ${hosts}; do
+        echo "          - ${host}:${OPERATIONS_PORT}"
+    done
+} > "${SAMPLE_DIR}/prometheus.yml"
+
+mkdir -p "${SAMPLE_DIR}/grafana/provisioning/datasources" \
+    "${SAMPLE_DIR}/grafana/provisioning/dashboards" \
+    "${SAMPLE_DIR}/grafana/dashboards"
+cp "${EXAMPLE_DIR}/grafana/datasource.yaml" "${SAMPLE_DIR}/grafana/provisioning/datasources/"
```
moradna: Do we need to copy the grafana files into `/tmp`? They can be mounted directly from `compose.yaml` instead

node/examples/grafana/scripts/run_sample.sh
```diff
@@ -0,0 +1,115 @@
...
+
+cd "${EXAMPLE_DIR}/../../.."
+make binary
+
+rm -rf "${SAMPLE_DIR}"
+./bin/armageddon generate --config="${DEPLOYMENT}" --output="${SAMPLE_DIR}"
+
+# The generator leaves the operations port unset, which means a random one, and binds it
+# to localhost. Prometheus scrapes it from another container and needs to know it in
+# advance. Every node has its own container, so one port serves them all.
```
moradna: maybe this comment can be shorter, for example  `Expose a fixed operations port so Prometheus can scrape the nodes.`

node/examples/grafana/scripts/run_dashboard.sh
```diff
@@ -0,0 +1,115 @@
...
+        echo "          - ${host}:${OPERATIONS_PORT}"
+    done
+} > "${SAMPLE_DIR}/prometheus.yml"
+
+mkdir -p "${SAMPLE_DIR}/grafana/provisioning/datasources" \
+    "${SAMPLE_DIR}/grafana/provisioning/dashboards" \
+    "${SAMPLE_DIR}/grafana/dashboards"
+cp "${EXAMPLE_DIR}/grafana/datasource.yaml" "${SAMPLE_DIR}/grafana/provisioning/datasources/"
+cp "${EXAMPLE_DIR}/grafana/dashboards.yaml" "${SAMPLE_DIR}/grafana/provisioning/dashboards/"
+cp "${EXAMPLE_DIR}/grafana/arma-dashboard.json" "${SAMPLE_DIR}/grafana/dashboards/"
```
moradna, line 68: Do we need to copy the grafana files into `/tmp`? They can be mounted directly from `compose.yaml` instead

node/examples/grafana/scripts/run_sample.sh
```diff
@@ -0,0 +1,115 @@
...
+            print "    ListenPort: " port
+            next
+        }
+        { print }
+    ' "${config}" > "${config}.new"
+    mv "${config}.new" "${config}"
+done
+
+# Scrape every node of the network. The hosts come from the deployment, so to change the
+# number of parties or shards edit ../config/example-deployment.yaml and ../compose.yaml.
```
moradna: I don't think this comment is needed

node/examples/grafana/README.md
```diff
@@ -0,0 +1,107 @@
...
+| `OPEN_BROWSER`     | `auto`    | `false` to never open a browser            |
+
+```
+EXAMPLE:
+
+RATE=500 DURATION_SECONDS=60 ./node/examples/grafana/scripts/run_sample.sh
+```
+
+The number of parties and shards comes from the network itself, so changing it means
+editing [../config/example-deployment.yaml] and
```
moradna: this looks like a link, but it doesn't have a target

node/examples/grafana/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+1. Add the metric in `node/<role>/metrics.go`, keeping `party_id` and `shard_id` among its
+   `LabelNames` as the existing metrics do. 
+   Its Prometheus name is `<namespace>_<name>`.
+2. Rebuild the image and start again: `build_docker.sh`, `clean_sample.sh`, `run_sample.sh`.
+3. Confirm it is exported, in the Prometheus UI or straight from a node:
+   `docker exec arma-grafana-prometheus-1 wget -qO- http://router.p1:8080/metrics | grep <name>`
+4. Add a panel in Grafana, selecting `${DS_PROMETHEUS}` as its data source rather than the
+   Prometheus data source itself, so the dashboard stays independent of this setup.
+5. Export it over `grafana/arma-dashboard.json` and document the metric in [docs/monitoring/metrics.md].
```
moradna: I think this section is a bit out of scope here.

node/examples/grafana/scripts/build_docker.sh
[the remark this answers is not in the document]
moran-abilea: fixed as suggested,
also not written here in the review note but also fixed as suggested: added correct links in the readme.md

node/examples/grafana/scripts/run_dashboard.sh
```diff
@@ -0,0 +1,115 @@
...
+
+# Prometheus and Grafana run as their images' own users, so let them read what they mount.
+chmod a+x "${SAMPLE_DIR}"
+chmod -R a+rX "${SAMPLE_DIR}/grafana" "${SAMPLE_DIR}/prometheus.yml"
+
+"${COMPOSE[@]}" up -d
```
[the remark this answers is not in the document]
moran-abilea: checked it, it's wrong
nothing to fix here

node/examples/grafana/grafana/arma-dashboard.json
```diff
@@ -0,0 +1,3492 @@
...
+                    "query": {
+                      "kind": "DataQuery",
+                      "group": "prometheus",
+                      "version": "v0",
+                      "datasource": {
+                        "name": "${DS_PROMETHEUS}"
+                      },
+                      "spec": {
+                        "editorMode": "code",
+                        "expr": "sum by (party_id, shard_id) (\r\n  batch_ledger_append_latency_seconds_sum\r\n)\r\n/\r\nsum by (party_id, shard_id) (\r\n   batch_ledger_append_latency_seconds_count\r\n)",
```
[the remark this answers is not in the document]
moran-abilea: checked it, it's wrong
nothing to fix here

node/examples/grafana/scripts/run_sample.sh
moran-abilea: fixed as suggested

node/examples/grafana/scripts/run_sample.sh
moran-abilea: removed entierly

node/examples/grafana/README.md
moran-abilea: alread fixed, with the rest of the links

node/examples/grafana/README.md
moran-abilea: section removed entierly

node/examples/grafana/scripts/build_docker.sh
moran-abilea: fixed as suggested

node/examples/grafana/scripts/run_dashboard.sh
moran-abilea: I tried mounting these files directly from the repository, and it simplifies the script. However, it makes the example dependent on the checkout permissions. With umask 027, the files are 640 and Grafana fails to read the bind mounts.

We could chmod the repository files before starting Compose, but I'd prefer not to modify permissions in the user's checkout. An init container also works, but adds a service and named volumes just to stage three files.

So I kept the small staging step in /tmp, and added a comment explaining that it intentionally normalizes permissions for the container.

Issue #1185, opened 2026-09-03
Batcher mem pool optimizations


HagarMeir:
For example, returning calculated req ids from some methods to be reused by other methods, since recalculating may be expensive. 
Another example is supporting bulk operations, like removing an entire batch of transactions. 
Or combining operations in some cases. 
Need to think of more optimizations.

PR #1205, opened 2026-09-10
batcher,request: reuse computed request ids in mem pool

HagarMeir:
## What

The mem pool already computes a request's id when it enters the pool, and the secondary batcher computes it again while verifying a pulled batch. Both were then discarded, forcing `removeRequests` to recompute the id of every request in a batch before removing it from the pool — a redundant hash per request on the removal hot path.

This change returns the already-computed ids so removal can reuse them:

- **`Pool.NextRequests` / `BatchStore.Fetch`** now return the requests **and their corresponding ids** (aligned by index). The **primary** threads these ids into `removeRequestsByIDs` instead of re-hashing.
- **`VerifyBatchedRequests`** now returns the per-request ids it already computes during verification; the **secondary** reuses them for removal. Each batch index is owned by exactly one verification worker (`workerID == j%numWorkers`), so filling the ids slice is race-free, and the ids are fully populated whenever verification succeeds (partial only on the error path, where the caller complains and never removes).
- `removeRequests` is split into `removeRequests` (secondary — computes ids from a pulled batch) and `removeRequestsByIDs` (reuses ids), replacing the previous `nil`-sentinel so each call site's intent is explicit.

## Testing

- Extended the `request` pool/batchstore tests and the batcher verifier tests to assert the returned ids correspond to their requests (`ids[i] == RequestID(reqs[i])`).
- `go test -race ./request/...` and `go test -race ./node/batcher/` pass.

Part of #1185 (primary + secondary mem-pool paths; bulk-remove and further ideas left for follow-up).


node/batcher/batcher_role.go
```diff
@@ -404,18 +408,16 @@ func (b *BatcherRole) runPrimary() {
...
 }
 
-func (b *BatcherRole) removeRequests(batch types.BatchedRequests) {
-	reqInfos := make([]string, 0, len(batch))
-	for _, req := range batch {
-		reqInfos = append(reqInfos, b.RequestInspector.RequestID(req))
-	}
-	b.MemPool.RemoveRequests(reqInfos...)
+// removeRequestsByIDs removes the given request ids from the mem pool. Used by
+// the primary, which gets the ids from NextRequests and so avoids recomputing them.
```
tock-ibm, line 418: The doc comment says this is "Used by the primary", but `removeRequestsByIDs` is also called by the secondary (line 491), which gets its ids from `VerifyBatchedRequests` rather than `NextRequests`. The comment as written suggests only the primary path uses it, which could mislead a future reader who expects a separate secondary removal helper (as the PR description's "split into `removeRequests`/`removeRequestsByIDs`" wording implies).

```suggestion
// removeRequestsByIDs removes the given request ids from the mem pool. Used by
// both the primary (ids from NextRequests) and the secondary (ids from
// VerifyBatchedRequests), so neither recomputes them.
```

node/batcher/batcher_role.go
```diff
@@ -29,7 +29,7 @@ type RequestInspector interface {
 
 //go:generate counterfeiter -o mocks/mem_pool.go . MemPool
 type MemPool interface {
-	NextRequests(ctx context.Context) [][]byte
+	NextRequests(ctx context.Context) ([][]byte, []string)
```
tock-ibm, line 32: Nit:
The parameter name of `RemoveRequests(requests ...string)` should be `RemoveRequests(requestIDs ...string)`

node/batcher/batcher_role.go
```diff
@@ -404,18 +408,15 @@ func (b *BatcherRole) runPrimary() {
...
 
-func (b *BatcherRole) removeRequests(batch types.BatchedRequests) {
-	reqInfos := make([]string, 0, len(batch))
-	for _, req := range batch {
-		reqInfos = append(reqInfos, b.RequestInspector.RequestID(req))
-	}
-	b.MemPool.RemoveRequests(reqInfos...)
+// removeRequestsByIDs removes the given request ids from the mem pool
+func (b *BatcherRole) removeRequestsByIDs(reqIDs []string) {
+	b.MemPool.RemoveRequests(reqIDs...)
```
tock-ibm, line 419: A question for a future PR:
Can the remove be done in parallel with multiple goroutines in the memPool?


`func (rp *Pool) RemoveRequests(requestsIDs ...string) ` runs serially, whereas in the pending it is done in parallel.

tock-ibm (approved):
Approved with two optional comments for a future PR.

PR #1210, opened 2026-09-10
consensus,batcher: revive BAFs one config behind via StaleConfigBAFs

### LINKED ISSUE #810: Consenter reconfig: handle BAFs submitted but not consolidated
Note:
During a config change, there may be BAFs that have been submitted to the consensus and are in the SmartBFT mempool.
These BAFs exist in the ledgers of batchers that have submitted them.
It is required, in dynamic, for BAFs with expired sqn to pop up in decisions even after a config change due to this effect.
These BAFs appear in one decision and are cleared in the next decision.
This will allow batchers to discover this, and revive the TXs in said batches.

The batchers, however, currently do rescan their ledgers when they restart as well to discover this. This will be solved in dynamic reconfig.

Investigate and resolve, Moved from #487 

This is blocked by #830 since as long as SmartBFT is stopped when dynamically reconfiguring the consenter then all reqs in its mem pool is lost.

One more issue is when the batcher is behind the consensus and tries to send a BAF with a previous config seq. That batch will be rejected by the consensus nodes but I do not think the batcher is aware it was rejected. This BAF will not appear in any pending list. Need to see how to handle this.

HagarMeir:
## Problem

During a config change a batcher's **local** config sequence lags the consenter's, so its batching loop keeps emitting BAFs at the old sequence. Two gaps followed (both under #810):

- **Case B** — a stale BAF arriving *after* the consenter advanced was rejected by `verifyCE` and silently swallowed, so the batcher removed the batch's requests and the transactions were lost.
- **Set 2** — a stale BAF *accepted* into the mempool *before* the bump was filtered out of every later proposal and never surfaced, so the batcher could not revive it.

## Approach

Instead of dropping a BAF exactly one config behind, the consenter **accepts** it (its signature is still verified), consensus **parks it for a single decision** in a new `State.StaleConfigBAFs` array, and batchers **monitor that array on every delivered decision** and resubmit their own such BAFs' requests from their ledger so they re-batch under the new sequence.

This needs **no per-BAF round-trip**: the batcher stays fire-and-forget on the send and learns of a stale BAF asynchronously over the decision stream it already consumes (an earlier per-BAF-ack prototype was rejected for its synchronous cost over WAN).

**Correctness invariant:** decisions are applied in order and the config-block decision precedes any decision that surfaces a mismatch, so a batcher has already applied the new config by the time it sees its BAF in `StaleConfigBAFs` — it re-batches under the correct sequence, so there is no re-batch loop. `StaleConfigBAFs` is **never** used to extract a Batch Attestation (kept separate from `Pending`).

## What changed

**Consensus**
- `node/consensus/state/state.go` — new `State.StaleConfigBAFs`; `Clone` deep-copies it; `Process` recomputes it fresh each round; `filterCEsWithDiffConfigSeq` now diverts `configSeq-1` BAFs into it instead of dropping them; `Serialize`/`Deserialize` updated.
- `node/protos/state/state.proto` (+ regenerated `state.pb.go`) — new `stale_config_bafs` field.
- `node/consensus/consensus.go` — `verifyCE` accepts a BAF at the current sequence *or exactly one behind* (still verifying its signature); the `VerifyProposal` per-request skip is relaxed to match.

**Batcher**
- `node/batcher/batcher_role.go` — new `ResubmitStaleConfigBAFs` and a shared `reviveBAFRequests` helper.
- `node/batcher/batcher.go` — triggers it per decision in `replicateDecision`.

## Scope / limitations

- **BAFs only** — stale complaints and config requests are still dropped.
- Bounded to **one config behind**; anything staler is still rejected (bounds the ingestion surface).
- Signatures are verified with the current verifier; if the signer set changed (an identity reconfig, which takes the admin-restart path anyway) an old BAF may not verify and is dropped.
- This covers the **dynamic** reconfig path. On the admin-restart path the batcher's mempool is rebuilt empty — a separate, pre-existing concern.

## Testing

- `node/consensus/state/state_test.go` — `Process` surfaces a one-behind BAF into `StaleConfigBAFs` (not `Pending`, not extracted), clears it the next round; `Serialize`/`Deserialize` round-trip.
- `node/consensus/consensus_test.go` — `TestVerifyRequestAcceptsStaleByOneConfigSeq`: accepts current / one-behind, rejects two-behind.
- `node/batcher/batcher_role_test.go` — `TestResubmitStaleConfigBAFs`.
- `make unit-tests-batcher` and `make unit-tests-consensus` pass locally with `-race`.

Closes #810


node/consensus/state/state.go
```diff
@@ -417,13 +452,24 @@ func (s *State) CollectAndDeduplicateEvents(l *flogging.FabricLogger, ces ...Con
...
+	var staleConfigBAFs []types.BatchAttestationFragment
 	for _, ce := range ces {
 		if ce.BAF != nil {
-			if ce.BAF.ConfigSequence() == configSeq {
+			switch {
+			case ce.BAF.ConfigSequence() == configSeq:
 				filteredEvents = append(filteredEvents, ce)
-			} else {
+			case configSeq > 0 && ce.BAF.ConfigSequence() == configSeq-1:
+				l.Warnf("baf is one config behind (currently %d); surfacing as stale for revival; %s", configSeq, ce.BAF.String())
```
tock-ibm, line 470: `filterCEsWithDiffConfigSeq` runs inside `Process` → `SimulateStateTransition`, which is invoked by both `AssembleProposal` (leader) and `VerifyProposal` (every follower) on every proposal. Logging each one-behind BAF at **Warn** means that during a reconfiguration window — exactly when a lagging batcher emits many stale BAFs — every consensus node floods its log at Warn level, once per stale BAF per proposal. Consider `Debugf`/`Infof` (matching the sibling drop branch just below, which uses `Debugf`), or dedup/rate-limit.

node/consensus/state/state.go
```diff
@@ -239,7 +267,12 @@ type ShardTerm struct {
 func (s *State) Process(l *flogging.FabricLogger, configSeq types.ConfigSequence, ces ...ControlEvent) (*State, []types.BatchAttestationFragment, []*ConfigRequest) {
 	nextState := s.Clone()
 
-	filteredCEs := filterCEsWithDiffConfigSeq(configSeq, l, ces...)
+	// StaleConfigBAFs is recomputed fresh every round so it surfaces a stale BAF for exactly one
+	// decision (Clone copied the parent's, so reset it here before repopulating).
+	nextState.StaleConfigBAFs = nil
```
tock-ibm, line 272: This `nextState.StaleConfigBAFs = nil` is dead: three lines down it is unconditionally overwritten by `nextState.StaleConfigBAFs = staleConfigBAFs` (a direct assignment, not an append), so the reset has no observable effect. Relatedly, `Clone()` now allocates and copies `StaleConfigBAFs`, which `Process` immediately throws away — harmless, but both can be dropped. (The `Clone` copy is still needed for the `ApplyConfigToState` caller, so keep it there.)

node/consensus/consensus.go
```diff
@@ -539,7 +539,11 @@ func (c *Consensus) VerifyProposal(proposal smartbft_types.Proposal) ([]smartbft
 		if err != nil {
 			return nil, fmt.Errorf("invalid request %s: %v", rawReq, err)
 		}
-		if configSeq != c.VerificationSequence() {
+		verSeq := c.VerificationSequence()
+		// A control event one config behind is accepted (a stale-by-one BAF is surfaced for revival),
+		// so it must still be verified here and reflected in reqInfos, consistently with the leader.
+		// Anything staler is ignored.
+		if configSeq != verSeq && !(verSeq > 0 && configSeq == verSeq-1) {
```
tock-ibm, line 546: The new one-behind acceptance here (`configSeq == verSeq-1`) has no direct test. `TestVerifyRequestAcceptsStaleByOneConfigSeq` exercises `verifyCE` via `VerifyRequest`, but not this `VerifyProposal` skip-relaxation, and the previously-covering assertion in `consensus_real_reconfig_test.go` (`configSeq-1` → "mismatch config sequence") was deleted without a positive replacement. Worth a focused test that a proposal carrying a one-behind BAF verifies and is reflected in `reqInfos` (so leader/follower `reqInfos` stay consistent).

node/consensus/consensus.go
```diff
@@ -539,7 +539,11 @@ func (c *Consensus) VerifyProposal(proposal smartbft_types.Proposal) ([]smartbft
...
 			continue // ignore (no need to verify) request with mismatch config sequence
 		}
```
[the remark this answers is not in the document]
HagarMeir: Restricted the one-config-behind exception to BAFs: getReqConfigSeq now reports whether the event is a BAF, and VerifyProposal only verifies requests that affect the computed state (current sequence, or a one-behind BAF). A stale complaint or config request is no longer routed to verifyCE; it is reported in reqInfos without verification so it is still removed from the request pool (its ID embeds the config sequence, so exactly the stale request is removed, never a legitimate one at the current sequence).

node/consensus/consensus.go
HagarMeir: Added TestVerifyProposalAcceptsOneBehindBAFAndSkipsOtherStaleCEs in consensus_test.go: a proposal carrying a one-behind BAF verifies, is surfaced in StaleConfigBAFs, and appears in reqInfos; a one-behind complaint in the same proposal is not verified but is still reported in reqInfos, and does not reject the proposal. Confirmed it fails against the previous type-agnostic logic and passes with the fix.

node/batcher/batcher_role.go
```diff
@@ -269,37 +269,56 @@ func (b *BatcherRole) ResubmitPendingBAFs(state *state.State, prevPrimary types.
...
-						continue
-					}
-					b.Logger.Errorf("Failed submitting request to pool; err: %v", err)
-				}
+		}
+		if err := b.MemPool.Submit(req); err != nil {
+			if strings.Contains(err.Error(), "already inserted") {
+				b.Logger.Debugf("Failed submitting request to pool; err: %v", err)
+				continue
 			}
```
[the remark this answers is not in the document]
HagarMeir: This failure mode is not specific to the stale-config path: reviveBAFRequests is shared with the pre-existing ResubmitPendingBAFs path, and a full mempool dropping a Submit is a general property of the revival/submit path rather than something introduced here. Rather than add a retryable queue only on this path, tracking it as a separate general concern.

node/consensus/state/state.go
HagarMeir: Changed Warnf to Infof. Kept it above Debug since surfacing a BAF for revival is a more notable event than a plain drop, but no longer at Warn so it does not read as an error during a reconfiguration window.

node/consensus/state/state.go
HagarMeir: Removed the dead nextState.StaleConfigBAFs = nil (it was unconditionally overwritten three lines down). Kept the Clone() copy of StaleConfigBAFs since the ApplyConfigToState caller relies on it.

node/protos/state/state.proto
```diff
@@ -60,4 +60,5 @@ message State {
   repeated BatchAttestationFragment pending = 3;  // Pending batch attestation fragments received by consenters from batchers
   repeated Complaint complaints = 4; // Complaints received by consenters from batchers
   bytes app_context = 5; // Auxiliary data that may be served as context to the application
+  repeated BatchAttestationFragment stale_config_bafs = 6; // BAFs one config behind, surfaced for one decision so batchers can revive their requests
```
tock-ibm: I would limit the comment here to "BAFs with stale config, surfaced for one decision so batchers can revive their requests"

node/consensus/consensus.go
```diff
@@ -1269,7 +1274,11 @@ func (c *Consensus) verifyCE(req []byte) (smartbft_types.RequestInfo, *state.Con
...
 		return reqID, ce, c.SigVerifier.VerifySignature(ce.Complaint.Signer, ce.Complaint.Shard, ce.Complaint.ToBeSigned(), ce.Complaint.Signature)
 	} else if ce.BAF != nil {
-		if ce.BAF.ConfigSequence() != configSeq {
+		// Accept a BAF whose config sequence is exactly one behind, in addition to the current one, so
+		// that a batcher which fell behind a config change is not silently dropped: consensus surfaces it
```
tock-ibm: ... that the BAFs from a batcher which fell behind a config change are not silently dropped: consensus surfaces it

tock-ibm: I don't a risk in accepting stale BAFs from any config-seq in the past. They only live for one decision so the memory bloat is minimal. It closes a potential (rare) gap of a batcher that fell behind more than 1 config change.

tock-ibm (approved):

PR #1211, opened 2026-09-14
request: fix deadlock between Pool.Close and BatchStore.Fetch (#843)

### LINKED ISSUE #843: Test flake - deadlock close and fetch from pool
When a batcher is stopped it cancels the ctx used for calling fetch `b.cancelBatch()` and closes the pool `b.MemPool.Close()`.
Pool close first needs to lock the pool's lock. 
Pool `NextRequests` also grabs the same lock and waits for fetch to return. 
In `batchstore` `Fetch` there is a wait on signal.
If the signal is called before wait is called we hang while holding the pool's lock.
The signal might be called early when the ctx is canceled even before calling fetch.

```
2026-04-19T11:28:05.0906336Z goroutine 17793 [sync.RWMutex.Lock, 18 minutes]:
2026-04-19T11:28:05.0906516Z sync.runtime_SemacquireRWMutex(0xc001e9f144?, 0x1?, 0x167eef1?)
2026-04-19T11:28:05.0906710Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/runtime/sema.go:105 +0x25
2026-04-19T11:28:05.0906806Z sync.(*RWMutex).Lock(0xc001e9f130)
2026-04-19T11:28:05.0907000Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/sync/rwmutex.go:155 +0x89
2026-04-19T11:28:05.0907224Z github.com/hyperledger/fabric-x-orderer/request.(*Pool).Close(0xc001e9f130)
2026-04-19T11:28:05.0907474Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/request/pool.go:265 +0x31
2026-04-19T11:28:05.0907730Z github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).Stop(0xc00170d680)
2026-04-19T11:28:05.0908026Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/batcher_role.go:191 +0x151
2026-04-19T11:28:05.0908270Z github.com/hyperledger/fabric-x-orderer/node/batcher.(*Batcher).Stop(0xc002beb040)
2026-04-19T11:28:05.0908543Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/batcher.go:173 +0x1f6
2026-04-19T11:28:05.0908835Z github.com/hyperledger/fabric-x-orderer/node/batcher_test.TestPrimaryConnector(0xc0155b3180)
2026-04-19T11:28:05.0909168Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/primary_connector_test.go:68 +0x9e5
2026-04-19T11:28:05.0909271Z testing.tRunner(0xc0155b3180, 0x1c89e90)
2026-04-19T11:28:05.0909490Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/testing/testing.go:1934 +0x21d
2026-04-19T11:28:05.0909595Z created by testing.(*T).Run in goroutine 1
2026-04-19T11:28:05.0909800Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/testing/testing.go:1997 +0x9d3
```

```
2026-04-19T11:28:05.0414032Z goroutine 17808 [sync.Cond.Wait, 18 minutes]:
2026-04-19T11:28:05.0414154Z sync.runtime_notifyListWait(0xc002cb4e10, 0x3)
2026-04-19T11:28:05.0414351Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/runtime/sema.go:606 +0x159
2026-04-19T11:28:05.0414439Z sync.(*Cond).Wait(0xc002cb4e00)
2026-04-19T11:28:05.0414619Z 	/opt/hostedtoolcache/go/1.25.5/x64/src/sync/cond.go:71 +0x6c
2026-04-19T11:28:05.0414952Z github.com/hyperledger/fabric-x-orderer/request.(*BatchStore).Fetch(0xc002cb4d80, {0x1e823a0, 0xc002bfd880})
2026-04-19T11:28:05.0415225Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/request/batchstore.go:162 +0x27d
2026-04-19T11:28:05.0415565Z github.com/hyperledger/fabric-x-orderer/request.(*Pool).NextRequests(0xc001e9f130, {0x1e823a0, 0xc002bfd880})
2026-04-19T11:28:05.0415813Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/request/pool.go:216 +0x15a
2026-04-19T11:28:05.0416093Z github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).runPrimary(0xc00170d680)
2026-04-19T11:28:05.0416390Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/batcher_role.go:333 +0x1039
2026-04-19T11:28:05.0416635Z github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).run(0xc00170d680)
2026-04-19T11:28:05.0416934Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/batcher_role.go:163 +0x516
2026-04-19T11:28:05.0417251Z created by github.com/hyperledger/fabric-x-orderer/node/batcher.(*BatcherRole).Start in goroutine 17793
2026-04-19T11:28:05.0417550Z 	/home/runner/work/fabric-x-orderer/fabric-x-orderer/node/batcher/batcher_role.go:145 +0x2ab
```

[job-logs-timed-out.txt](https://github.com/user-attachments/files/26896016/job-logs-timed-out.txt)

HagarMeir:
## Problem

Fixes #843 — a test flake caused by a deadlock when a batcher is stopped.

`BatchStore.Fetch` spawned a goroutine that called `signal.Signal()` on
context cancellation **without holding the lock**. When the context was
already cancelled (or was cancelled in the window before `Fetch` reached
`signal.Wait()`), the signal was delivered with no waiter registered and
was therefore lost, leaving `Wait()` blocked forever while holding `bs.lock`.

That stuck `Fetch` is reached from `Pool.NextRequests` under the pool's
**read** lock, so `Pool.Close` (invoked from `BatcherRole.Stop` right after
`cancelBatch()`) blocked forever on the pool's **write** lock — the deadlock
seen in the reported goroutine stacks:

- one goroutine on `RWMutex.Lock` in `Pool.Close` (via `BatcherRole.Stop`);
- one goroutine on `Cond.Wait` in `BatchStore.Fetch` (via `Pool.NextRequests`).

## Fix

Correct the condition-variable usage in `BatchStore.Fetch`:

1. The `ctx.Done` goroutine now acquires `bs.lock` before `Signal()`. Because
   `Wait()` releases the lock atomically, the signal can only be delivered
   **after** `Fetch` is actually waiting on it — it can no longer be lost.
2. `Wait()` is guarded by a predicate loop that also checks the context:
   `for len(bs.readyBatches) == 0 && ctx.Err() == nil`. An already-cancelled
   context short-circuits the wait, a cancellation wake terminates the loop
   (instead of re-blocking on a signal that will never come again), and any
   stray/late wakeup is re-checked rather than acted on.

Both parts are required: lock-before-signal without the ctx-aware loop still
hangs on the ordinary timeout path; the loop without lock-before-signal still
loses the wakeup at entry.

## Test

Added `TestFetchWithCanceledContext`, which drives `Fetch` with an
already-cancelled context under a watchdog. It **deadlocks reliably against
the old code** (hangs within ~1–2k iterations) and passes with the fix.

## Verification

- `go test -race ./request/...` — pass
- `go test -race ./node/batcher/...` — pass
- `go vet` / `gofmt` — clean


request/batchstore.go
```diff
@@ -157,21 +157,27 @@ func (bs *BatchStore) Fetch(ctx context.Context) ([]interface{}, []interface{})
 	go func() {
 		select {
 		case <-ctx.Done():
+			// Acquire the lock before signaling. This guarantees the signal is
+			// delivered only after Fetch is actually waiting on it (Wait releases
+			// the lock atomically); otherwise a signal fired before Wait would be
+			// lost, leaving Fetch blocked forever while holding bs.lock and thus
+			// deadlocking Pool.Close.
+			bs.lock.Lock()
 			bs.signal.Signal()
```
tock-ibm: The ctx-cancel wakeup uses `sync.Cond.Signal()`, which wakes only one arbitrary waiter. This is safe **today** only because there is one `NextRequests`/`Fetch` caller per `BatchStore` (the batcher primary loop, sequential). But `NextRequests` takes only `rp.lock.RLock`, so nothing structurally prevents two concurrent `Fetch` calls on one store.

If that ever happens: both park in `signal.Wait()`; the ctx of Fetch-A is canceled; A's goroutine `Signal()`s, but the single wakeup can land on Fetch-B, which re-checks its own (non-canceled) ctx and re-parks — leaving Fetch-A blocked forever and re-introducing the very `Pool.Close` deadlock this PR fixes. `Broadcast()` would be robust to that; at minimum it's worth a comment documenting the one-caller-per-store invariant this relies on.

tock-ibm (commented):
## Review: fix deadlock between Pool.Close and BatchStore.Fetch

The core fix looks correct. I traced the interleavings:

- **Pre-canceled ctx**: the new loop condition `ctx.Err() == nil` is false, so `Wait()` is skipped and `Fetch` returns promptly — the deadlock is gone.
- **ctx cancels in the window between the condition check and parking in `Wait()`**: no lost signal, because `Fetch` holds the write lock continuously across that window and the ctx-watch goroutine cannot `Lock()` to `Signal()` until `Wait()` has parked (and released the lock).
- **Insert signal / spurious wakeup**: guarded by the same lock and the condition-recheck loop.
- **Inner goroutine lifetime**: always terminates via `<-finished` (closed by defer) or `<-ctx.Done()`; no leak.

The inline comments below are latent/quality issues, not blockers.

One minor, pre-existing nit (it sits outside this diff, so noting it here rather than inline): the `defer func() { close(finished) }()` in `Fetch` could be the simpler `defer close(finished)` — equivalent, and closer to the repo's "favor explicit over clever" guidance. Optional, and not introduced by this PR.

request/batchstore.go
```diff
@@ -157,21 +157,27 @@ func (bs *BatchStore) Fetch(ctx context.Context) ([]interface{}, []interface{})
 	go func() {
```
tock-ibm: The ctx-watch goroutine is spawned on every `Fetch`, before checking `readyBatches`. On the hot path — sustained load, where `readyBatches` is usually non-empty on entry — `Fetch` could return immediately, but instead pays a goroutine create + teardown (via `close(finished)`) once per batch.

Consider hoisting an early return for `len(bs.readyBatches) > 0 || ctx.Err() != nil` above the `go func()`, so the goroutine is spawned only when `Fetch` is actually about to `Wait()`.

request/batchstore_test.go
```diff
@@ -32,6 +32,34 @@ func requireMatchingBatchIDs(t *testing.T, requestID func([]byte) string, batch,
...
 
+// TestFetchWithCanceledContext reproduces a deadlock:
+// when the context passed to Fetch is canceled around the time Fetch is called,
+// the ctx.Done goroutine may Signal before Fetch reaches signal.Wait(). Since a
+// sync.Cond signal delivered with no waiter is lost, Wait() then blocks forever
+// while holding bs.lock, which in turn deadlocks Pool.Close.
+func TestFetchWithCanceledContext(t *testing.T) {
+	sugaredLogger := testutil.CreateLogger(t, 0)
+
+	for i := 0; i < 2000; i++ {
```
tock-ibm: Two small test-quality points:

- The `2000` iteration count is a magic number with no comment tying it to the race window it exercises, which makes the stress bound arbitrary to tune later. A brief comment (or a named const) would help.
- On the failure path, `t.Fatalf` fires from the main test goroutine while the blocked `Fetch` goroutine (and its inner goroutine) are left running for the rest of the process; and a genuine deadlock makes the suite hang up to the full 10s per iteration before reporting. Neither is fatal for a regression test, but both are worth noting.

Issue #1207, opened 2026-09-10
Metrics: Router client streams


tock-ibm:
### Current Status

No metrics for these events

### Expected

Add metrics for:
no. of broadcast / submit-stream streams
no. submit-request invocations

PR #1213, opened 2026-09-14
add new router client streams metrics

moran-abilea:
related tp #1207 

Adding metrics for:
no. of broadcast / submit-stream streams - active ones
no. submit-request invocations


node/router/metrics.go
```diff
@@ -43,13 +43,37 @@ var (
 		Help:       "The number of incoming requests rejected by the rate limiter.",
 		LabelNames: []string{"party_id"},
 	}
+
```
DorKatzelnick, line 46: would it be better to just count the number of active streams?
it would be simpler. we can use a gauge. Add(1) on stream open and deferred Add(-1) on close.

node/router/router.go
```diff
@@ -390,6 +390,9 @@ func (r *Router) Broadcast(stream orderer.AtomicBroadcast_BroadcastServer) error
 		r.logger.Infof("Client's certificate: \n%s", utils.CertificateToString(clientCert))
 	}
 
+	r.metrics.IncrementStreamsOpened("broadcast")
```
DorKatzelnick, line 393: add unit test
see in TestSubmitToStubBatchersGetMetrics (e.g.)
how to fetch metrics with a regular expression pattern

node/router/metrics.go
DorKatzelnick, line 141: add new metrics here

node/router/metrics.go
```diff
@@ -136,3 +163,15 @@ func (m *RouterMetrics) increaseErrorCount(err error) {
 		m.rejectedTxsWithCode500.Add(1)
 	}
 }
+
+func (m *RouterMetrics) IncrementStreamsOpened(streamType string) {
```
DorKatzelnick, line 167: why exported functions?

node/router/metrics.go
```diff
@@ -76,6 +100,9 @@ func NewRouterMetrics(routerNodeConfig *config.RouterNodeConfig, logger *floggin
 		rejectedTxsWithCode400: rejectedTxs.With([]string{"400", partyID}...),
 		rejectedTxsWithCode500: rejectedTxs.With([]string{"500", partyID}...),
 		throttledTxs:           provider.NewCounter(throttledTxs).With([]string{partyID}...),
+		streamsOpened:          provider.NewCounter(streamsOpenedOpts),
```
DorKatzelnick, line 103: why are defined differently, then other metrics (also counters) in the router?

DorKatzelnick (commented):
In addition to the inline comments:
update the documentation in docs/monitoring/metrics.md accordingly, so it has all the router metrics.

node/router/metrics.go
moran-abilea: fixed as suggested

node/router/router.go
moran-abilea: added as suggested

node/router/metrics.go
moran-abilea: fixed as suggested

node/router/metrics.go
moran-abilea: fixed as suggested

node/router/metrics.go
moran-abilea: fixed as suggested

node/router/router_test.go
```diff
@@ -219,6 +219,87 @@ func TestBroadcastThrottled(t *testing.T) {
...
+
+	URL := testSetup.router.MonitoringServiceAddress()
+	require.NotEmpty(t, URL, "monitoring service address should not be empty")
+
+	activeStreams := func(streamType string) int {
+		re := regexp.MustCompile(fmt.Sprintf(`router_active_streams\{party_id="%d",stream_type="%s"\} \d+`, types.PartyID(1), streamType))
+		return testutil.FetchPrometheusMetricValue(t, re, URL)
+	}
+
+	requireActiveStreams := func(streamType string, expected int) {
```
DorKatzelnick, line 206: `requireActiveStreams` always waits up to 30s, but two of its call sites assert a value that should already hold: L261 (`submit_stream` is 0 right after a *broadcast* stream opened) and L274 (the mirror). `Eventually` succeeds on the first poll there, so it adds no coverage — but if the gauge ever did get incremented under the wrong label, each of those failures costs a full 30s before reporting.

Suggest splitting by intent:
- keep `Eventually` for the two post-`cancel()` checks (L264, L277) — teardown genuinely is async;
- keep it for the first check at L247 too, since it doubles as "wait for the metrics endpoint to be serving";
- use `require.Equal` (or `require.Never` if you mean "stays 0") for L261 and L274, where the value is expected to already be settled.

node/router/router_test.go
```diff
@@ -219,6 +219,87 @@ func TestBroadcastThrottled(t *testing.T) {
...
+
+	broadcastCtx, cancelBroadcast := context.WithCancel(context.Background())
+	broadcastStream, err := ab.NewAtomicBroadcastClient(testSetup.clientConn).Broadcast(broadcastCtx)
+	require.NoError(t, err)
+	// exchange a request so the server side of the stream is certainly running
+	require.NoError(t, broadcastStream.Send(tx.CreateStructuredEnvelope(buff)))
+	_, err = broadcastStream.Recv()
+	require.NoError(t, err)
+
+	requireActiveStreams("broadcast", 1)
```
DorKatzelnick, line 239: Coverage here is 0→1→0 with a single stream at a time, which can't distinguish `Add(+1)/Add(-1)` from `Set(1)/Set(0)` — a gauge implemented with `Set` would pass this test unchanged, as would a double-increment on open. Accumulating across concurrent streams is the whole reason this is a gauge rather than the opened/closed counter pair, so it's worth pinning: open a second broadcast stream here and assert 2, close one and assert 1.

node/router/router_test.go
```diff
@@ -219,6 +219,87 @@ func TestBroadcastThrottled(t *testing.T) {
 	}, 30*time.Second, 100*time.Millisecond)
 }
 
+// TestRouterActiveStreamsMetric verifies that the router_active_streams gauge reports zero
```
DorKatzelnick: This file documents its tests with numbered scenario steps — 21 such comments, e.g. `TestRouterSoftStop` just above L930:

```
// 3) set the stub batcher to not answer the router requests
// 4) in parallel, send requests using submitStream and perform a soft stop on the router
// 5) expect the requests to fail with relevant error
```

Both new tests use prose instead. Worth matching the file's convention.

node/router/router_test.go
```diff
@@ -219,6 +219,87 @@ func TestBroadcastThrottled(t *testing.T) {
...
+		}, 30*time.Second, 250*time.Millisecond, "%s active streams should reach %d", streamType, expected)
+	}
+
+	// both series exist at zero before any stream is opened
+	requireActiveStreams("broadcast", 0)
+	requireActiveStreams("submit_stream", 0)
+
+	buff := make([]byte, 300)
+
+	broadcastCtx, cancelBroadcast := context.WithCancel(context.Background())
```
DorKatzelnick: Three small things around the stream setup:

- `cancelBroadcast` here and `cancelSubmit` at L266 aren't deferred, so an early `require` failure leaks both client streams for the remainder of the run.
- `testutil.FetchPrometheusMetricValue` does `t.Log(string(body))` with the entire scrape body on every call — i.e. every 250ms poll. With six `Eventually` loops in this test, a single failure dumps well over a hundred full scrapes into the test output.
- That same helper's internal `require.NoError` executes inside `Eventually`'s goroutine, so on an HTTP error it calls `t.FailNow()` off the test goroutine. Pre-existing pattern in this file, not introduced here, but combined with the point above it's what would make a failure in this test hard to read.

docs/monitoring/metrics.md
```diff
@@ -28,6 +28,18 @@ It queries Prometheus for metric data and displays it as dashboards, graphs, and
   **Name**: "requests_rejected"  
   **Help**: "The number of incoming requests that have been rejected."
 
+- **Namespace**: "router"  
+  **Name**: "requests_throttled"  
+  **Help**: "The number of incoming requests rejected by the rate limiter."
+
+- **Namespace**: "router"  
+  **Name**: "active_streams"
```
DorKatzelnick, line 36: The `stream_type` label and its two values (`broadcast`, `submit_stream`) aren't mentioned, and they're the one thing an operator needs in order to write the query:

```
router_active_streams{party_id="1", stream_type="broadcast"}
```

The surrounding entries document no labels either, so this is consistent — but for the other router metrics the labels are incidental, whereas here the split by stream type is the point of the metric. One extra line would cover it.

DorKatzelnick (commented):
Re-review of the force-pushed version. The gauge rework and the added tests address the substantive points from the previous round; the gauge children being materialized in the constructor is exactly right, and I confirmed `testutil.FetchPrometheusMetricValue` returns `-1` on no-match, so the opening `requireActiveStreams(..., 0)` genuinely proves the series exist rather than passing vacuously.

What follows is test- and doc-level polish only. Separately: the `Integration Basic Tests` failure is unrelated — it's `TestSubmitTXWithKnownCertID` asserting 2 blocks and getting 1 because the pull races the assembler's commit; nothing on that path is touched here, and it passes 3/3 locally on this branch. Just re-run the job.

node/router/router_test.go
moran-abilea: fixed as suggested

node/router/router_test.go
moran-abilea: alligned to convetion as requested

docs/monitoring/metrics.md
moran-abilea: added the one extra line as part of the readme

node/router/router_test.go
moran-abilea: fixed as suggested

node/router/router_test.go
moran-abilea: fixed as suggested

DorKatzelnick (approved):

Issue #1215, opened 2026-09-15
Config: pin the router's and the assembler's signing certificate in the shared configuration


DorKatzelnick:
Preparation for access control on the internal delivery services. `ordererpb.SharedConfig` carries a `sign_cert` for the batcher and for the consenter, so the network holds an agreed view of the signing identity of each of those nodes, and a node whose local signing identity does not match the certificate pinned for it fails to start. The router and the assembler have no such field, even though `armageddon generate` already issues a signing certificate and a corresponding key for every node, so there is no network-wide view of the signing identity of a router or an assembler to authorize against.

Add `sign_cert` to `RouterNodeConfig` and to `AssemblerNodeConfig` in `hyperledger/fabric-x-common` (https://github.com/hyperledger/fabric-x-common/pull/189), then carry it through the orderer: the yaml shared config, the `generate` path, certificate validation, and the per-role runtime config. Compare it at startup against the node's own signing identity, and treat a change to it in a new config block as requiring a restart, the way the batcher and the consenter already do.

PR #1216, opened 2026-09-15
[1/6] config: pin the router's and assembler's signing certificate in the shared config

DorKatzelnick:
https://github.com/hyperledger/fabric-x-orderer/issues/1215


common/tools/armageddon/cryptogen.go
```diff
@@ -29,8 +29,7 @@ import (
 
 // GenerateCryptoConfigWithProfile provides all crypto material of Arma network, divided into parties and written to files in a folder structure.
 // For each party for all nodes (i.e. Router, Batchers, Consenter and Assembler) it is required to provide TLS certificate and private key for secure communications between nodes using TLS.
-// For Batchers and Consenters is it also required to provide a signing certificate with a corresponding private key that are used to sign BAS's by Batchers and blocks by Consenter.
-// NOTE: for compatability with Fabric cryptogen tool and for future use, signing certificate and a corresponding private key will be created for all nodes.
+// For all nodes it is also required to provide a signing certificate with a corresponding private key, used to sign BAS's by Batchers and blocks by Consenter, and pinned in the shared configuration for every node so that a mismatch between a node's local identity and the network's view of it is detected at startup.
```
tock-ibm: BAF's

config/config.go
```diff
@@ -775,7 +775,7 @@ func GetLastConfigBlockUsingBlockFromConsensusLedger(block *common.Block, consen
 	return lastConfigBlock, nil
 }
 
-func (config *Configuration) CheckIfRouterNodeExistsInSharedConfig() error {
+func (config *Configuration) CheckIfRouterNodeExistsInSharedConfig(localSignCert []byte) error {
```
tock-ibm: why passing the cert like that? local config points to the MSP that knows where to load the signer from

tock-ibm:
## Code review notes

The core approach looks correct: the `NodeConfig` interface consolidation is safe now that all four proto types expose `GetSignCert()`, the `reflect.TypeOf` type-match check is a real improvement over the old `currOK != newOK` comparison (it catches Batcher-vs-Consenter mismatches), and cert comparison via `RawTBSCertificate` makes it insensitive to PEM formatting. The generate/testutil paths were updated consistently.

Two concerns worth a look before merge:

### 1. Backward compatibility — possible hard breaking change with no fallback (`config/config.go:787`)

A node bootstrapping from a pre-existing genesis/config block whose `RouterConfig.SignCert` (and, symmetrically, `AssemblerConfig.SignCert`) is empty will fail at startup: `AreCertificatesEqual(localSignCert, <empty>)` can't parse the empty pinned cert and returns an error, which `arma.go` `launchRouter`/`launchAssembler` turns into a panic. `LoadSharedConfig` on an older `shared_config.yaml` without a `SignCert` path hits the same via `utils.ReadPem("")`.

So an existing network upgraded in place — routers/assemblers resuming from a config block created before this change — can't start until it's regenerated/reconfigured with sign certs. Is that intended? If in-place upgrade is a supported path, consider skipping the check when the pinned cert is absent, or gating it on a capability/config version.

### 2. Scope — the `fabric-x-common` bump is broad for a config-focused change (`go.mod:13`)

The only dependency change this PR actually needs is the added `sign_cert` field on `RouterNodeConfig`/`AssemblerNodeConfig` in `configuration.pb.go`. The same bump also rewrites protolator JSON round-tripping and touches `channelconfig`/`capabilities` (~800 lines across 27 vendored files). The orderer relies on protolator/channelconfig to (de)serialize config blocks, so a subtle behavior change there could regress config-block handling with no first-party code flagging it. Unit tests pass, but the config-block round-trip is mainly exercised by the `test/reconfig` integration suites — worth making sure CI runs those, or splitting the unrelated bump out if it isn't required here.

MayRosenbaum (approved):

common/tools/armageddon/cryptogen.go
DorKatzelnick: fixed

tock-ibm (approved):

PR #1219, opened 2026-09-16
batcher,request: remove mem pool requests in parallel

HagarMeir:
## What

Follow-up to review feedback on #1205. As [@tock-ibm noted there](https://github.com/hyperledger/fabric-x-orderer/pull/1205#discussion_r4024129861):

> Can the remove be done in parallel with multiple goroutines in the memPool? `func (rp *Pool) RemoveRequests(requestsIDs ...string)` runs serially, whereas in the pending it is done in parallel.

On the primary path `Pool.RemoveRequests` looped over the ids and called `BatchStore.Remove` serially, while the pending store (secondary path) already fans removal out across `runtime.NumCPU()` workers. This mirrors that pattern for the batch store.

- Add `BatchStore.RemoveRequests`, which partitions keys across `NumCPU` workers (each key handled by exactly one worker, so no two goroutines touch the same key). `Pool.RemoveRequests` now calls it instead of looping `Remove` serially.
  - Safe because everything `Remove` touches is concurrency-safe (`sync.Map` `LoadAndDelete`/`Delete`), and `onDelete` (`semaphore.Release` + `atomic.AddInt64`) is the same callback the pending store already invokes concurrently from its workers. `LoadAndDelete` also guarantees `onDelete` fires at most once per key, so no double-decrement.
- Rename the `MemPool.RemoveRequests` parameter `requests` -> `requestIDs`, addressing [@tock-ibm's nit](https://github.com/hyperledger/fabric-x-orderer/pull/1205#discussion_r4023981201) on the interface.

## Testing

- Added `TestBatchStoreRemoveRequests` covering multi-key parallel removal, idempotency (no double `onDelete` on already-removed / unknown keys), and the empty call.
- `go test -race ./request/...` passes; existing `TestBasicBatching` (pool-level, batching-enabled removal path) still passes.

Part of #1185.


tock-ibm (commented):
Automated review of the parallel mem-pool removal change.

**Correctness: no issues.** `Remove` only touches `sync.Map`s (`keys2Batches`, `batch.m`) plus the concurrency-safe `onDelete`; index-mod partitioning gives each distinct index to exactly one worker; and `LoadAndDelete` atomicity means duplicate keys or repeat removals fire `onDelete` at most once (no semaphore double-release).

The inline notes are efficiency / reuse / convention / coverage cleanups. The two efficiency items — goroutine overhead on small batches, and each worker scanning the full slice — are the most impactful.

request/batchstore.go
```diff
@@ -121,6 +122,30 @@ func (bs *BatchStore) Prune(f func(k, v interface{}) error) {
...
+// runtime.NumCPU() workers. Each key is handled by exactly one worker (partitioned
+// by index), so no two goroutines touch the same key. Mirrors PendingStore.RemoveRequests.
+func (bs *BatchStore) RemoveRequests(keys ...string) {
+	workerNum := runtime.NumCPU()
+
+	var wg sync.WaitGroup
+	wg.Add(workerNum)
+
+	for workerID := 0; workerID < workerNum; workerID++ {
+		go func(workerID int) {
```
tock-ibm: `RemoveRequests` unconditionally forks `runtime.NumCPU()` goroutines + a `WaitGroup` join on every call, with no size threshold. `removeRequestsByIDs` invokes this for every committed batch, including small timeout-cut batches of a few requests — where goroutine-spawn/scheduling/join overhead exceeds the work, making this a regression vs. the serial loop it replaced. Consider falling back to a serial loop when `len(keys)` is below a threshold (e.g. `< workerNum`).

request/batchstore.go
```diff
@@ -121,6 +122,30 @@ func (bs *BatchStore) Prune(f func(k, v interface{}) error) {
...
+			defer wg.Done()
+			for i, key := range keys {
+				if i%workerNum != workerID {
+					continue
```
tock-ibm: Modulo-stride partitioning makes every worker range the entire `keys` slice (skipping non-owned indices), so total loop work is `workerNum × len(keys)` iterations to perform `len(keys)` removals — e.g. NumCPU=64 with 5000 keys is ~320k `i%workerNum` checks for 5000 removes. A contiguous chunk partition (`keys[start:end]` per worker) touches each key exactly once. Wasted CPU scales with core count on the batching hot path.

request/batchstore.go
```diff
@@ -121,6 +122,30 @@ func (bs *BatchStore) Prune(f func(k, v interface{}) error) {
 	bs.currentBatch.Prune(f)
 }
 
+// RemoveRequests removes multiple keys concurrently, fanning the work out across
+// runtime.NumCPU() workers. Each key is handled by exactly one worker (partitioned
+// by index), so no two goroutines touch the same key. Mirrors PendingStore.RemoveRequests.
+func (bs *BatchStore) RemoveRequests(keys ...string) {
```
tock-ibm, line 127: This fan-out (NumCPU workers, `WaitGroup`, index-mod partition) duplicates `PendingStore.RemoveRequests`/`removeRequestsByWorker` in `pending.go` (~L273-303). CLAUDE.md asks to reuse existing helpers before adding new ones. The two can already drift — `PendingStore` re-checks `isClosed()` mid-loop, this one doesn't. A shared `parallelForEachKey(keys, fn)` helper would cover both.

request/batchstore.go
```diff
@@ -121,6 +122,30 @@ func (bs *BatchStore) Prune(f func(k, v interface{}) error) {
...
+	workerNum := runtime.NumCPU()
```
tock-ibm: CLAUDE.md Performance guidance: "Profile before optimizing and validate assumptions with benchmarks rather than guessing." This parallelizes a hot path (commit → `RemoveRequests`) with only correctness testing (`go test -race`); nothing shows the serial `Remove` was a bottleneck or that fan-out is a net win. Combined with the two notes above, small/typical batches may be net-slower. A benchmark comparing serial vs. fan-out across representative batch sizes would justify the change.

request/batchstore_test.go
```diff
@@ -111,3 +111,47 @@ func TestBatchStore(t *testing.T) {
...
 }
+
+func TestBatchStoreRemoveRequests(t *testing.T) {
+	max := uint32(10000)
+	lenByte := uint32(8)
+	var removed uint32
+
+	sugaredLogger := testutil.CreateLogger(t, 0)
+
+	bs := NewBatchStore(max, max*lenByte, func(string) {
```
tock-ibm: With `batchMaxSize=10000` and `batchMaxSizeBytes=80000` (`max*lenByte`), inserting 5000 keys of 8 bytes (40000 bytes / 5000 count) never trips the "full" branch, so no batch rotation occurs and every key resolves against a single `currentBatch`. The realistic post-`Fetch` state — keys spread across several rotated `readyBatches`, each a distinct `*batch` — is left uncovered. Consider sizing the store so `Fetch` rotates several batches before removal, to exercise the multi-batch concurrent-removal path.

request/pool.go
```diff
@@ -238,9 +238,7 @@ func (rp *Pool) RemoveRequests(requestsIDs ...string) {
 		return
 	}
 
-	for _, requestID := range requestsIDs {
-		rp.batchStore.Remove(requestID)
-	}
+	rp.batchStore.RemoveRequests(requestsIDs...)
```
tock-ibm: The PR standardizes this parameter to `requestIDs` on the `MemPool` interface (`batcher_role.go`) and `PendingStore` (`pending.go`), but `Pool.RemoveRequests` — the primary `MemPool` implementer — still declares `func (rp *Pool) RemoveRequests(requestsIDs ...string)` (transposed `requestsIDs`, on the signature line just above). Worth renaming here too so the inconsistency the PR set out to remove is fully gone.

HagarMeir: Renamed the parameter to `requestIDs` in edd04428, so the standardization is now complete across the interface, `PendingStore`, and `Pool`.

request/batchstore.go
HagarMeir: Switched to contiguous-chunk partitioning in the new `parallelForEachKey` helper (edd04428): each worker iterates only its own `keys[start:end]` slice, so total work is `len(keys)` iterations rather than `workerNum × len(keys)`.

request/batchstore.go
HagarMeir: Extracted a shared `parallelForEachKey(keys, fn)` helper in `request/utils.go` (edd04428); both `BatchStore.RemoveRequests` and `PendingStore.RemoveRequests` now call it, removing the duplicated fan-out. `PendingStore` keeps its mid-loop `isClosed()` early-abort via the helper's bool-returning callback (returns `false` to stop the worker). `BatchStore` has no close concept, so its callback always returns `true`.

request/batchstore_test.go
HagarMeir: Resized the store in `TestBatchStoreRemoveRequests` (edd04428) to `batchMaxSize=100` with a large byte cap, so inserting 5000 keys rotates ~50 batches, and added `require.Greater(len(bs.readyBatches), 1)` to assert the keys genuinely span multiple `*batch` instances before removal — exercising the multi-batch concurrent-removal path.

request/batchstore.go
HagarMeir: Added `BenchmarkBatchStoreRemoveRequests` (edd04428) comparing fan-out vs. a serial `Remove` loop across sizes 1/10/100/1000/5000. Results (M1 Max, NumCPU=10): fan-out wins on large, near-max batches and loses by microseconds on small ones, with the crossover around ~1000 keys. Since default `MaxMessageCount` is 10000, full-batch commits are firmly in fan-out's favor while the small/timeout-cut penalty is microsecond-scale in absolute terms; the benchmark doc comment records the tradeoff.

request/batchstore.go
HagarMeir: Added a serial fallback in `parallelForEachKey` (edd04428) when `len(keys) < runtime.NumCPU()`. Per the benchmark added for the note below, the fan-out/serial crossover is actually ~1000 keys, so this `< NumCPU` threshold only guards the very smallest batches — kept at that level deliberately, since the small-batch penalty is microsecond-scale in absolute terms while a full 10000-key batch removal wins ~200µs. Happy to raise the threshold toward the crossover if you'd prefer to eliminate the small-batch penalty entirely.

tock-ibm (approved):

PR #1220, opened 2026-09-16
policy: set a TxID on the config TX envelope a consenter signs

### LINKED ISSUE #1152: When a consenter signs a config TX it should set a TxID
When a leader consenter consenter signs a proposed config TX, it should set a TxID on the outer envelope it creates. This helps the Committer track its TX processing.

The TxID should match the creator and nonce - use `func ComputeTxID(nonce, creator []byte) string` from `github.com\hyperledger\fabric-x-common\protoutil\proputils.go`

tock-ibm:
## What

When a leader consenter signs a proposed config transaction, it wraps the verified `ConfigEnvelope` in an outer `HeaderType_CONFIG` envelope. Until now that outer envelope's channel header had an empty `TxId`, because `protoutil.CreateSignedEnvelope` never calls `SetTxID`.

This change builds the outer envelope explicitly in `BuildVerifiedConfigRequest` so the channel header carries a `TxId` computed from the envelope's own creator and nonce (`ComputeTxID(nonce, creator)`), letting the Committer track the config TX's processing.

Fixes #1152.

## Notes

- Reuses the established manual-build idiom already used in `common/utils/genesis.go`; no vendored code is changed and no new imports are added.
- Additive/non-breaking: nothing asserts the outer CONFIG envelope's `TxId` is empty, and `ConfigRequest.ID()` reads the inner `CONFIG_UPDATE` TxId (a different location), so it is unaffected.
- The change lives in the shared `BuildVerifiedConfigRequest`, reached from both the consenter-leader path and the router path; the router discards its outer envelope, so only the consenter's envelope carries the TxId downstream — exactly as the issue intends.

## Testing

- New `TestProposeConfigUpdate_SetsTxIDOnConfigEnvelope` in `config/verify/orderer_rules_test.go` asserts the proposed outer envelope has type `CONFIG`, a non-empty `TxId`, and `TxId == ComputeTxID(nonce, creator)`.
- `go test -race ./common/policy/... ./config/verify/...` passes; `make linter` is green.


HagarMeir (commented):
## Review

Small, clean, well-constructed PR. The new `createSignedConfigEnvelope` helper is behaviorally equivalent to the previous `protoutil.CreateSignedEnvelope` + `SetTxID`, and setting an explicit outer `TxID` is a sensible improvement.

I traced both callers (consenter path in `node/consensus/consensus.go` and router path in `node/router/config_submitter.go`), confirmed the downstream `ConfigRequest.ID()` reads the inner `CONFIG_UPDATE` TxId (unaffected), and verified the re-verification path does not inspect the outer TxID. The added non-determinism (TxID derived from a fresh nonce) is consistent with pre-existing nonce/timestamp non-determinism, so no follower byte-comparison is broken.

**No correctness bugs found.** Two minor robustness notes — both latent (not reachable by current callers), for consideration only:

1. **Unconditional `signer` dereference** (`common/policy/policy.go`): the helper calls `signer.Serialize()` / `signer.Sign(...)` unconditionally, whereas the replaced `protoutil.CreateSignedEnvelope` guarded every signer use with `if signer != nil`. Both call sites supply a real signing identity today, so it's not reachable, but the new helper silently drops the nil-tolerance the old primitive guaranteed. This is the higher-value of the two to guard, since a nil signer is the more plausible future misuse.

2. **`MarshalOrPanic` drops the old nil-check** (`common/policy/policy.go`): `protoutil.MarshalOrPanic(configEnvelope)` panics on a nil/invalid `ConfigEnvelope`, whereas `CreateSignedEnvelope` did a `ProtoReflect().IsValid()` nil-check and returned a graceful error. Not reachable today (`AuthorizeAndVerifyConfigUpdate` always returns non-nil).

Neither blocks merge.

tock-ibm:
Addressed both robustness notes in bee721c0:

1. **Nil signer** — `createSignedConfigEnvelopeWithTxID` now returns an explicit `"signer is nil"` error instead of dereferencing a nil signer, restoring the nil-safety the old `protoutil.CreateSignedEnvelope` primitive provided (as an error rather than an unsigned envelope, since a signed config TX inherently requires a signer).
2. **Nil config envelope** — added an explicit `configEnvelope == nil` guard, mirroring the existing check in `BuildBundleFromBlock`.

Also renamed the helper `createSignedConfigEnvelope` → `createSignedConfigEnvelopeWithTxID` so the name reflects the added outer TxID, and added an internal test (`TestCreateSignedConfigEnvelopeWithTxID_NilInputs`) covering both guards, which aren't reachable through the exported API.

HagarMeir (approved):

Issue #1221, opened 2026-09-17
deliver: an internal deliver request's signer is never resolved to a node


DorKatzelnick:
Both internal deliver services — the batcher's and the consenter's — authorize nothing: the policy checker they install returns `nil` for every request.

The information needed is already present. The shared configuration names the signing certificate of each party's router, batcher, consenter and assembler. But nothing in the deliver path reads the identity out of a request, matches it against those certificates, or verifies the request's signature. A batcher cannot tell the assembler of its own party from any other holder of a certificate chained to the same CA, and a consenter cannot tell a router of the network from an arbitrary client.

Both services need the same check, so absent a shared one it gets written twice.

Part of #1050.

PR #1225, opened 2026-09-17
[2/6] common/deliver: authorize an internal deliver request by the node that signed it

DorKatzelnick:
https://github.com/hyperledger/fabric-x-orderer/issues/1221


common/deliver/node_policy.go
```diff
@@ -0,0 +1,271 @@
...
+		return errors.Wrapf(err, "failed parsing the signing certificate of the %s", identity)
+	}
+
+	publicKey, ok := cert.PublicKey.(*ecdsa.PublicKey)
+	if !ok {
+		return errors.Errorf("the signing certificate of the %s holds a %T public key, expected ECDSA",
+			identity, cert.PublicKey)
+	}
+
+	if existing, taken := v.nodes[string(cert.RawTBSCertificate)]; taken {
```
tock-ibm, line 203: **A shared signing certificate blocks the whole service from building, even for roles it never serves.**

`newNodeVerifier` indexes *all* roles of *all* parties (`NewBatcherDeliverVerifier` adds routers and consenters too). If an operator reuses one signing identity for, say, a party's router and consenter — neither of which the batcher deliver service authorizes — `addNode` returns `"the router ... and the consenter ... share a signing certificate"`, `NewBatcherDeliverVerifier` propagates it, and (once wired in a later PR) the batcher's deliver service cannot start — even though it would never authorize a router or consenter anyway.

The uniqueness requirement is broader than the roles the service actually serves. Consider enforcing uniqueness only across the roles this verifier will authorize.

common/deliver/node_policy.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	signer, known := v.nodes[string(cert.RawTBSCertificate)]
+	if !known {
+		return NodeIdentity{}, errors.New(
+			"the request carries a signing certificate that is not in the shared configuration",
+		)
+	}
+
+	digest := sha256.Sum256(signedData[0].Data)
```
tock-ibm, line 262: **A non-P256 signing key indexes successfully but can never be authorized.**

`addNode` (line 197) accepts any `*ecdsa.PublicKey` (P-256/P-384/P-521), but `VerifyRequest` always hashes with `sha256.Sum256`. If a deployment configures a node's signing identity on P-384/P-521 (whose MSP/BCCSP signer hashes with SHA-384/512), the verifier builds without complaint yet every internal deliver request from that node fails with `"signature ... does not verify"`, giving the operator no hint that the curve is the cause.

This matches the existing sha256-hardcoded convention in `node/crypto/verifier.go` (so P-256 works), but the silent accept-then-always-reject for other curves is a trap. Consider rejecting a non-P256 curve in `addNode` so the failure surfaces at construction time.

common/deliver/node_policy_test.go
```diff
@@ -0,0 +1,700 @@
...
+		{PartyID: 1, Role: deliver.RoleConsenter},
+		{PartyID: 1, Role: deliver.RoleAssembler},
+		{PartyID: 2, Role: deliver.RoleRouter},
+		{PartyID: 2, Role: deliver.RoleBatcher, ShardID: 1},
+		{PartyID: 2, Role: deliver.RoleBatcher, ShardID: 2},
+		{PartyID: 2, Role: deliver.RoleConsenter},
+		{PartyID: 2, Role: deliver.RoleAssembler},
+	}
+
+	certs := make(map[deliver.NodeIdentity][]byte, len(nodes))
```
tock-ibm, line 98: **Duplicates the `signersOfNodes` helper.**

`signersOfNodes` (line 628) does precisely this certs/signers map-building loop and is used by `TestDeliverServiceRoles` and `TestBatcherDeliverServiceShard`. `TestVerifyRequestOfEveryNode` (here) and `TestCheckPolicy` (lines 375-381) inline its body instead of calling it — duplicated maintenance across four tests for one construction pattern.

tock-ibm (commented):
Automated code review of this PR. A clean, purely-additive `NodeVerifier` for authorizing internal deliver requests, with thorough tests. I verified both packages compile and `go vet` cleanly, the new `common/deliver` unit tests pass, and the `config` integration test genuinely runs a real 4-party/2-shard generated network and passes. The core authorization crypto is sound: identity is bound inside the signed payload and lookup is by `RawTBSCertificate`, so the public key is pinned from config rather than trusted from the request — presenting a victim's public cert cannot bypass verification. No correctness bugs found in the production authorization logic; the inline comments are robustness/altitude/cleanup observations.

common/deliver/node_policy.go
DorKatzelnick: a shared certificate is now an error only when one of the two nodes connects to this service

common/deliver/node_policy.go
DorKatzelnick: Fixed: `addNode` now rejects a key that is not on P-256

common/deliver/node_policy_test.go
DorKatzelnick: Fixed: both tests now call `signersOfNodes`.

common/deliver/node_policy.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// NodeVerifier authorizes a request one node sends to another: it resolves the signer by the signing
+// certificate the request carries, verifies the signature against it, and refuses a role that does
+// not connect to the service. Certificates are public, so only the signature proves possession.
+type NodeVerifier struct {
+	nodes        map[string]node
+	allowedRoles map[NodeRole]struct{}
+	// shardOfBatchers restricts batchers to one shard. Nil accepts the batchers of every shard.
+	shardOfBatchers *types.ShardID
```
tock-ibm, line 57: Why a pointer? shardID==0 is forbidden, it can mean `missing`.

common/deliver/node_policy.go
```diff
@@ -0,0 +1,293 @@
...
+// follows the stream of decisions, so all four are served.
+func NewConsenterDeliverVerifier(bundle channelconfig.Resources) (*NodeVerifier, error) {
+	return newNodeVerifier(bundle, nil, RoleRouter, RoleBatcher, RoleConsenter, RoleAssembler)
+}
+
+// NewNodeVerifier collects the signing certificate of every node of the shared configuration the
+// bundle carries, and accepts only a node whose role is in rolesThatConnect. A node with no
+// certificate is left out; an unusable certificate is an error.
+func NewNodeVerifier(bundle channelconfig.Resources, rolesThatConnect ...NodeRole) (*NodeVerifier, error) {
+	return newNodeVerifier(bundle, nil, rolesThatConnect...)
```
tock-ibm, line 81: shardID==0 could mean missing/none

common/deliver/node_policy.go
```diff
@@ -0,0 +1,293 @@
...
+
+// NodeRole is the service a node runs.
+type NodeRole string
+
+const (
+	RoleRouter    NodeRole = "router"
+	RoleBatcher   NodeRole = "batcher"
+	RoleConsenter NodeRole = "consenter"
+	RoleAssembler NodeRole = "assembler"
+)
```
tock-ibm, line 33: I think we already have these constants defined somewhere, don't we?

common/deliver/node_policy.go
tock-ibm: why do we need both `NewConsenterDeliverVerifier` and `NewNodeVerifier`?

common/deliver/node_policy.go
```diff
@@ -0,0 +1,293 @@
...
+	return append(nodes, signCertOfNode{
+		identity: NodeIdentity{PartyID: partyID, Role: RoleAssembler},
+		signCert: party.GetAssemblerConfig().GetSignCert(),
+	})
+}
+
+// addNode indexes a node by the DER of the signed part of its signing certificate, so that neither
+// the PEM encoding nor a re-encoded certificate signature matters: an MSP rewrites a high-S ECDSA
+// signature to low-S, which leaves the certificate equivalent but changes its bytes.
+func (v *NodeVerifier) addNode(signCert []byte, identity NodeIdentity) error {
```
tock-ibm, line 188: Note that we are not verifying anywhere that the signing certs of entities are different, especially not within the same party... up until now, for all we care, they can be the same. This may be a good implementation call, but if a config block has this kind of feature, the network will simply refuse to connect. If it happens on the genesis block it will do no harm, but if it is on a config change, we have a real problem.

We should either strictly prevent this in the config block verification, or be more flexible here.

See also how it is implemented here: 
https://github.com/hyperledger/fabric-x-orderer/blob/b8b456c2738769359635e594a837174dad37f46e/node/comm/clusterservice.go#L125

common/deliver/node_policy.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+type node struct {
+	identity  NodeIdentity
+	publicKey *ecdsa.PublicKey
+}
+
+// NewBatcherDeliverVerifier builds the access control of the batcher deliver service of a shard.
+// Only the batchers of that shard and the assemblers pull its batches.
+func NewBatcherDeliverVerifier(bundle channelconfig.Resources, shardID types.ShardID) (*NodeVerifier, error) {
```
tock-ibm: I think this style is very implicit in what is allowed and what not. You really have to dig in and read internal functions. A more easily readable style would be to create a NodeVerifier with just the bundle and the self-shard (=0 if not batcher). Then invoke:

`(v *NodeVerifier) AddNode(party, role, shard) error` as many times as needed.

alternatively, supply an explicit list of tuples `[]<party, role, shard>`.
This way, looking at the first creation of this verifier it is easy to see what it allows.
What do you think?

common/deliver/node_policy.go
tock-ibm: After discussion - let's not enforce uniqueness here, we will enforce it on the config verification.

common/deliver/node_policy.go
DorKatzelnick: Fixed. the field is now types.ShardID, with AnyShard = 0 naming the unrestricted case.

common/deliver/node_policy.go
DorKatzelnick: fixed.
NewNodeVerifier is the only constructor now

common/deliver/node_policy.go
DorKatzelnick: Checked: the closest are protos.NodeType, which is the config-ack wire enum and has no consenter, and the config.*Str strings used by config parsing — neither fits, so I kept these.

common/deliver/node_policy.go
DorKatzelnick: Changed it. Now we call NewNodeVerifier with an explicit roles list and the corresponding shard id

common/deliver/node_policy.go
DorKatzelnick: no more checking for collisions here

common/deliver/node_policy.go
tock-ibm: OK, we can fix it later, maybe in the types package.

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+
+type node struct {
+	identity  NodeIdentity
+	publicKey *ecdsa.PublicKey
+}
+
+// NewNodeVerifier collects the signing certificate of every node of the shared configuration the
+// bundle carries, and accepts only a node whose role is in rolesThatConnect, and a batcher only from
+// shardOfBatchers. A node with no certificate is left out; an unusable certificate is an error.
+func NewNodeVerifier(
```
tock-ibm: What I meant in my previous comment is to have 

```
NewNodeVerifier(
  bundle channelconfig.Resources, 
  shardID types.ShardID,                     //>0 for batchers, ==0 for non-batchers
  nodesThatConnect ...NodeIdentity,  // <<<
) (*NodeVerifier, error) {
```

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+func (n NodeIdentity) String() string {
+	if n.Role == RoleBatcher {
+		return fmt.Sprintf("batcher of party %d in shard %d", n.PartyID, n.ShardID)
+	}
+	return fmt.Sprintf("%s of party %d", n.Role, n.PartyID)
+}
+
+// AnyShard lifts the shard restriction of a verifier, for a service the batchers of every shard
+// connect to.
+const AnyShard types.ShardID = 0
```
tock-ibm: AnyShard -> NoShard

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+		return nil, errors.New("no role connects to the service, so every request would be refused")
+	}
+
+	v := &NodeVerifier{
+		nodes:           make(map[string]node),
+		roles:           make(map[NodeRole]struct{}, len(rolesThatConnect)),
+		shardOfBatchers: shardOfBatchers,
+	}
+
+	for _, role := range rolesThatConnect {
```
tock-ibm: if we use `nodesThatConnect` this is simpler and explicit

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+		signCert: party.GetAssemblerConfig().GetSignCert(),
+	})
+}
+
+// addNode indexes a node by the DER of the signed part of its signing certificate, so that neither
+// the PEM encoding nor a re-encoded certificate signature matters: an MSP rewrites a high-S ECDSA
+// signature to low-S, which leaves the certificate equivalent but changes its bytes.
+func (v *NodeVerifier) addNode(signCert []byte, identity NodeIdentity) error {
+	if len(signCert) == 0 {
+		return nil
```
tock-ibm: why not an error?

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+	// A request is verified over a SHA-256 digest, so another curve indexes a node whose every
+	// request would fail to verify.
+	if publicKey.Curve != elliptic.P256() {
+		return errors.Errorf("the signing certificate of the %s holds a %s public key, expected P-256",
+			identity, publicKey.Curve.Params().Name)
+	}
+
+	// The verification of a configuration refuses a signing certificate that two nodes share, so a
+	// certificate resolves to exactly one node here.
+	v.nodes[string(cert.RawTBSCertificate)] = node{identity: identity, publicKey: publicKey}
```
tock-ibm: This is fine, in case two nodes are mapped to the same key, they both are admitted, but only the last added is remembered as the owner of the key. We make sure this does not happen in the config verification.
Maybe add a warning when we overwrite here.

common/deliver/node_policy.go
```diff
@@ -0,0 +1,264 @@
...
+
+	return nil
+}
+
+// CheckPolicy refuses a request that no node of the shared configuration signed, that a role which
+// does not connect to this service signed, or that a batcher of another shard signed. The channel is
+// not part of the decision.
```
tock-ibm: a simpler description - refuses any node that is not in the "NodesThatConnect"

tock-ibm:
@DorKatzelnick look at the extended crypto/verifier that @HagarMeir is building here:
https://github.com/hyperledger/fabric-x-orderer/pull/1241
it seems to overlap what you do here, can we reuse her code?

common/deliver/node_policy.go
DorKatzelnick: Done

common/deliver/node_policy.go
DorKatzelnick: I kept the AnyShard.
A batcher entry that uses shardId != AnyShard means batcher of a **specific** shard is allowed
A batcher entry that uses shardId == AnyShard means batcher of **any** shard is allowed

common/deliver/node_policy.go
DorKatzelnick: Done

common/deliver/node_policy.go
DorKatzelnick: Fixed: it is an error now

common/deliver/node_policy.go
DorKatzelnick: Added the warning

common/deliver/node_policy.go
DorKatzelnick: Done

DorKatzelnick: Done

Issue #1222, opened 2026-09-17
node/batcher: any mTLS client can stream every batch a batcher holds


DorKatzelnick:
`policyChecker.CheckPolicy` on the batcher's deliver service returns `nil` unconditionally, so per-request authorization never happens. Transport is the only gate: `CreateGRPCBatcher` hardcodes `RequireClientCert: true, UseTLS: true`, so a caller must present a certificate chained to a shard-batcher CA — and nothing beyond that is checked.

Anyone past that gate can open a Deliver stream and read every batch the batcher has stored. Only the assemblers of the network and the other batchers of the same shard have any reason to pull from it.

Part of #1050.

PR #1226, opened 2026-09-17
[3/6] node/batcher: authorize the node that pulls a batch by its signature

DorKatzelnick:
https://github.com/hyperledger/fabric-x-orderer/issues/1222


common/deliver/node_policy.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	signer, known := v.nodes[string(cert.RawTBSCertificate)]
+	if !known {
+		return NodeIdentity{}, errors.New(
+			"the request carries a signing certificate that is not in the shared configuration",
+		)
+	}
+
+	digest := sha256.Sum256(signedData[0].Data)
```
tock-ibm, line 262: `VerifyRequest` hardcodes `sha256.Sum256`, but `addNode` accepts *any* ECDSA public key (it only rejects non-ECDSA at line 197 — no curve check). A node whose signing key is P-384/P-521, or whose MSP `SignatureHashFamily` is not SHA2/SHA256, will have `signingidentity.Sign` hash the payload with a different digest, so `ecdsa.VerifyASN1(pub, sha256(payload), sig)` will *always* fail. The certificate is still indexed happily in `addNode`, so the mismatch is silent: that node can never pull batches or replicate, and every internal deliver request from it is refused with a confusing "does not verify" error. Default armageddon deployments (P-256 + SHA-256) are unaffected, but this couples correctness to an unchecked crypto assumption. Consider deriving the hash from the certificate/curve (or the MSP hash family), or explicitly rejecting non-P256 keys in `addNode`.

— flagged by Claude Code

node/batcher/batcher_builder.go
```diff
@@ -83,12 +84,18 @@ func (b *Batcher) configureBatcher(senderCreator ConsenterControlEventSenderCrea
 		b.logger.Panicf("Failed creating BatchLedgerArray: %s", err.Error())
 	}
 
+	accessControl, err := deliver.NewBatcherDeliverVerifier(b.config.Bundle, b.config.ShardId)
+	if err != nil {
+		b.logger.Panicf("Failed creating the access control of the batcher deliver service: %s", err)
```
tock-ibm, line 89: `configureBatcher` runs on every reconfiguration (`stopAndReconfigure` -> `configureBatcher`), not only at startup. Building the access control here with `Panicf` on failure means a *new config block* whose shared configuration cannot produce a verifier — e.g. it drops the shard's sign certs, carries a non-ECDSA cert, or two nodes that end up sharing a signing certificate (the `addNode` collision check) — will now crash a running batcher instead of being rejected/handled gracefully. Before this PR the deliver policy was a no-op, so reconfig could not fail this way. It is consistent with the other `Panicf` calls in this function, but worth confirming that the reconfig path can never legitimately hand this a shared config the verifier rejects.

— flagged by Claude Code

test/utils/utils.go
```diff
@@ -539,10 +590,10 @@ func RecoverConsenter(t *testing.T, ca tlsgen.CA, conf *node_config.ConsenterNod
 	return consenter
 }
 
-func RecoverAssembler(t *testing.T, conf *node_config.AssemblerNodeConfig, logger *flogging.FabricLogger, lastConfigBlock *common.Block) *assembler.Assembler {
+func RecoverAssembler(t *testing.T, conf *node_config.AssemblerNodeConfig, logger *flogging.FabricLogger, lastConfigBlock *common.Block, assemblerIdentity *SigningIdentity) *assembler.Assembler {
 	configuration := testutil.ConfigurationWithDefaultCluster()
 	configuration.LocalConfig.ClusterConfig.ReplicationPolicy = config.ReplicationPolicyAssembler
-	assembler := assembler.NewAssembler(conf, configuration, lastConfigBlock, make(chan struct{}), logger, &mocks.SignerSerializer{})
+	assembler := assembler.NewAssembler(conf, configuration, lastConfigBlock, make(chan struct{}), logger, assemblerIdentity.Signer)
```
tock-ibm, line 596: `RecoverAssembler` unconditionally dereferences `assemblerIdentity.Signer`, so a `nil` argument panics. Its sibling `CreateBatchersForShard` was deliberately made nil-tolerant (`sharedConfigOfShard` guards `i < len(assemblerIdentities)`), and callers pass `nil` there (e.g. `test/faulttolerance/batcher_consenter_test.go`). This helper currently has no callers, but the asymmetry is a latent trap: a future recovery test mirroring the faulttolerance pattern (which has no assemblers) would nil-panic here. Consider tolerating a nil identity or documenting that one is required.

— flagged by Claude Code

common/deliver/node_policy.go
DorKatzelnick: the curve check lives in `addNode`, fixed in #1225 which this branch builds on

node/batcher/batcher_builder.go
DorKatzelnick: kept as `Panicf`: a batcher that cannot resolve its node set from a config block cannot serve pulls, and the sibling ledger failure beside it panics too

test/utils/utils.go
DorKatzelnick: `RecoverAssembler` now fails fast on a nil identity — a stub signer would have the batcher refuse its pulls and hang the test

